PR #49, opened 2022-11-07
fix: internal error when Prisma client runtime error occurs

ymc9:
We can't check Prisma's error type against PrismaClientKnownError class because we can't reference generated Prisma classes from the `internal` package. Use a weakly-typed checking instead.


ymc9:
#49

PR #59, opened 2022-11-07
fix: handling of bytes/data/bigint/decimal data types

ymc9:
use extra typing information for back and forth serialization


ymc9:
use extra typing information for back and forth serialization

PR #154, opened 2023-01-01
refactor: refactor policy check code, add withPassword enhancer, fix tests

ymc9:


ymc9:
A few plugins are treated as builtin and don't need to be explicitly configured in ZModel:
@zenstack/prisma, @zenstack/model-meta, @zenstack/access-policy.

Model meta and policy code are, by default, generated into node_modules/.zenstack

PR #165, opened 2023-01-14
feat: revision to "zenstack init" command

ymc9:


ymc9:
- revise "zenstack init"
- generate "do not modify" header ot prisma schema
- more robustly load zenstack version

PR #211, opened 2023-02-16
fix: incremental bugfixes

ymc9:


ymc9:
- wrong http method for "upsert" in generated react hooks
- recognize access policies in relation filters: *-to-many (some, none, every) and *-to-one (is, isNot)
- allow to set a custom logger in next.js integration

PR #286, opened 2023-03-21
fix: Support implicit many-to-many

jiashengguo:


ymc9:
Maybe add a test case for it? There're cases for relation validation in `/packages/schema/tests/schema/validation/datamodel-validation.test.ts`

PR #336, opened 2023-04-06
fix: Load plugin models in vscode extension

jiashengguo:


packages/schema/src/language-server/zmodel-workspace-manager.ts
```diff
@@ -1,21 +1,115 @@
...
+        const documents: LangiumDocument[] = [];
+        const collector = (document: LangiumDocument) => {
+            documents.push(document);
+            if (!this.langiumDocuments.hasDocument(document.uri)) {
+                this.langiumDocuments.addDocument(document);
+            }
+        };
+        // Even though we don't await the initialization of the workspace manager,
+        // we can still assume that all library documents and file documents are loaded by the time we start building documents.
+        // The mutex prevents anything from performing a workspace build until we check the cancellation token
```
ymc9, line 42: What does this "mutex" mean here?

packages/schema/src/language-server/zmodel-workspace-manager.ts
```diff
@@ -1,21 +1,115 @@
...
+                            collector(document);
+                            console.log(`Adding plugin document from ${path}`);
+
+                            pluginModels.delete(plugin);
+                            // early exit if all plugins are loaded
+                            if (pluginModels.size === 0) {
+                                return;
+                            }
+                        } catch {
+                            //no-op
```
ymc9, line 106: Shall we add a warning log here?

packages/schema/src/language-server/zmodel-workspace-manager.ts
jiashengguo: It's copied from the base class `DefaultWorkspaceManager`.  I think it means that the `initializeWorkspace` function itself is actually running in the `mutex.lock`.  Anyway, I think I will just remove it

packages/schema/src/language-server/zmodel-workspace-manager.ts
jiashengguo: There might be multiple node_modules locations,  the warning will be shown eventually after traversing all the folders but not found.

ymc9 (approved):

PR #354, opened 2023-04-15
fix: better cli reporting of missing/invalid config file

### LINKED ISSUE #343: CLI should report error when a non-existent config file is specified

potion-cellar:
fixes #343 

 - slightly more granular error reporting for config file issues
 - slightly better test coverage

### DIFF
--- a/packages/schema/src/cli/config.ts
+++ b/packages/schema/src/cli/config.ts
@@ -5,10 +5,10 @@ import { fromZodError } from 'zod-validation-error';
 import { CliError } from './cli-error';
 
 const schema = z
-    .object( {
+    .object({
         guardFieldName: z.string().default( GUARD_FIELD_NAME ),
         transactionFieldName: z.string().default( TRANSACTION_FIELD_NAME ),
-    } )
+    })
     .strict();
 
 export type ConfigType = z.infer<typeof schema>;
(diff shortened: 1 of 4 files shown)


packages/schema/src/cli/config.ts
```diff
@@ -19,22 +19,22 @@ export let config: ConfigType = schema.parse({});
...
-        return;
-    }
-
-    let content: unknown;
+export function loadConfig( filename: string ) {
     try {
-        content = JSON.parse(fs.readFileSync(filename, 'utf-8'));
-    } catch {
-        throw new CliError(`Config is not a valid JSON file: ${filename}`);
+        const fileData = fs.readFileSync( filename, `utf-8` );
```
ymc9, line 24: I see there're extra leading and trailing spaces within parentheses. Did this happen when the code is auto-formatted in VSCode? Not sure if there's any inconsistency in prettier settings between our settings.

potion-cellar: Yes it's VSCode doing that specific formatting. I have prettier running but that looks like there is a conflict going on

potion-cellar: Okay I have prettier taking precedence now, looks like that's resolved.

PR #355, opened 2023-04-18
fix date string zod validation

Xemah:


packages/schema/src/plugins/zod/transformer.ts
```diff
@@ -114,7 +114,7 @@ export default class Transformer {
             } else if (inputType.type === 'Boolean') {
                 result.push(this.wrapWithZodValidators('z.boolean()', field, inputType));
             } else if (inputType.type === 'DateTime') {
-                result.push(this.wrapWithZodValidators('z.date()', field, inputType));
+                result.push(this.wrapWithZodValidators('z.union([z.date(), z.string().datetime()])', field, inputType));
```
ymc9, line 117: Hey, thanks for working on this!

I tried it out, and there's a problem with the distribution of `union` over `array`:

image

image

Here the inferred type doesn't conform to Prisma filter's typing. The `union` needs to wrap around `array`. Maybe we should refactor `wrapWithZodValidators` and and pass in a list of validators so we can deal with the array distribution inside of it?

Xemah: Maybe, we have to do like if its a union, then use '.array()' for each of the rule inside union
Like if it's `z.union([z.date(), z.string().datetime()])`, it should do `z.union([z.date().array(), z.string().datetime().array()])`

ymc9: Yes, I think so.

PR #357, opened 2023-04-18
chore: fix test directory cleanup

ymc9:


ymc9:
- Make sure test directory is removed when run completes on an error happens.
- Pin zod version due to: https://github.com/colinhacks/zod/issues/2184

PR #363, opened 2023-04-22
fix: zod typing for `DateTime` field, improve overall code generation

ymc9:


ymc9:
- Fixed zod generation for `DateTime` field by @Xemah 
- Don't save ts files for core plugins when the generation target is the default location (can be overridden by the `perserveTsFiles` plugin option
- Added the `compile` plugin option to core plugins (always compiled if generating to default location)
- Optimize ts-morph project init settings
- Generate `index.ts` for zod schemas for enums and objects
- Don't explicitly clean up test temp folders

Issue #391, opened 2023-05-05
field can't be named `model`


he-la:
**Description and expected behavior**
Create a model with a field named `model`:

```
model MyModel {
  model  String
}
```

`npx zenstack generate` errors with

```
Validation errors:
line 548: Expecting: expecting at least one iteration which starts with one of these possible Token sequences::
  <[TRIPLE_SLASH_COMMENT] ,[ID] ,[in] ,[@@]>
but found: 'model' []
line 548: Expecting: one of these possible Token sequences:
  1. [ID]
  2. [in]
but found: 'String' []
line 548: Expecting token of type '{' but found `String`. []
line 548: Expecting token of type 'EOF' but found `String`. []
```

**Environment (please complete the following information):**

-   ZenStack version: `1.0.0-alpha.113`
-   Prisma version: `4.11.0`
-   Database type: PostgreSQL

**Additional context**
Yup, I actually have a legit usecase for a field named `model`. Don't ask...

ymc9:
I won't ask but I believe "model" is a good name for model haha. Prisma did a good job permitting it. We should fix it here.

ymc9:
Hi @he-la , this issue is resolved by the latest alpha.114 release.

ymc9:
Fixed by #400

Issue #390, opened 2023-05-04
`@@index()` doesn't support `type`


he-la:
**Is your feature request related to a problem? Please describe.**
Prisma supports setting an index type for PostgreSQL databases, e.g. `@@index([field1, field2], type: BTree)` ([docs](https://www.prisma.io/docs/concepts/components/prisma-schema/indexes)).

Zenstack complains with `Attribute "@@index" doesn't have a parameter named "type" []`

**Describe the solution you'd like**
Zenstack should be fully compatible with the prisma schema.

**Describe alternatives you've considered**

**Additional context**
- ZenStack version: 1.0.0-alpha.113
- Prisma version: 4.11.0
- Database type: Postgresql

ymc9:
Good catch! It seems several attribute parameters are missing.

ymc9:
If it's blocking, you can use the `@@prisma.passthrough` as a temporary workaround.

```
model User {
  @@prisma.passthrough('@@index([field1, field2], type: BTree)')
}
```

he-la:
Thanks!

ymc9:
I'll include a fix in the next release 😄. PR is also welcome!

he-la:
I'm currently trying to migrate our production to zenstack. Will try to find some time to PR once I have confidence that ZenStack will work for us :rocket:

ymc9:
Sure. Please keep the issues coming, and let us know which are blocking you so we can fix them with a higher priority.

ymc9:
Hi @he-la , this issue is resolved by the latest alpha.114 release.

ymc9:
closed by #400

PR #394, opened 2023-05-05
fix: add missing attribute parameters and loosen keyword restrictions

ymc9:
- add missing params for `@@index`, `@@unique`, `@@id`
- allow more keywords as declaration names


ymc9:
#390 #391

Issue #438, opened 2023-05-29
[Bug] Error with multiple fields based ID


Azzerty23:
**Description and expected behavior**
I'm trying to set up a rest flavor api with NextAuth and I encountered this error: `An unhandled error occurred: Error: model verificationToken does not have an id field`

**Screenshots**
![Capture d’écran 2023-05-29 à 16 03 28](https://github.com/zenstackhq/zenstack/assets/43639468/9d1fd6d9-3d64-42fc-8890-99ef07915dd7)

**Environment (please complete the following information):**

-   ZenStack version: [e.g., 1.0.0-alpha.119]
-   Prisma version: [e.g., 4.12.0]
-   Database type: [e.g. Postgresql]

**Additional context**
I'm trying to use NextAuth with normal setup : https://authjs.dev/reference/adapter/prisma#create-the-prisma-schema-from-scratch

Azzerty23:
When I retry the same request immediately after, I receive a 403 error (even though it's a public endpoint): 
![Capture d’écran 2023-05-29 à 16 38 31](https://github.com/zenstackhq/zenstack/assets/43639468/7cf58bd7-9e81-4b42-8ec1-2d8f44f71b1b)

ymc9:
Hi @Azzerty23 , models with multiple ID fields are not supported yet. The main challenge is how to format the RESTful URL ..., and there's no specification for it in JSON:API. Do you have a use case for it?

The 500 error is indeed not ideal. I'll look into fixing it.

ymc9:
Got it. Let me check it and make a fix soon.

ymc9:
Hi @Azzerty23 , the issue should have been fixed with release 122. Please update and give it a try. Thanks!

Azzerty23:
It works perfectly, I can't thank you enough. 🙏

You even added a log for this :
<img width="815" alt="Capture d’écran 2023-06-02 à 10 22 43" src="https://github.com/zenstackhq/zenstack/assets/43639468/8462e53c-82f5-44ab-8334-ded524b726f4">

PR #446, opened 2023-06-01
fix: open-api issues

ymc9:
- generate aux fields when a model is related by another model with policies
- avoid throwing then a model doesn't have id field
- add Prisma error code to error response if available


ymc9:
Fixes #438

Issue #449, opened 2023-06-02
Incorrect error response when a policy rejection occurs (in REST API flavor context)


Azzerty23:
**Description and expected behavior**
I expect to receive a 403 error response indicating a lack of permission.

**Screenshots**
![Capture d’écran 2023-06-02 à 15 16 59](https://github.com/zenstackhq/zenstack/assets/43639468/b9bd0fd5-ebad-4fa2-b789-aba92ca9a728)

**Environment (please complete the following information):**

-   ZenStack version: [e.g., 1.0.0-alpha.122]
-   Prisma version: [e.g., 4.14.0]
-   Database type: [e.g. Sqlite]

**Additional context**
Response detail: Cannot read properties of undefined (reading 'PrismaClientKnownRequestError')\nTypeError: Cannot read properties of undefined (reading 'PrismaClientKnownRequestError')\n    at prismaClientKnownRequestError (/Users/augustin/Documents/Dev/Test/zenstack-rest-turbo/apps/express-api/dist/index.js:15890:50)\n    at PolicyUtil.deniedByPolicy (/Users/augustin/Documents/Dev/Test/zenstack-rest-turbo/apps/express-api/dist/index.js:24086:58)\n    at PolicyProxyHandler.<anonymous> (/Users/augustin/Documents/Dev/Test/zenstack-rest-turbo/apps/express-api/dist/index.js:24460:30)\n    at Generator.next (<anonymous>)\n    at fulfilled (/Users/augustin/Documents/Dev/Test/zenstack-rest-turbo/apps/express-api/dist/index.js:24200:28)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Azzerty23:
It seems to be an issue on my side since it's working fine in dev mode...

ymc9:
Got it. @Azzerty23 let me know in case you find something suspicious on ZenStack side.

Azzerty23:
Hi @ymc9, I have identified the source of my issue and found a way to resolve it by modifying how the ZenStack runtime loads the Prisma module. You can find the relevant code here: https://github.com/zenstackhq/zenstack/blob/afa23882422d0fdf772d9a23e4cc157399b19a99/packages/runtime/src/enhancements/utils.ts#LL65C13-L65C45.
By replacing this line with: `return require('@prisma/client/runtime');`, everything works as expected.

=> When I build the project with esbuild, `require(loadPath).Prisma;` returns `undefined`... hence my problem.

[...]

PR #461, opened 2023-06-05
fix: avoid return loaded prisma if undefined

Azzerty23:
https://github.com/zenstackhq/zenstack/issues/449

### DIFF
--- a/packages/runtime/src/enhancements/utils.ts
+++ b/packages/runtime/src/enhancements/utils.ts
@@ -62,7 +62,9 @@ function loadPrismaModule(prisma: any) {
     if (prisma._engineConfig?.datamodelPath) {
         const loadPath = path.dirname(prisma._engineConfig.datamodelPath);
         try {
-            return require(loadPath).Prisma;
+            const _prisma = require(loadPath).Prisma;
+            if (typeof _prisma !== 'undefined') return _prisma;
+            return require('@prisma/client/runtime');
         } catch {
             return require('@prisma/client/runtime');
         }


Azzerty23:
I would like to run more tests locally before updating the source code. Perhaps I should have copied the Prisma runtime as part of my build process. In order to perform these tests, I need to run a local copy of the @zenstackhq/runtime, but I have not been able to do so successfully. @ymc9, could you please explain how to achieve this? Having a dedicated contribute section in the documentation would be greatly appreciated. Thank you!

packages/runtime/src/enhancements/utils.ts
```diff
@@ -62,7 +62,9 @@ function loadPrismaModule(prisma: any) {
     if (prisma._engineConfig?.datamodelPath) {
         const loadPath = path.dirname(prisma._engineConfig.datamodelPath);
```
ymc9, line 63: Hi @Azzerty23 , thanks a lot for making this PR!

I'm trying to reproduce the issue with the repo you shared (express-api project) but things seem to work well from my environment. I added a line of log to print the `prisma._engineConfig?.datamodelPath` and it's like:

    `<root>/node_modules/.prisma/client/schema.prisma`

, and loading the module from that folder works (since it contains Prisma generated code there)...

So I'm still struggling to understand what could have caused failure on your side. Do you mind adding a similar line of log in your environment and seeing what it prints? Thanks!

ymc9:
> 

I usually push a locally built package to a private test npm registry for testing but that'll take some configuration. I think a simpler approach would be just do a `pnpm build` in the `packages/runtime` folder, and then copy over the build result from `dist` folder to the `node_modules` folder in your test project 😄.

Indeed a decent contribution guide is missing. I'll try to add it soon. Thanks for pushing for it!

packages/runtime/src/enhancements/utils.ts
Azzerty23: Thank you so much! I think I have understood the issue. In my build script, I copy the `schema.prisma` to my `dist/` folder to avoid this warning:
Capture d’écran 2023-06-05 à 14 45 21
Solution from: https://github.com/prisma/prisma/issues/10512#issuecomment-1254779466

Copying the `schema.prisma` to my `dist/` folder prevents `prisma._engineConfig?.datamodelPath` from recognizing the correct path for the Prisma client.

Indeed, when I console.log `prisma._engineConfig?.datamodelPath`, I get:
```
/Users/augustin/Documents/Dev/Test/zenstack-rest-turbo/apps/express-api/dist/schema.prisma
```
That's why I only had this issue with my built project. However, I don't understand why you didn't encounter the same problem with my repo. Ideally, I would like to find a solution to avoid the warning altogether but it's not blocking.

PR #464, opened 2023-06-05
fix: generate .zenstack with the same level of @zenstackhq

### LINKED ISSUE #462: .zenstack dir is not generated into correct location for pnpm project
Repro: https://github.com/Azzerty23/zenstack-rest-turbo

Run `pnpm zen:up` from the root folder, and you'll find `.zenstack` folder is generated into `/@zenstackhq/runtime/node_modules/`. The reason is pnpm can install packages locally under other packages. After finding the location of `@zenstackhq/runtime` package, I think we need to move up two levels to find the real `node_modules` location.

@jiashengguo

jiashengguo:
fix #462

### DIFF
--- a/packages/schema/src/plugins/plugin-utils.ts
+++ b/packages/schema/src/plugins/plugin-utils.ts
@@ -9,7 +9,9 @@ export const ALL_OPERATION_KINDS: PolicyOperationKind[] = ['create', 'update', '
  */
 export function getNodeModulesFolder(startPath?: string): string | undefined {
     startPath = startPath ?? process.cwd();
-    if (fs.existsSync(path.join(startPath, 'node_modules'))) {
+    if (startPath.endsWith('node_modules')) {
+        return startPath;
+    } else if (fs.existsSync(path.join(startPath, 'node_modules'))) {
         return path.join(startPath, 'node_modules');
     } else if (startPath !== '/') {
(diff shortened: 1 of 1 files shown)


packages/schema/src/plugins/plugin-utils.ts
```diff
@@ -25,7 +25,11 @@ export function getNodeModulesFolder(startPath?: string): string | undefined {
  */
 export function getDefaultOutputFolder() {
     // Find the real runtime module path, it might be a symlink in pnpm
-    const runtimeModulePath = require.resolve('@zenstackhq/runtime');
+    let runtimeModulePath = require.resolve('@zenstackhq/runtime');
```
ymc9, line 28: I think this willl resolve to ".../index.js". We need to get dirname and then walk up?

ymc9: > I think this willl resolve to "index.js". We need to get dirname and then walk up?

Maybe safer to check if it's a file? I'm not sure if the resolving behavior is always consistent ... I remember when I tried it's index.js

jiashengguo: yes, it is.  changed to use folder name to locate the folder

ymc9 (approved):

Issue #527, opened 2023-06-30
[Feature Request] Support returning "count" in RESTful API handler


ymc9:
Getting the total count during pagination is a very common requirement. Since JSON:API allows custom meta, we can probably use it to contain that information in the response. Also, there should be an option to opt-in for this feature to avoid a default performance hit.

chemitaxis:
hi @ymc9 I'm trying to build and test the project, but when I exec the` pnpm install` and `pnpm run build` I have some packages with error like this: 

![Screenshot 2023-06-30 at 19 06 05](https://github.com/zenstackhq/zenstack/assets/1420409/24f0ef7a-a1a4-40b2-b1df-5cb98aa3e942)

what i'm missing? thanks!

chemitaxis:
hi @ymc9 can you please give access to make a PR? thanks :)

ymc9:
Fixed by #529

PR #529, opened 2023-06-30
changes: enable total count meta on get

chemitaxis:
This PR solves https://github.com/zenstackhq/zenstack/issues/527 

I have added a new optional parameter on options to enable add metadata, it looks like:

```
{
      jsonapi: { version: '1.1' },
      links: {
        self: 'http://localhost/api/user',
        first: 'http://localhost/api/user?page%5Blimit%5D=5',
        last: 'http://localhost/api/user?page%5Boffset%5D=0',
        prev: null,
        next: null
      },
      data: [
        {
          type: 'user',
          id: 'user1',
          attributes: [Object],
          links: [Object],
          relationships: [Object]
        },
        {
          type: 'user',
          id: 'user2',
          attributes: [Object],
          links: [Object],
          relationships: [Object]
        }
      ],
      meta: { total: 2 }
    }
```


packages/server/src/api/rest/index.ts
```diff
@@ -618,14 +634,24 @@ class RequestHandler {
...
             return {
                 status: 200,
-                body: await this.serializeItems(type, entities, options),
+                body: body,
             };
         }
     }
 
+    private addTotalCountMeta(body: Partial<DataDocument<any>>, total: number | undefined): Partial<DataDocument<any>> {
+        if (total) {
```
ymc9, line 648: Shall we explicitly chek `undefined` so that `0` count can be included?

packages/server/src/api/rest/index.ts
```diff
@@ -599,9 +606,18 @@ class RequestHandler {
 
         if (limit === Infinity) {
             const entities = await prisma[type].findMany(args);
+
+            let body = await this.serializeItems(type, entities, { include });
+            // Count items if totalCount is requested using only where filter, it not depends the others filter
+            const total = this.options.totalCount
```
ymc9, line 612: After looking at the original code in the "else" branch, I want to change my idea about using an option to opt-in the "count" behavior 😄 because it's counted anyway when pagination is enabled. Sorry for the change of mind. Maybe we remove the option for now?

And for this branch of fetching with `Infinity` limit, since `findMany` is already exhausting the list, I think we don't need to do a separate `count` query, and can just use `entities.length` for it, right?

packages/server/src/api/rest/index.ts
ymc9: Also, shall we merge `total` into `body.meta` instead of overwriting it? Just thinking maybe in the future there'll already be a `meta` object in body containing other stuff.

ymc9 (changes requested):
Thanks a lot for the contribution @chemitaxis !

ymc9:
Btw, there're conflicts in the `pnpm-lock.yaml ` file, I guess cause by the latest changes in the "dev" branch and also a mismatch in the pnpm version. You can just revert the changes to that file and the CI should pass then.

chemitaxis:
@ymc9 changes applied! 👍

ymc9:
> @ymc9 changes applied! 👍 

Thanks! Looking good now. I'll merge when I get back to my laptop. It'll be included in the next release landing in a day or two.

ymc9 (approved):

ymc9:
Fixes #527

Issue #522, opened 2023-06-28
Validation errors when using true or false in the column name


abdullahahmeda:
**Description and expected behavior**
I have the following schema for true & false questions:
```prisma
model Question {
  id           Int            @id @default(autoincrement())
  text         String         @db.Text()
  trueText     String?        @db.Text()
  falseText    String?        @db.Text()
  answer       String         @db.Text()
}
```
When I run `npx zenstack generate` I get the following error:
```
Validation errors:
line 79: Expecting token of type '}' but found `true`. [true]
line 80: Expecting token of type '}' but found `false`. [false]
```
Also the zenstack vs code extension is linting the error:
![image](https://github.com/zenstackhq/zenstack/assets/23556281/163f9682-1760-4d8f-9f35-be6636f5ca7c)

**Environment:**

-   ZenStack version: 1.0.0-beta.2
-   Prisma version: 4.9.0
-   Database type: Postgresql

jiashengguo:
@abdullahahmeda   It's a bug for sure, thanks for reporting it!  We will make sure it is fixed in the next release.

abdullahahmeda:
Thank you.
For now I fixed this error by renaming `falseText` to `textForFalse` and `trueText` to `textForTrue`.

ymc9:
Fixed by #530

PR #530, opened 2023-06-30
fix: Validation errors when using true or false as prefix of id

jiashengguo:
change Boolean from terminal to rule

### DIFF
--- a/packages/language/src/zmodel.langium
+++ b/packages/language/src/zmodel.langium
@@ -39,7 +39,7 @@ Expression:
     LogicalExpr;
 
 LiteralExpr:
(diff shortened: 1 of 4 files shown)


packages/language/src/zmodel.langium
```diff
@@ -39,7 +39,7 @@ Expression:
     LogicalExpr;
 
 LiteralExpr:
-    value=(BOOLEAN | NUMBER | STRING);
+    value=(Boolean | NUMBER | STRING);
```
ymc9, line 42: Is the casing change relevant here?

ymc9:
Fixes #522

Issue #531, opened 2023-06-30
[Proposal] Prevent express middleware send response


chemitaxis:
Hello, the middleware functions should primarily be responsible for processing the request and preparing the response, but not actually sending it. Instead, once they have done their part in processing, they should pass the control to the next middleware in line using the next() function. 

This ensures a proper flow of control in the application, allowing other middleware functions or the main request handler to perform additional actions if necessary.

For instance, you may have middleware functions for tasks like data validation, logging, handle error, etc. After these tasks are completed, the response is then generated and sent in the main request handler, providing a central location for response generation.

This approach separates concerns more effectively, making the codebase more modular and maintainable. It also allows more flexibility for further processing of the response before it is sent. Plus, error handling can be more centralized and robust, as you can have specific error-handling middleware.

For testing, this design makes it much easier, as each middleware function can be tested independently for its specific task. Testing becomes straightforward as you just need to ensure that each middleware behaves correctly when receiving a request and calling next().

This way, we keep up with the principle of each part of the software doing one thing and doing it well. The middleware processes the request and prepares the response, the main request handler (or other middleware) sends the response, and the code becomes more maintainable.

What do you think?

ymc9:
Hi @chemitaxis , thanks for the detailed explanation I think it all makes very good sense.

For simple cases, I think it's easier to use if the middleware just directly sends the response. Shall we add an option to the middleware factory to control this behavior?

```ts
app.use(
    '/api/model',
    ZenStackMiddleware({
        getPrisma: (request) => withPresets(prisma, { user: getSessionUser(request) }),
        sendResponse: false
    })
);
```

When `sendResponse` is false, the prepared response object can be stored into `res.locals` (and then call `next()`) and be used by the subsequent middleware or request handlers:

```ts
app.all('/api/model/*', (req, res) => {
    const { status, body } = res.locals.zenstack;
    // further processing
    ...
    res.status(status).json(body);
});
```

Does this sound reasonable to you?

chemitaxis:
This sounds perfect. I can do another PR if you want

ymc9:
> This sounds perfect. I can do another PR if you want

That'll be great! Thank you!

chemitaxis:
@ymc9 here you have :) https://github.com/zenstackhq/zenstack/pull/533

ymc9:
Fixed by #533

PR #533, opened 2023-07-01
changes: add option to manage response middleware

chemitaxis:
This PR solves this issue proposal: https://github.com/zenstackhq/zenstack/issues/531

:)


chemitaxis:
hi @ymc9 I'm having this error on the CI:

`Error: packages/server build: src/express/middleware.ts(41,12): error TS7030: Not all code paths return a value.`

What do you think is the best approach to solve this warning?

chemitaxis:
I don't know why is failing the CI 😢  I thinks a problem installing a package:

` Command failed: npm install @tanstack/svelte-query`

ymc9:
> I don't know why is failing the CI 😢 I thinks a problem installing a package:
> 
> ` Command failed: npm install @tanstack/svelte-query`

I believe it's due to the recent release of svelte V4. Please use "dev" branch as the source branch for making changes in the future, and it should work then :D.

packages/server/src/express/middleware.ts
```diff
@@ -68,9 +83,18 @@ const factory = (options: MiddlewareOptions): Handler => {
                 zodSchemas,
                 logger: options.logger,
             });
-            response.status(r.status).json(marshalToObject(r.body, useSuperJson));
+            if (manageCustomResponse) {
```
ymc9, line 86: Just to make sure I understand how to use this. Since when `manageCustomResponse` is on the code here, simply returns the status and body, is it supposed to use the middleware like regular function call in a request handler then?

```js

const zenstack = ZenStackMiddleware({ manageCustomResponse: true, ... });

app.all('/api/model/*', async (req, res) => {
    const { status, body } = await zenstack(req, res);
    // more processing
    ...
    return res.status(status).json(body);
});

```

chemitaxis:
Hi, yes @ymc9, this is the idea but I don't know why this is not working... I have pushed a new temporal test before merging with the other express tests with `manageCustomResponse: true` but it's not working as expected... 🤔 the test never finish.

You can run only this test using this command from `server` folder

`./node_modules/.bin/jest tests/adapter/express-temp.test.ts --silent=false`

The `silent` arg is for checking `console.log` debugging. 

I am trying to figure out where is the problem...

ymc9:
> 

I think the problem is that when `manageCustomResponse` is set true, the middleware neither sends a response nor calls `next()`, so the processing is stuck.

I feel the general pattern of express middleware, if not sending a response, is not to return some data - I'm not sure how the returned data would be used by subsequent middleware or route handler. Instead, should it attach information to `req` or `res` object and call `next()`, and subsequent handlers fetch data there and decide what to do? `res.locals` looks like a good place to store custom data?

chemitaxis:
Hi @ymc9 yes, exactly, there was the problem... I found it but I don't know why, I don't have notifications on my inbox when I get mentioned... I have created a test, what do you think? Best.

rsaladocid:
good job @chemitaxis

rsaladocid:
this will be very useful @chemitaxis !

ymc9:
> 

It looks great now! I'll merge it and publish a new version soon. Thank you!

ymc9 (approved):

PR #548, opened 2023-07-04
chore: rename rest-api's option from manageCustomResponse to sendResponse

ymc9:
@chemitaxis I'm thinking of renaming the option for better clarity. Do you mind taking a look?


chemitaxis:
excellent @ymc9 :) the best name ever!! 😂

PR #549, opened 2023-07-04
fix: optimize the way how generated packages are loaded in test environment

ymc9:


packages/server/src/express/middleware.ts
```diff
@@ -14,14 +14,17 @@ export interface MiddlewareOptions extends AdapterBaseOptions {
...
+     * the response is stored in the `res.locals` object and then the middleware
+     * calls the `next()` function to pass the control to the next middleware.
+     * Subsequent middleware or request handlers need to make sure to send
+     * a response.
      *
-     * Defaults to false;
+     * Defaults to true;
      */
-    manageCustomResponse?: boolean;
+    sendResponse?: boolean;
```
chemitaxis, line 27: where are you setting the default value to true @ymc9 ?

packages/server/src/express/middleware.ts
```diff
@@ -86,7 +89,8 @@ const factory = (options: MiddlewareOptions): Handler => {
                 zodSchemas,
                 logger: options.logger,
             });
-            if (manageCustomResponse) {
+            if (sendResponse === false) {
```
chemitaxis, line 92: I think here we have a problem... `sendReponse` is `undefined`... I think this will be `true` always

ymc9: When not set `sendResponse` will be undefined, so `=== false` evaluates to `false`. I think it works. @chemitaxis

chemitaxis (approved):
Thanks for the clarification and PR!

Issue #542, opened 2023-07-03
[Feature Request] Support customizing requests sent by SWR


ymc9:
This is a peer feature request for SWR for #541

ymc9:
Fixed by #556

PR #556, opened 2023-07-05
feat: allow to use custom fetch with generated hooks

### LINKED ISSUE #541: [Feature Request] Support customizing requests sent by TanStack Query
Right now the hooks code generated by "@zenstackhq/tanstack-query" plugin sends requests directly by calling `fetch`, and users don't have a way to do customizations like adding extra headers. It's designed to have some global configuration mechanism for customization.

One possible solution is to support passing in a custom `fetch` function from the context:

```ts
function App({ Component, pageProps: { session, ...pageProps } }: AppProps) {
    return (
        <QueryClientProvider client={queryClient}>
            <ZenStackHooksProvider value={{ endpoint: '/api/model', fetch: myFetch }}>
                <AppContent />
            </ZenStackHooksProvider>
        </QueryClientProvider>
    );
}
```

Internally, the generated code checks if there's a `fetch` from the context, and if so, use it instead.

ymc9:
Fixes #541 #542

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
         "test-ci": "ZENSTACK_TEST=1 pnpm -r run test --silent",
         "lint": "pnpm -r lint",
         "publish-all": "pnpm --filter \"./packages/**\" -r publish --access public",
-        "publish-preview": "pnpm --filter \"./packages/**\" -r publish --registry http://localhost:4873"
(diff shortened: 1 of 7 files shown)


ymc9:
Hi @chemitaxis , here's the changes for allowing custom "fetch", FYI.

Issue #609, opened 2023-07-31
Policy check incorrectly requires "update" permission when connecting a self-relation


ymc9:
**Schema**

```prisma
        model User {
            id String @id @default(cuid())
            comments Comment[]
        }

        model Comment {
            id                 String      @id @default(cuid())
            parentCommentId    String? 
            replies            Comment[]   @relation("CommentToComment")
            parent             Comment?    @relation("CommentToComment", fields: [parentCommentId], references: [id])
            comment            String
            author             User        @relation(fields: [authorId], references: [id])
            authorId           String      
            
            @@allow('read,create', true)
            @@allow('update,delete', auth() == author)
        }    
```

**Description**

When connecting a new comment to parent, currently enhanced Prisma requires the parent item to be updatable, which is incorrect because making a connection only updates foreign key on the child side, not the parent.

PR #610, opened 2023-07-31
fix: incorrect relation owner analysis

ymc9:
Should check if a `@relation` attribute really has `fields` argument to determine if it holds foreign key.


ymc9:
Fixes #609

PR #617, opened 2023-08-05
merge dev to main

ymc9:
- refactor: don't use auxiliary fields for policy checking anymore (fields are still generated and will be removed in a future release).
- more complete logging for queries sent to Prisma


ymc9:
Fixes #562

Issue #476, opened 2023-06-09
 [Feature Request] Support field comparison in access policy rules


ymc9:
Today policy rules allow you to do comparisons but only one field can be involved in an expression, which means with:

```prisma
model Foo {
  x Int
  y Int
}
```

`@@allow(x > 0)` is valid, but `@@allow(x > y)` is rejected at compile time.

Prisma has a preview feature called [fieldReference](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#compare-columns-in-the-same-table) which allows you to compare between fields in the same table. We can use that to relax some of the restrictions here.

Issue #634, opened 2023-08-16
Conditions hoisted from nested read overwrites toplevel where conditions


ymc9:
**Version**
1.0.0-Beta.18

**Schema**

```prisma
model User {
    id String @id @default(uuid())
    email String @unique
    password String? @password @omit
    name String?
    orgs Organization[]
    posts Post[]
    groups Group[]
    comments Comment[]
    // can be created by anyone, even not logged in
    @@allow('create', true)
    // can be read by users in the same organization
    @@allow('read', orgs?[members?[auth() == this]])
    // full access by oneself
    @@allow('all', auth() == this)
}

model Organization {
    id String @id @default(uuid())
    name String
    members User[]
    post Post[]
    groups Group[]
    comments Comment[]

    // everyone can create a organization
    @@allow('create', true)
    // any user in the organization can read the organization
    @@allow('read', members?[auth() == this])
}

abstract model organizationBaseEntity {
    id String @id @default(uuid())
    createdAt DateTime @default(now())
    updatedAt DateTime @updatedAt
    isDeleted Boolean @default(false) @omit
    isPublic Boolean @default(false)
    owner User @relation(fields: [ownerId], references: [id], onDelete: Cascade)
    ownerId String
    org Organization @relation(fields: [orgId], references: [id], onDelete: Cascade)
    orgId String
    groups Group[]

    // when create, owner must be set to current user, and user must be in the organization
    @@allow('create', owner == auth() && org.members?[this == auth()])
    // only the owner can update it and is not allowed to change the owner
    @@allow('update', owner == auth() && org.members?[this == auth()] && future().owner == owner)
    // allow owner to read
    @@allow('read', owner == auth())
    // allow shared group members to read it
    @@allow('read', groups?[users?[this == auth()]])
    // allow organization to access if public
    @@allow('read', isPublic && org.members?[this == auth()])
    // can not be read if deleted
    @@deny('all', isDeleted == true)
}

model Post extends organizationBaseEntity {
    title String
    content String
    comments Comment[]
}

model Comment extends organizationBaseEntity {
    content String
    post Post @relation(fields: [postId], references: [id])
    postId String
}

model Group {
    id String @id @default(uuid())
    name String
    users User[]
    posts Post[]
    comments Comment[]
    org Organization @relation(fields: [orgId], references: [id])
    orgId String

    // group is shared by organization
    @@allow('all', org.members?[auth() == this])
}
```

**Query**

```ts
const result = await client.comment.findMany({
[...]

Issue #632, opened 2023-08-15
[Bug] Zod breaking error: Two enums one model


amcmillion:
**Description and expected behavior**
Using an enum in two separate fields of a model breaks Zod generation:

```
⠙ Running plugin ZodError compiling generated code:
node_modules/.zenstack/zod/models/TwoEnumsOneModelTest.schema.ts:4:10 - error TS2300: Duplicate identifier 'InventoryUnitSchema'.

4 import { InventoryUnitSchema } from '../enums/InventoryUnit.schema';
           ~~~~~~~~~~~~~~~~~~~
node_modules/.zenstack/zod/models/TwoEnumsOneModelTest.schema.ts:5:10 - error TS2300: Duplicate identifier 'InventoryUnitSchema'.

5 import { InventoryUnitSchema } from '../enums/InventoryUnit.schema';
           ~~~~~~~~~~~~~~~~~~~

✖ Running plugin Zod
: Error compiling generated code
```

**Environment (please complete the following information):**

-   ZenStack version: 1.0.0-beta.18
-   Prisma version: 5.1.1
-   Database type: Postgresql

### Test setup
```
enum InventoryUnit {
  DIGITAL // no limit but we can still track sales
  FL_OZ
  GRAMS
  MILLILITERS
  OUNCES
  UNIT
  UNLIMITED
}

model TwoEnumsOneModelTest {
  id String @id @default(cuid())

  inventoryUnit   InventoryUnit @default(UNIT)
  inputUnit       InventoryUnit @default(UNIT)
}
```

ymc9:
Thanks for reporting this @amcmillion ! I'll look into it and make a fix soon.

ymc9:
Hey @amcmillion , the issue has been fixed in the latest Beta-19 release.

PR #637, opened 2023-08-16
merge dev to main

ymc9:
- feat: add support for comparing fields in the same model
- fix: duplicated zod schema imported when there're multiple fields with the same enum type
- fix: conditions hoisted from nested read overwrites top-level where conditions


ymc9:
Fixes #476 #632 #634

PR #638, opened 2023-08-19
feat: field-level access control

ymc9:


ymc9:
Implementing field-level access control: for "read" and "update" only, with field-level attributes `@allow` and `@deny`.

For "read": when a field has policy rules, it'll be omitted from the read if any deny rule evaluates to "true" or none of the allow rule evaluates to "true".

For "update": when a field has policy rules, "update" will be rejected or filtered if any deny rule evaluates to "true" or none of the allow rule evaluates to "true".

PR #651, opened 2023-08-26
feat: more flexible "createRouter" typings

mateus-p:
Let the user dictate the typing of the router and the procedure, rather than just dictating the configuration. This would allow creating more complex types (such as middleware types).

**Motivation**
image
image


ymc9:
Thank you for making the PR @mateus-p ! It looks great, and I'm merging it now. Will include it in the next release.

ymc9 (approved):

PR #652, opened 2023-08-27
Add package.exports to @zenstack/server

krist7599555:
Many bundler with mixed `esm`/`cjs` will not work without `exports` field in `package.json`

image


ymc9:
Thanks for making this PR @krist7599555 !

I'm thinking maybe we should really generate cjs/esm dual format during build (like already done in the runtime package with tsup). I'm not really an expert on bundlers. Do you feel that to be necessary?

krist7599555:
`@zenstack/server` will mostly be use with framework like `next`, `sveltekit` which have another step to build production. And output is likely to be `esm` with `import` syntax.

So, I suggest using `tsup` with `minify=false` is the way.


If @ymc9 agree I welcomely to create pull request about **Migrate 
`@zenstack/server` to tsup** for you. (breaking!! change because npm folder structure will nested 1 level)


from
```
package.json
api/base.js
api/rest/index.js
api/rpc/index.js
api/utils.js
express/index.js
express/middleware.js
fastify/index.js
fastify/plugin.js
```

to
```
package.json
dist/api/base.js
dist/api/rest/index.js
dist/api/rpc/index.js
dist/api/utils.js
dist/express/index.js
dist/express/middleware.js
dist/fastify/index.js
dist/fastify/plugin.js
```

ymc9:
> 

The "package.json" file has a publishConfig setting:
```json
    "publishConfig": {
        "directory": "dist",
        "linkDirectory": true
    }
```
So when published, the content of "dist" will be deployed instead of the root folder. I think we don't need the "dist" level in the "exports" section.

I'm not even sure if cjs/ems dual output is necessary. If framework bundlers are fine with CJS we can probably keep it as is. Just curious, with what framework did you initially run into issues with the "exports" section missing?

packages/server/package.json
```diff
@@ -55,5 +55,21 @@
         "supertest": "^6.3.3",
         "ts-jest": "^29.0.5",
         "typescript": "^4.9.4"
+    },
+    "exports": {
+        "./api/base": "./api/base.js",
```
ymc9, line 60: I think we can selectively export here. The following should be enough:

```
    "exports": {
        "./api/rest": "./api/rest/index.js",
        "./api/rpc": "./api/rpc/index.js",
        "./express": "./express/index.js",
        "./fastify": "./fastify/index.js",
        "./next/app-route-handler": "./next/app-route-handler.js",
        "./next": "./next/index.js",
        "./next/pages-route-handler": "./next/pages-route-handler.js",
        "./sveltekit": "./sveltekit/index.js",
        "./types": "./types.js"
    }
```

ymc9 (approved):

PR #654, opened 2023-08-27
feat: flexible 'createRouter' typings

mateus-p:
This is the second attempt to create a more flexible typing for 'createRouter', as the first one didn't have the expected result (it just didn't work). 

This time I made myself in charge of performing a double check on the generated typing

image
image

The only possible problem I noticed here is the use of `.strict()` in Zod schemas, as this can generate the following bug:

> Zod throws an error when parsing the schema as it is a strict schema, but the entry has extra keys from the middleware.

**Possible solution (based on the example above)** 
image

Or we can simply stop using `.strict()`


mateus-p:
BTW, I have no idea why the tests are failing (I'm using my fork on the project I'm working on and everything is going great).
Also, the jest logs didn't help me.

ymc9:
Hi @mateus-p , thanks for continuing to work on this. Do you mind sharing the trpc project that exposes the original problem? I just wanted to make sure I fully understand the problem. Thank you!

mateus-p:
```ts
import { Context } from "./context";
import { createRouter } from "./generated/trpc/routers";
import { TRPCError, initTRPC } from "@trpc/server";
import z from "zod";
import { compare } from "./hash";
import { PrismaContext } from "./prisma";
import { enhance } from "@zenstackhq/runtime";

function calcMaxAge(): Date {
  return new Date(Date.now() + Number(process.env.SESSION_MAX_AGE!));
}

const userSessionInput = z.object({ __userSessionId: z.string().optional() });
type UserSessionInput = z.infer<typeof userSessionInput>;

export default function route({ prisma }: PrismaContext) {
  const t = initTRPC.context<Context>().create();

  const procWithUserSessionInput = t.procedure.input(userSessionInput);

  const withAuthProc = procWithUserSessionInput.use(
    t.middleware(async (opts) => {
      const { __userSessionId } = opts.input as UserSessionInput;

      console.log("mid input", opts.input);

      delete (opts.input as UserSessionInput).__userSessionId;

      const session = __userSessionId
        ? await prisma.session.findUnique({
            where: {
              id: __userSessionId,
            },
            include: { owner: true },
          })
        : null;

      if (!session) {
        return opts.next({
          ctx: { prisma: enhance(prisma), session: null } as unknown as Context,
        });
      }

      if (session.maxAge.getTime() < Date.now()) {
        await prisma.session.delete({ where: { id: session.id } });

        throw new TRPCError({
          code: "FORBIDDEN",
          message: "invalid session: max age reached",
        });
      }

      return opts.next({
        ctx: {
          prisma: enhance(prisma, { user: session.owner }),
          session,
        } as unknown as Context,
      });
    })
  );

// With the current 'createRouter' typing 'withAuthProc' is not allowed
// since a middleware is not fully compatible with the pure 't.procedure'
  const appRoutes = createRouter(t.router, withAuthProc);

  const specialRoutes = t.router({
    login: t.procedure
      .input(
        z
          .object({
            email: z.string(),
            password: z.string(),
            location: z.string().optional(),
          })
          .strict()
      )
      .mutation(async ({ ctx, input }) => {
        const { req } = ctx;

        const user = await prisma.user.findUnique({
          where: { email: input.email },
        });

        if (!user)
          throw new TRPCError({ code: "NOT_FOUND", message: "user not found" });

        if (!compare(input.password, user.password))
          throw new TRPCError({
            code: "BAD_REQUEST",
            message: "wrong password",
          });

        const newSession = await prisma.session.create({
          data: {
            location: input.location || "%n%",
            maxAge: calcMaxAge(),
            userAgent: req.headers["user-agent"] || "%n%",
            owner: { connect: user },
          },
        });

        return newSession.id;
      }),

    logout: withAuthProc.mutation(async ({ ctx }) => {
      if (!ctx.session) return false;

      await prisma.session.delete({ where: { id: ctx.session.id } });

      return true;
    }),
  });

  return { t: t, router: t.mergeRouters(appRoutes, specialRoutes) };
}

```

packages/plugins/trpc/src/generator.ts
```diff
@@ -89,35 +89,85 @@ function createAppRouter(
...
             typeof createRouterFactory<Config>
         >;
         
-        export type ProcBuilder<Config extends BaseConfig> = ReturnType<
-            typeof createBuilder<Config>
-        >;
+        ${
+            /** this is needed in order to prevent type errors between a procedure and a middleware-extended procedure */ ''
+        }
+        export type ProcBuilder<Config extends BaseConfig> = ProcedureBuilder<{
```
mateus-p, line 123: And this whole workaround is to "extract and merge" the types of I/O coming from the procedure with those from the schemas.

packages/plugins/trpc/src/generator.ts
```diff
@@ -89,35 +89,85 @@ function createAppRouter(
...
+            : never;
+          
+        ${/** workaround to avoid creating 'typeof unsetMarker & object' on the procedure output */ ''}
+        type Join<A, B> = A extends symbol ? B : A & B;
+
+        ${
+            /** you can name it whatever you want, but this is to make sure that 
+                the types from the middleware and the procedure are correctly merged */ ''
+        }
+        export type ProcReturns<
```
mateus-p, line 147: And this whole workaround is to "extract and merge" the types of I/O coming from the procedure with those from the schemas.

ymc9: Got it, thanks for the context! Let me look into why the tests are failing.

packages/plugins/trpc/src/helpers.ts
```diff
@@ -21,14 +21,24 @@ export function generateProcedure(
         writer.write(`
         ${opType}: procedure.input(${typeName}).query(({ctx, input}) => checkRead(db(ctx).${lowerCaseFirst(
             modelName
-        )}.${prismaMethod}(input as any))),
+        )}.${prismaMethod}(input as any))) as ProcReturns<
+            "query",
+            Proc,
+            (typeof ${modelName}InputSchema)["${opType.replace('OrThrow', '')}"],
```
ymc9, line 27: I think the test failure was due to missing a `upperCaseFirst` call here. Prisma has some crazy casing behavior in generated types, and at the zod generation time, those are canonicalized to upper-case-first.

packages/plugins/trpc/src/helpers.ts
```diff
@@ -21,14 +21,24 @@ export function generateProcedure(
...
     } else if (procType === 'mutation') {
         // the cast "as any" is to circumvent a TS compiler misfired error in certain cases
         writer.write(`
         ${opType}: procedure.input(${typeName}).mutation(async ({ctx, input}) => checkMutate(db(ctx).${lowerCaseFirst(
             modelName
-        )}.${prismaMethod}(input as any))),
+        )}.${prismaMethod}(input as any))) as ProcReturns<
+                "mutation",
+                Proc,
+                (typeof ${modelName}InputSchema)["${opType.replace('OrThrow', '')}"],
```
ymc9, line 39: And here

packages/plugins/trpc/src/generator.ts
```diff
@@ -89,35 +89,85 @@ function createAppRouter(
...
 
         export type RouterFactory<Config extends BaseConfig> = ReturnType<
             typeof createRouterFactory<Config>
         >;
         
-        export type ProcBuilder<Config extends BaseConfig> = ReturnType<
-            typeof createBuilder<Config>
-        >;
+        ${
+            /** this is needed in order to prevent type errors between a procedure and a middleware-extended procedure */ ''
```
ymc9, line 121: Thanks a lot for making this great fix! It took me a while to comprehend and it looks very good now!

ymc9:
> This is the second attempt to create a more flexible typing for 'createRouter', as the first one didn't have the expected result (it just didn't work).
> 
> This time I made myself in charge of performing a double check on the generated typing
> 
> image image
> 
> The only possible problem I noticed here is the use of `.strict()` in Zod schemas, as this can generate the following bug:
> 
> > Zod throws an error when parsing the schema as it is a strict schema, but the entry has extra keys from the middleware.
> 
> **Possible solution (based on the example above)** image
> 
> Or we can simply stop using `.strict()`

I thought having `.strict()` there could help catch extra properties that users pass to prisma input (maybe typo). Deleting extra fields from middleware seems OK to me. Shall we just keep it that way for now and see if we get feedback from other people in the future?

ymc9 (changes requested):

packages/plugins/trpc/src/generator.ts
```diff
@@ -89,35 +89,85 @@ function createAppRouter(
...
             moduleSpecifier: '@trpc/server',
         },
         {
-            namedImports: ['PrismaClient'],
+            namedImports: ['type PrismaClient', 'type Prisma'],
             moduleSpecifier: prismaImport,
-            isTypeOnly: true,
         },
         {
-            namedImports: ['createRouterFactory', 'AnyRouter'],
```
ymc9, line 101: The `AnyRouter` import needs to be kept as it's used in extra code generation when the "generateClientHelpers" option is enabled (will cause test failure). We can change it into a type import, though.

mateus-p:
> > This is the second attempt to create a more flexible typing for 'createRouter', as the first one didn't have the expected result (it just didn't work).
> > This time I made myself in charge of performing a double check on the generated typing
> > image image
> > The only possible problem I noticed here is the use of `.strict()` in Zod schemas, as this can generate the following bug:
> > > Zod throws an error when parsing the schema as it is a strict schema, but the entry has extra keys from the middleware.
> > 
> > 
> > **Possible solution (based on the example above)** image
> > Or we can simply stop using `.strict()`
> 
> I thought having `.strict()` there could help catch extra properties that users pass to prisma input (maybe typo). Deleting extra fields from middleware seems OK to me. Shall we just keep it that way for now and see if we get feedback from other people in the future?

Yes of course!

packages/plugins/trpc/src/helpers.ts
mateus-p:
```ts
aggregate: procedure
            .input(PubicFieldEntryInputSchema.aggregate)
            .query(({ ctx, input }) => checkRead(db(ctx).pubicFieldEntry.aggregate(input as any))) as ProcReturns<
            'query',
            Proc,
            (typeof PubicFieldEntryInputSchema)['aggregate'],
            ReturnType<PrismaClient['pubicFieldEntry']['aggregate']>
        >,
```

This is the generated code and it is working for me. Where exactly is `upperCaseFirst` missing?

packages/plugins/trpc/src/helpers.ts
mateus-p: done ✔️

packages/plugins/trpc/src/helpers.ts
mateus-p: done ✔️

mateus-p: weird...

ymc9 (approved):

ymc9:
Merged! 🙏 

@mateus-p have you joined ZenStack discord server? I want to add you to the contributors group there 😄

PR #672, opened 2023-09-04
merge dev to main

ymc9:


ymc9:
- trpc plugin: more flexible typing for the generated "createRouter" function
- zod plugin: export Prisma input object schemas from "@zenstackhq/runtime/zod"
- Added "exports" field to package.json of "@zenstackhq/server"
- zod/openapi plugin: fixed generation error when "fullTextSearch" preview feature is enabled
- Fixed the problem that small number literal is converted to e-notation when generating Prisma schema
- Fixed zod schema generation for decimal fields
- Added fluent API support
- trpc plugin: added the missing "count" router
- Support string literal escaping in ZModel
- Fixed policy generation error when "this" is used in field-level policies

Issue #676, opened 2023-09-06
Generated Zod types are lost with @@validate 


tlancina:
Hello! Generated model types for Zod change to ZodEffects and lose most of their helpful information when using a validate rule. This came up because I'd like to enforce some string literals and can't use enum because some of them have disallowed chars (I'd like to enforce that the field/form is one of `"a" | "b" | "c'd"` - if there's another way to do this please let me know.)

Using https://github.com/ymc9/zenstack-form-validation.
The lines in question are
```
  // sqlite doesn't support enum, you should use enum in a real application
  @@validate(beverage in ['SODA', 'COFFEE', 'BEER', 'COCKTAIL'], 'Please choose a valid beverage')

  @@validate(adult || beverage in ['SODA', 'COFFEE'], 'You must be an adult to drink alcohol')
  ```

`.zenstack/zod/models/Signup.schema.d.ts` with `@@validate`:
```ts
import { z } from 'zod';
export declare const SignupSchema: z.ZodEffects<z.ZodEffects<z.ZodType<any, z.ZodTypeDef, any>, any, any>, any, any>;
```

Without validate:
```ts
import { z } from 'zod';
export declare const SignupSchema: z.ZodObject<{
    id: z.ZodNumber;
    name: z.ZodString;
    email: z.ZodString;
    adult: z.ZodBoolean;
    beverage: z.ZodString;
    createdAt: z.ZodDate;
    updatedAt: z.ZodDate;
}, "strip", z.ZodTypeAny, {
    id: number;
    name: string;
    email: string;
    adult: boolean;
    beverage: string;
    createdAt: Date;
    updatedAt: Date;
}, {
    id: number;
    name: string;
    email: string;
    adult: boolean;
    beverage: string;
    createdAt: Date;
    updatedAt: Date;
}>;
```

**Environment (please complete the following information):**
Using https://github.com/ymc9/zenstack-form-validation. I removed package-lock because there's a reference to localhost, and updated to latest zenstack (beta 21).

Happy to contribute if you can point me in the right direction.

tlancina:
[...]

PR #678, opened 2023-09-07
fix: more precise Zod refinement types

tlancina:
Addresses https://github.com/zenstackhq/zenstack/issues/676.

I just edited this in Github after testing it out by hand, so I'm not 100% sure if it works.


ymc9:
Merged. Thank you for the contribution @tlancina !

Issue #774, opened 2023-10-20
[Feature Request] Using model not named "User" as typing of `auth()`


ymc9:
Today the CLI is hard-coded to resolve `auth()` call to `User` model. We can introduce a new `@@auth` attribute and let developers use it to annotate the model that he wants to use for backing `auth()`.

Related discussion here: https://github.com/zenstackhq/zenstack/issues/402#issuecomment-1771871962

PR #787, opened 2023-10-24
feat: add @@auth option for declaring auth model

jasonmacdonald:
This is by no means complete yet, but I want to check that I'm on the right track. I wasn't able to find a simple way to pull the attribute value during the linker phase, so I added two utility functions - but there's probably a better way.

I'll need to figure out how to test in VScode Extension as well, as this is my first time delving into that.


packages/schema/src/language-server/zmodel-linker.ts
```diff
@@ -278,9 +278,11 @@ export class ZModelLinker extends DefaultLinker {
                 const model = getContainingModel(node);
 
                 if (model) {
-                    const userModel = getAllDeclarationsFromImports(this.langiumDocuments(), model).find(
-                        (d) => isDataModel(d) && d.name === 'User'
-                    );
+                    const dataModel = this.getContainingDataModel(node);
```
ymc9, line 281: I think we need to find data model with `@@auth` attribute from all data model declarations, but not just the container of the current `auth()` invocation. Just like how we used to find the model named "User" in the code below this.

packages/schema/src/res/stdlib.zmodel
```diff
@@ -376,6 +376,11 @@ attribute @@deny(_ operation: String, _ condition: Boolean)
  */
 attribute @deny(_ operation: String, _ condition: Boolean)
 
+/**
+ * Defines the model to use when when checking access policies.
+ */
+attribute @@auth(_ modelName: String)
```
ymc9, line 382: Shall we just declare this attribute without a parameter? My thought is to use it like the following:

```prisma
model Foo {
    ...
    @@auth()
}
```

packages/schema/src/language-server/zmodel-linker.ts
```diff
@@ -278,9 +278,11 @@ export class ZModelLinker extends DefaultLinker {
...
+                    const customAuthModel = dataModel && getModelAttributeValue('@@auth', dataModel);
```
ymc9, line 282: If we declare the attribute without a parameter, we can simply find it by checking attribute name. There's a `hasAttribute` helper from `@zenstackhq/sdk` for doing this.

ymc9 (changes requested):
Hi @jasonmacdonald , thanks a lot for making this PR. I've added some comments. Please let me know if you have time to continue working on it, and if not, I'll be happy to take it over and make tunings as needed.

For testing, I think you can check if `zenstack generate` works properly. If so, the VSCode extension will just work (since the CLI runs more check phases than the VSCode extension). The simplest way to test the CLI is to run `pnpm build` in "schema" folder, and copy over the content of "schema/dist" to "node_modules/zenstack" of an existing project.

packages/schema/src/language-server/zmodel-linker.ts
jasonmacdonald: ok, let me try that

packages/schema/src/res/stdlib.zmodel
jasonmacdonald: Ah, so you were thinking the attribute would declare what Model is the "User" model. Ok, for some reason, I thought you wanted it to be declared in each place it was used, but that makes more sense.

jasonmacdonald:
> Hi @jasonmacdonald , thanks a lot for making this PR. I've added some comments. Please let me know if you have time to continue working on it, and if not, I'll be happy to take it over and make tunings as needed.
> 
> For testing, I think you can check if `zenstack generate` works properly. If so, the VSCode extension will just work (since the CLI runs more check phases than the VSCode extension). The simplest way to test the CLI is to run `pnpm build` in "schema" folder, and copy over the content of "schema/dist" to "node_modules/zenstack" of an existing project.

I'll take another stab at it. I thought I might be going in the wrong direction, hence posting this early. I think I understand your thoughts on it now.

jasonmacdonald:
@ymc9 Changed the attribute to be placed on the target model.

ymc9:
> @ymc9 Changed the attribute to be placed on the target model.

Thank you, @jasonmacdonald ! It looks good to me now. I'll merge it soon and include it in the upcoming v1.2.0 release.

ymc9 (dismissed):

ymc9 (approved):

ymc9:
Fixes #774

Issue #814, opened 2023-11-08
Field level policies don't apply to relations on read


tlancina:
-   ZenStack version: 1.2.0

It seems that https://github.com/zenstackhq/zenstack/blob/ac3206b42c2f4dfae25b5ed52df764d7256d8585/packages/runtime/src/enhancements/policy/policy-utils.ts#L1099 means that instead of calling `this.hasFieldLevelPolicy(model)` on the nested model it is using the `hasFieldLevelPolicy` of the parent, so no field level policies are applied.

We are using the pattern described in https://github.com/zenstackhq/zenstack/issues/809 to `@deny` all fields except one on our User model, and this appears to be leaking our entire user model when it is accessed via relation/`include` (edit - sorry I should emphasize that this is largely a non-issue since we are only selecting the field we need, I just noticed that in the default case the policies won't apply).

Happy to provide a reproduction if necessary.

Also - congrats on 1K stars! 🤩

ymc9:
Hi @tlancina , thanks for filing this issue and submitting the PR! It looks like a bug to me. It'll be great if you can provide a repro, and I'd like to include it as a regression test. I'll also add independent cases to cover it as well.

Since it's not blocking as far as I understand, I think we can merge it to "dev" branch and contain it in the next 1.3 release (around 11/15)?

I'm super happy with hitting 1K start too 😄. Thank you for the great support all along!

ymc9:
I've managed to get a repro and added a regression test to the PR. Will merge it when CI passes. Thanks!

tlancina:
Ah I'm sorry you had to do that, I was away from my computer until now. Thanks for cleaning it up!

ymc9:
> Ah I'm sorry you had to do that, I was away from my computer until now. Thanks for cleaning it up!

No worries. Thanks for making the fix!

PR #816, opened 2023-11-08
Apply field level policies to relations

tlancina:
Proposed fix for https://github.com/zenstackhq/zenstack/issues/814.

I'm not sure if this is correct, or if `fieldInfo.type` will be the correct description of "model" always.


ymc9:
Fixes #814

PR #904, opened 2023-12-23
feat: JetBrains plugin for ZModel

ymc9:


Lordfirespeed:
You may want to consider putting this into its own repository, which would allow you to utilise features like GitHub Releases (and workflows from the [IntelliJ plugin template](https://github.com/JetBrains/intellij-platform-plugin-template)).

ymc9:
> You may want to consider putting this into its own repository within the `zenstackhq` org, which would allow you to utilise features like GitHub Releases (and workflows from the [IntelliJ plugin template](https://github.com/JetBrains/intellij-platform-plugin-template)).

I struggled with this decision ... I think it makes good sense to let it reside in its own repo, which also makes it easier for people to contribute. The main problem is it depends on the build output from the main repo for the language server and textmate grammar. With a separate repo, we need to somehow share these assets.

ymc9:
> > You may want to consider putting this into its own repository within the `zenstackhq` org, which would allow you to utilise features like GitHub Releases (and workflows from the [IntelliJ plugin template](https://github.com/JetBrains/intellij-platform-plugin-template)).
> 
> I struggled with this decision ... I think it makes good sense to let it reside in its own repo, which also makes it easier for people to contribute. The main problem is it depends on the build output from the main repo for the language server and textmate grammar. With a separate repo, we need to somehow share these assets.

If we put them into the same repo, then the releases will follow the same version and cadence. As for GitHub workflows, I guess we can add a separate CI configure just for build and publish stuff related to gradle and jetbrains?

Lordfirespeed:
> The main problem is it depends on the build output from the main repo for the language server and textmate grammar. With a separate repo, we need to somehow share these assets.

You can either 
1. Publish the language server as a package on NPM and depend on that
2. use [Git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules)

Lordfirespeed:
> If we put them into the same repo, then the releases will follow the same version and cadence. As for GitHub workflows, I guess we can add a separate CI configure just for build and publish stuff related to gradle and jetbrains?

This makes me feel uncomfortable :joy: But I don't think that feeling is necessarily correct.

I think my gut instinct is that the JetBrains plugin should be agnostic of the ZenStack language version. 

For example, if you have an old version of ZenStack installed then the plugin needs to be able to tell you "Hey, this language feature is not available!"

ymc9:
> > If we put them into the same repo, then the releases will follow the same version and cadence. As for GitHub workflows, I guess we can add a separate CI configure just for build and publish stuff related to gradle and jetbrains?
> 
> This makes me feel uncomfortable 😂 But I don't think that feeling is necessarily correct.
> 
> I think my gut instinct is that the JetBrains plugin should be agnostic of the ZenStack language version.
> 
> For example, if you have an old version of ZenStack installed then the plugin needs to be able to tell you "Hey, this language feature is not available!"
> 
> But I am not sure that that would be a real problem when releasing the plugin under the same version numbers as the language.

I agree with you that'll be how the extension/plugin works, but we're not at that granularity yet. The current assumption is that we don't remove language features, so a newer version of the extension always works with an older version of ZenStack, but it doesn't complain when you use a feature not available in your toolkit version ... I think we should revisit the entire strategy, especially if we make language-level breaking changes in V2 and see how to handle semver in VSCode/JetBrains.

Before that, it's probably OK to let them all huddle in the same repo 😅

ymc9:
> > The main problem is it depends on the build output from the main repo for the language server and textmate grammar. With a separate repo, we need to somehow share these assets.
> 
> 
> 
> You can either 
> 
> 1. Publish the language server as a package on NPM and depend on that
> 
> 2. use [Git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules)
> 
> 
> 
> I think the 2nd option would do pretty well. Let me know if you would like a hand with that.

Sorry I missed this comment previously. Yes, git submodule could be a solution, but we'll need to check in the compiled/bundled result if we don't want to recompile them when building the extension repo. Also I always felt submodule to be a pain to manage 😄. Npm dependency may be a simpler solution in the future.

Lordfirespeed:
I do have to agree with you on the Git submodules thing 😂😂

It is also possible to use a Git/GitHub repository as an NPM dependency.

https://docs.npmjs.com/cli/v8/configuring-npm/package-json#git-urls-as-dependencies

ymc9:
Fixes #904

PR #946, opened 2024-01-12
fix: typing generated for options parameter in the hooks method

### LINKED ISSUE #945: [svelte-query] Wrong typing generated for options parameter in the hooks method
Today query hooks are generated like:

```ts
export function useFindManyUser<T extends Prisma.UserFindManyArgs>(
    args?: Prisma.SelectSubset<T, Prisma.UserFindManyArgs>,
    options?: Omit<
        QueryOptions<
            Array<Prisma.UserGetPayload<T> & { $optimistic?: boolean }>
        >,
        'queryKey'
    >,
    optimisticUpdate: boolean = true
);
```

The `QueryOptions` type for the `options` parameter is too generic and is missing a bunch of useful fields like "refetchInterval". We should generate `CreateQueryOptions` instead.

bbozzay:
fixes #945

### DIFF
--- a/packages/plugins/tanstack-query/src/generator.ts
+++ b/packages/plugins/tanstack-query/src/generator.ts
@@ -569,7 +569,7 @@ function makeBaseImports(target: TargetFramework, version: TanStackVersion) {
         case 'svelte':
             return [
                 `import { derived } from 'svelte/store';`,
-                `import type { MutationOptions, QueryOptions, CreateInfiniteQueryOptions } from '@tanstack/svelte-query';`,
+                `import type { MutationOptions, CreateQueryOptions, CreateInfiniteQueryOptions } from '@tanstack/svelte-query';`,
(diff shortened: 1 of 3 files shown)


ymc9:
Thanks for working on this @bbozzay !

Inside the generator code there are also cases where `QueryOptions` instead of `CreateQueryOptions` are used for "svelte-query" generation target. Do you mind getting those changed as well? Otherwise the signature of the generated hooks methods still have the narrow options interface.

https://github.com/bbozzay/zenstack/blob/tanstack-query-svelte-options-type-fix/packages/plugins/tanstack-query/src/generator.ts

ymc9 (approved):

ymc9:
Cool! I'll get it merged when CI passes. It'll be included in the next release early next week.

bbozzay:
thank you!

Issue #947, opened 2024-01-14
empty enum array default


chunkerchunker:
**Description and expected behavior**

The following schema fails compilation for zenstack but compiles successfully for prisma:

```
datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}

generator client {
  provider = "prisma-client-js"
}

model Test {
  id    String @id
  props TestEnum[] @default([])
}

enum TestEnum {
  A
  B
}

```

Zenstack generate fails with:
```
Validation errors:
line: xxx: Value is not assignable to parameter [[]]
```

**Environment**

-   ZenStack version: 1.6.2
-   Prisma version: 5.8.0
-   Database type: Postgresql

**Additional context**

ymc9:
Hi @chunkerchunker , thanks for filing this issue and making the PR!

PR #948, opened 2024-01-14
bugfix: allow empty array default

chunkerchunker:
This is a proposed fix for #947.  Not sure if it's the best way to fix the issue..


ymc9:
Hi @chunkerchunker , thanks for making this change! I've made a modification to cover the case where the attribute's argument type is `any` or `any[]` in a more generic way. It should be functionally equivalent to your fix, but I feel it's a bit more extensible for the future.

The type compatibility checking part is a bit messy indeed ...

ymc9:
I'll merge it when CI passes and will include it in the upcoming 1.7.0 release.

Issue #310, opened 2023-03-30
[Feature Request] Support for auth() in @default annotation


bvkimball:
**Is your feature request related to a problem? Please describe.**
If i already set the UserId when creating the `EnhancedPrisma` it would be nice not have to pass an additonal argument around for the user or session to ensure a field is populated with the right value.

**Describe the solution you'd like**

```diff
model Post {
    id String @id
    title String @length(1, 100)
    author User @relation(fields: [authorId], references: [id])
+    authorId String @default(auth().id)

    @@allow('update', author == auth() && future().author == author)
}
```

Additional features that would be nice might be using other fields related to the user.

```diff
model Post {
    id String @id
    title String @length(1, 100)
    author User @relation(fields: [authorId], references: [id])
    authorId String @default(auth().id)

+   tenantId String  @default(auth().tenantId)

    @@allow('update', author == auth() && future().author == author)
}
```

bvkimball:
Yeah i think it would be fine and expected to have the runtime error, unless the `author` was Optional.

PR #958, opened 2024-01-23
Support for auth() in @default attribute

Azzerty23:
Hi, I'm trying to implement this feature: https://github.com/zenstackhq/zenstack/issues/310

It was an excellent opportunity to get hands-on experience with the codebase.

I believe I managed to get something working. But when I run the last 2 tests of `auth.test.ts` in isolation, they pass. However, if I run all the tests in the file, they do not. Could you help me figure out what I missed, @ymc9, @jiashengguo, please?

Also, I did not succeed in having the `@default(auth())` comments in the generated Prisma schema (I simply omitted them).

Here is the approach I followed:
- Retrieving and inserting values from the user context in the new `withDefaultAuth()` proxy. I did not fetch user data from the Prisma client. Is it worth a separated proxy?
- I worked around the existing @default attribute. While creating a new one might have been simpler, I believe this aligns more naturally in a API design perspective.
- Omitted the generation of `@default(auth())` in the Prisma schema, as it is not Prisma compatible.
- I am considering adding validation for the fields in the provided context (like in `withPolicy` proxy).

I'd appreciate going further after a quick review to ensure I'm heading in the right direction :)


ymc9:
Hi @Azzerty23 , you are amazing! Thanks for working on this feature. I've been dreaming of having it for quite a while and was thinking about it yesterday 😄. It'll definitely make the code of all ZenStack users cleaner.

I've skimmed through the change, and the structure looks brilliant. You must have spent quite some time to figure things out , as many pieces of code are not yet thoroughly documented.  Really appreciate it!

I'm not sure what happened with the test cases. Let me debug and figure out why. I'll also do a thorough review pass and attach my comments.

packages/runtime/src/enhancements/default-auth.ts
```diff
@@ -0,0 +1,83 @@
...
+                            field,
+                            fieldInfo.attributes?.find((attr) => attr.name === '@default')?.args[0]?.value as
+                                | string
+                                | undefined,
+                        ])
+                );
+                const defaultAuthFields = Object.fromEntries(
+                    Object.entries(defaultAuthSelectorFields).map(([field, selector]) => [
+                        field,
+                        selector ? userContext[selector] : userContext,
```
ymc9, line 71: Does it work if there's multi-level selection? Like `@default(auth().profile.image)`.

Azzerty23:
Thank you so much @ymc9! I must say I'm learning a lot from you :) Let me reply to your comments.

packages/ide/jetbrains/package.json
```diff
@@ -6,7 +6,7 @@
     "homepage": "https://zenstack.dev",
     "private": true,
     "scripts": {
-       "build": "./gradlew buildPlugin"
+      "build": "./gradlew buildPlugin"
```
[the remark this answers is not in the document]
Azzerty23: I updated the build script to skip it. And then by deleting my changes, it has corrected the indentation error 😄 (I just checked the indentation is correct now).

packages/runtime/src/enhancements/default-auth.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    private async preprocessWritePayload(model: string, action: PrismaWriteActionType, args: any) {
+        const visitor = new NestedWriteVisitor(this.options.modelMeta, {
+            field: async (field, action, data, context) => {
+                const userContext = this.context?.user;
+                if (!userContext) {
+                    throw new Error(`Invalid user context`);
+                }
+                const fields = this.options.modelMeta.fields[model];
```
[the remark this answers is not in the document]
Azzerty23: I'm new to visitor pattern, I didn't do it by purpose. That could explain why there are side effects between tests? I will try to refactor this part.

packages/schema/src/plugins/prisma/schema-generator.ts
```diff
@@ -285,7 +286,12 @@ export default class PrismaSchemaGenerator {
             !!attrDecl.attributes.find((a) => a.decl.ref?.name === '@@@prisma') ||
             // the special pass-through attribute
             attrDecl.name === MODEL_PASSTHROUGH_ATTR ||
-            attrDecl.name === FIELD_PASSTHROUGH_ATTR
+            attrDecl.name === FIELD_PASSTHROUGH_ATTR ||
+            // auth() in @default() is not supported by Prisma
+            // FIXME: condition is inverted to avoid error...
+            !!attrDecl.attributes.find(
+                (a) => a.decl.ref?.name === '@default' && a.args[0].value.$cstNode?.text.startsWith('auth()')
```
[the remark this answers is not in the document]
Azzerty23: Noted, I will use it. However, this part doesn't work as I expected. I anticipated having the attributes in the comment with this change. I'm unsure about this part.

packages/schema/src/plugins/prisma/schema-generator.ts
```diff
@@ -334,6 +340,10 @@ export default class PrismaSchemaGenerator {
             } else {
                 throw new PluginError(name, `Invalid arguments for ${FIELD_PASSTHROUGH_ATTR} attribute`);
             }
+            // do not write @default(auth()) field attribute as it is not supported by Prisma
+            // TODO: we should add a comment to the field
+        } else if (attrName === '@default' && attr.args[0].value.$cstNode?.text.startsWith('auth()')) {
```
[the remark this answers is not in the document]
Azzerty23: Absolutely!

packages/runtime/src/enhancements/default-auth.ts
Azzerty23: Good point, I don't think so

packages/runtime/src/enhancements/default-auth.ts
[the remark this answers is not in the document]
Azzerty23: I now remember why I did that... Since a default value may not be passed in the payload, the `field` callback is not always executed on it.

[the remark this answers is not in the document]
ymc9: I see, I completely missed that part ...

Thinking about it again, it's actually more complex, and we can't simply rely on the "field" callback. For example:

```prisma
model Foo {
  id Int @default(autoincrement())
  ownerId @default(auth().id)
}
```

```ts
await db.foo.create({data: {}});
```

In this case, `field` callback is not called at all. The `NestedWriterVisitorCallback` type has a bunch of recursive hooks for payload like `create`, `createMany`, `update`, `updateMany`, etc. They allow you to fully traverse the entire update payload, regardless if there are any field assignments.

The policy proxy uses this approach extensively to inject into nested payloads. One example is here: https://github.com/zenstackhq/zenstack/blob/3b9a6c4a09be2ca24fff448d4265414c6e4909c1/packages/runtime/src/enhancements/policy/handler.ts#L302

Using these callback hooks will probably be more complex than your current implementation. Let me know if you need help to translate your current code to that. Always happy to help!

packages/schema/src/plugins/prisma/schema-generator.ts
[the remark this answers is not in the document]
ymc9: Oh, I see. I think we can even skip it for now if it's complicated. The comment generation is for convenience, and I think few people really look into the generated prisma schema anyway 😄.

packages/runtime/src/enhancements/create-enhancement.ts
```diff
@@ -120,13 +123,26 @@ export function createEnhancement<DbClient extends object>(
 
     let result = prisma;
 
-    if (hasPassword === undefined || hasOmit === undefined) {
+    if (
+        process.env.ZENSTACK_TEST === '1' || // avoid caching in tests
```
ymc9, line 127: It turns out the test failure is due to the cache of these flags. I've added the condition to exclude the caching in test env.

packages/runtime/src/enhancements/default-auth.ts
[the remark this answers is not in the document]
Azzerty23: I understand. I have revised this section to use only 'create' for now. I will incorporate additional callbacks once the basic unit test passes.

packages/runtime/src/enhancements/default-auth.ts
Azzerty23: I need your help for this one. I made a `deepGet` helper for the recursive matter (I'm used to use the `get` lodash function, do you already have this kink of helper somewhere in the code ?) but the issue is from the zmodel syntax now. It cannot parse the `auth().a.b` syntax. I tried to update the `zmodel.langium` but without success. I test this case in `auth.test.ts` (last one).

packages/runtime/src/enhancements/create-enhancement.ts
Azzerty23: Thank you!

packages/runtime/src/enhancements/default-auth.ts
ymc9: Sure. Let me think about how to deal with this.

ymc9: Hi @Azzerty23 , could you help grant me with write permission to your fork of ZenStack? I've made some refactor and would like push it to a separate branch to show the difference, instead of directly committing to this one. Thanks!

Azzerty23: Sure, I just did it.

ymc9: Thanks! I've just submitted a PR over there, with descriptions on things changed. Please help check if they make sense or there's anything unclear.

Azzerty23: Wow, that's really nice! The `defaultValueProvider` is so much cleaner. Do you need me to add anything (tests)?
I noticed how you changed the data property directly in the visitor (I had to see it to believe it, ahah). I suppose I can remove my workaround with `newArgs` since it's unnecessary.

ymc9: I think we can keep the `newArgs` there because it's good that the proxy doesn't modify the args that user passes in (instead works on a clone of it).

If it looks good to you, I'll merge the PR to your branch, and let's see if CI passes here 😄. If so, I think the feature is in pretty good shape now, and we can merge it to V2. I'll probably add a set of test cases using Postgres to cover cases not supported by SQLite (e.g., `createMany`). But that'll require some additional setup and a local pg instance. I can do it later.

Thanks for working on this great feature! It'll benefit many people.

Azzerty23: Okay, I'll let you ignore my last commit (removing `newArgs`) and proceed to merge the PR. Thank you for your prompt and wonderful support!

Issue #242, opened 2023-03-06
[Feature Request] Permissions checker


keanugrieves:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to check whether user has the permission to perform a create, update or delete operation so I could reflect it on the frontend.

**Describe the solution you'd like**
A "simple" (ha ha) function on each record that optionally takes an input and returns a boolean that indicates whether the user has permission.

**Describe alternatives you've considered**
Rewriting my zmodel permissions in an additional place as a separately callable function.

**Additional context**
Brainstorming the structure, given the model `Foo`:

```
// If user has any create permissions, return true when input not provided or whether they can create with input
foo.canCreate(input?: Prisma.FooUncheckedCreateInput) => boolean;

// If user has update permissions on current record, return true when input not provided or return whether they can update with current record and input
foo.canUpdate(input?: Prisma.FooUncheckedUpdateInput) => boolean;

// If user has delete permissions on current record, return true
foo.canDelete() => boolean;
```

gczh:
@ymc9 adding some questions here:

1. Would the approach to implement this require more database calls or can these all be done at runtime? Ideally, it shouldn't require more database calls for performance reasons.

ymc9:
> @ymc9 adding some questions here:
> 
> 
> 
> 1. Would the approach to implement this require more database calls or can these all be done at runtime? Ideally, it shouldn't require more database calls for performance reasons.

Thanks @gczh . I believe it doesn't involve database calls, however it's possible it'll be served as a backend api so it'll involve an api from the frontend.

We haven't settled the design yet, and I'll share more information when the work starts.

Azzerty23:
Hi @ymc9, I'm interested in working on it. How do you envision the design? Initially, I was wondering if it might be better to create a single generic hook to fetch multiple permissions in a single API call.

To ensure that I could achieve this, I started a draft that I will share here. Hopefully, this will help in the process.

ymc9:
> Hi @ymc9, I'm interested in working on it. How do you envision the design? Initially, I was wondering if it might be better to create a single generic hook to fetch multiple permissions in a single API call.
> 
> To ensure that I could achieve this, I started a draft that I will share here. Hopefully, this will help in the process.

Thank you for starting this work @Azzerty23 ! Another dream feature and I really appreciate it!

Let me go through the draft PR and share my thoughts there.

nahtnam:
[...]

PR #966, opened 2024-01-29
feat: permissions checker

Azzerty23:
Draft for Permissions checker / issue #242


Azzerty23:
In `policy`, I am thinking of adding a `permissions` property that should look like this:

```ts
function permissionsChecker(args: any, user?: AuthUser) {
  return {
    post: {
      read: {
        OR: [
          {
            field: args.authorId,
            operator: "===", // or comparator ?
            value: user?.id,
          },
          {
            field: args.author?.id,
            operator: "===",
            value: user?.id,
          },
          {
            field: args.author?.name,
            operator: "===",
            value: user?.name,
          },
        ],
      },
      delete: {
        AND: [
          // {
          //     'author': { // or nested like this ?
          //         'name': {
          //             'operator': '===',
          //             'value': user.name
          //         }
          //     }
          // },
          // 'AND': [
          //         {
          //             'score': {
          //                 'operator': '<',
          //                 'value': 10
          //             }
          //         },
          //         {
          //             'role': {
          //                 'operator': 'in',
          //                 'value': ['ADMIN', 'REVIEWER']
          //             }
          //         }
          // ]
        ],
      },
      // note for myself
      create: { AND: [] }, // return true
      update: { OR: [] }, // return false
    },
    user: {
      create: true,
      update: false,
      read: true,
      delete: false,
    },
  };
}

const policy: PolicyDef = {
  guard: {...},
  validation: {...},
  permissions: permissionsChecker,
};
export default policy;
```

Azzerty23:
Imagined implementation details / steps:
 - [ ] Generation of permissions from the zmodel AST, at the same level as other generated policy properties (guard, validation, etc.)
 - [ ]  Addition of the `check` operation on the Prisma proxy, which takes the operation type (of type PolicyOperationKind) and the passed arguments (where/data...) as parameters and returns a boolean promise.
- [ ]  Modification of the API handler to handle CRUD operations
Endpoint: api/model/[…path]/routes ⇒ e.g., api/model/user/check/read
Or a new API handler on a specific endpoint?
- [ ]  Generation of hooks to fetch the check API.

Azzerty23:
Thinking again about `permissionsChecker()`...

I should instead return an object with inline conditions, like this :

```ts
function permissionsChecker(args: any, user?: AuthUser) {
  return {
    post: {
      read: args.authorId === user?.id || args.author?.id === user?.id || args.author?.name === user?.name,
      delete: true,
     // ...
    },
    user: {
      // ...
    },
  };
}
```

I don't know why I love making things more complicated than they need to be ^^

b-barry:
Looking forward for this feature. At the moment do we have a workaround for this feature?

Azzerty23:
For now, the only solution I have is to manually check the property on the authenticated user/resources. This is problematic as it is not synchronized with our zmodel schemas.

I have planned to dedicate tomorrow to establishing the basics of this, I will share my progress in any case.

Azzerty23:
What I have in mind to build the permissions checker itself:

1. Obtain policy rules from the zmodel AST.
2. Resolve each one (with transformation rules below).
3. Build complete conditions:

```ts
// `@@deny` / `@@allow` → deny first
!(denyCondition || denyCondition) && (allowCondition || allowCondition)
```
___

Transformation rules:

1. transform auth() expressions with TypeScriptExpressionTransformer
`@@allow("read", auth().id == 1)`
```ts
user?.id === 1
```

2. append other expressions with **args**
`@@allow("read", status == "public")`
```ts
args?.status === "public"
```

3. transform `null` to `undefined`
_e.g. require login_
`@@deny('all', auth() == null)`
```ts
user?.id === undefined
```

4. primary / foreign key / model → create conditions for primary / foreign key and all **unique** fields
`@@allow("read", auth().id == author.id)`
`@@allow("read", auth() == author)`
```ts
user?.id === (args?.author?.id || args?.authorId) || user?.email === args?.author?.email
```

5. prevent undefined value → wrap each value with `define` guard
```ts
define(args?.authorId) === user?.id || define(args?.author?.id) === user?.id
```

6. collection predicate expression
`@@allow('read', members?[user == auth()]`
```ts
// enhancedDb.space.check("read", { where: {id: 1} })
args?.id === define(user?.spaceId)

// enhancedDb.space.check("read", { where: {name: "best-space"} })
args?.name === define(user?.space?.name)
```

7. field policy → ignored I think
`content  String  @deny('read', !published)`

ymc9:
Hi @Azzerty23 , thanks for continuing to work on it. My apologies for the late response.

I've been thinking about this feature for a while, and I think there are several questions to be answered before getting into details.

### 1. What's the semantic of "check"?

Let's say if we call `db.model.check('read')`. When should it return `true` (without making a database call)? My proposed definition is:

> Whether there can exist any database record satisfying the "read" policy. 

This effectively means whether the "read" policy has any self-contradiction, like:

```
@@allow('read', x > 0 && x <= 0)
```

Note that this can also involve relations like:

```
@@allow('read', a.x > 0 && a.x <= 0)
```

If we accept this semantic, other CRUD operations should be consistent:
- `create`: whether there can exist an input args satisfying the "create" policy.
- `update`: whether there can exist a database record satisfying the "update" policy.
- `delete`: whether there can exist a database record satisfying the "delete" policy.

### 2. Should we allow more fine-grained checks?

Just checking for CRUD is very coarse, and I believe there will be many scenarios where one wants to check for a smaller scope by providing "partial conditions". Like:

```
db.model.check('read', { x: { gt: 0 } });
db.model.check('create', { x: 1 });
```

To support that, we can simply merge the user-provided conditions with policy rules and check if it's satisfiable.

### 3. How to implement the check?

With the interpretation above, permission checking seems to become "constraint solving". The policy rules and user-provided conditions together form a set of constraints to satisfy. If we can find a model value (a solution) that satisfies the conditions, we can return `true`.

Implementing such a solver can be quite tricky. We should probably use an existing constraint solver library (like [Microsoft Z3](https://www.npmjs.com/package/z3-solver)) to get the job done. In a nut shell it let's you do things like:

```js
const x = Int.const('x');
const solver = new Solver();
solver.add(And(x.ge(0), x.le(9)));
console.log(await solver.check());
```

If so we can just encode the policy expressions into constraint solver's input and let it do the magic.

---

Over all things, I'm most unsure about #2. To be honest I don't have very good ideas about how the API will be used in real-world applications. Maybe we should create a channel/thread to gather from people who need it?

Azzerty23:
Thank you @ymc9 I had not imagined such a use case where we would add constraints/conditions to the permission checker. In my understanding, all 'constraints' were supposed to be present in the zmodel schema, and we would check their equality against actual values/available values at runtime. But you are absolutely right; let me take a closer look at this constraint solver library.

Azzerty23:
To address each of your points:

1. This was the understanding I had following your [comment](https://github.com/zenstackhq/zenstack/issues/242#issuecomment-1841904092).
2. Now that you've mentioned it, I don't see how to skip the "partial conditions." As soon as we want to compare sets of values, it becomes essential.
3. I am exploring this avenue.

Let me create the thread in Discord.

Azzerty23:
I compared [kiwi](https://github.com/lume/kiwi) and [Z3](https://github.com/Z3Prover/z3/tree/master/src/api/js) and decided to proceed with Z3 because the syntax is simpler, and there are many missing features in Kiwi (such as the OR condition). Here are my experiments with Z3: https://github.com/Azzerty23/z3-test/blob/main/src/solver.ts. I faced challenges before finding the solution to handle the process that is not automatically terminated. However, I now have a better idea how to implement the feature. Hopefully, I will find time in a week or two to continue working on it and complete the task.

packages/schema/src/plugins/prisma/schema-generator.ts
```diff
@@ -584,7 +584,7 @@ export class PrismaSchemaGenerator {
                 'FieldReference',
                 new PrismaFieldReference(
                     resolved(node.target).name,
-                    node.args.map((arg) => new PrismaFieldReferenceArg(arg.name, this.exprToText(arg.value)))
+                    node.args.map((arg) => new PrismaFieldReferenceArg(arg.name, this.exprToText(arg.$container)))
```
Azzerty23, line 587: @ymc9 I tried to fix the typing by updating it. Could you check it please ? I will revert my change to make the tests pass.

packages/sdk/src/zmodel-code-generator.ts
```diff
@@ -267,7 +267,7 @@ ${ast.fields.map((x) => this.indent + this.generate(x)).join('\n')}${
 
     @gen(ReferenceArg)
     private _generateReferenceArg(ast: ReferenceArg) {
-        return `${ast.name}:${this.generate(ast.value)}`;
```
Azzerty23, line 270: @ymc9 and here too

packages/schema/src/plugins/prisma/schema-generator.ts
Azzerty23: Sorry, it's ok. My editor complained about this and not anymore... very weird.

ymc9:
Hi @Azzerty23 , I went through Z3 documentation and thought about how to model constraints with it too. I've captured my thoughts in this repo: https://github.com/ymc9/z3-exp

I think it overlaps with your draft implementation in many ways, but with some differences too. Hope it helps. There might be things we need to discuss in more details.

Azzerty23:
> I went through Z3 documentation and thought about how to model constraints with it too. I've captured my thoughts in this repo: https://github.com/ymc9/z3-exp

No, Z3 handles string and array constraints natively :o I used the [playground](https://microsoft.github.io/z3guide/programming/Z3%20JavaScript%20Examples) as a reference and dove headfirst into the high-level API, knowing it had limitations, but without delving into them in detail... Thank you for taking the time to explain everything and for looking at my implementation. I'll study it more carefully tonight.

PR #1011, opened 2024-02-19
fix: zenstack cli errors while using bun/bunx during docker build

ErikMCM:
It's an obscure issue but it exists in my CI pipeline. 

This PR attempts to fix the error that occurs when the following criteria are met:
1. Zenstack CLI commands are being ran on the Bun runtime
2. The terminal is non-TTY or not interactive (such as during docker build)
3. The REPL action is instantiated, requiring `pretty-repl` which subsequently requires `repl` which is not supported by the Bun runtime at this time

This PR fixes the error by:
1. Importing `pretty-repl` after checking if the runtime is bun and in a non-TTY terminal
2. Using bunx instead of npx within the `zenstack generate` command

---

This PR could be reverted once Bun REPL support is fully implemented, allowing the `zenstack repl` command to be used in such a case. However, considering the current support for REPL on Bun is still experimental, it may be some time before it is resolved.

### DIFF
--- a/packages/schema/src/plugins/prisma/schema-generator.ts
+++ b/packages/schema/src/plugins/prisma/schema-generator.ts
@@ -124,12 +124,12 @@ export default class PrismaSchemaGenerator {
         }
         await writeFile(outFile, this.PRELUDE + prisma.toString());
 
-        const packageExec = process?.versions?.bun ? 'bunx' : 'npx';
+        const packageManager = process?.versions?.bun ? 'bunx' : 'npx';
 
         if (options.format === true) {
             try {
                 // run 'prisma format'
-                await execSync(`${packageExec} prisma format --schema ${outFile}`);
+                await execSync(`${packageManager} prisma format --schema ${outFile}`);
             } catch {
                 warnings.push(`Failed to format Prisma schema file`);
             }
@@ -138,7 +138,7 @@ export default class PrismaSchemaGenerator {
         const generateClient = options.generateClient !== false;
 
         if (generateClient) {
-            let generateCmd = `${packageExec} prisma generate --schema "${outFile}"`;
+            let generateCmd = `${packageManager} prisma generate --schema "${outFile}"`;
             if (typeof options.generateArgs === 'string') {
                 generateCmd += ` ${options.generateArgs}`;
             }
--- a/packages/schema/src/plugins/prisma/schema-generator.ts
+++ b/packages/schema/src/plugins/prisma/schema-generator.ts
@@ -124,10 +124,12 @@ export default class PrismaSchemaGenerator {
         }
         await writeFile(outFile, this.PRELUDE + prisma.toString());
 
+        const packageExec = process?.versions?.bun ? 'bunx' : 'npx';
+
         if (options.format === true) {
             try {
                 // run 'prisma format'
-                await execSync(`npx prisma format --schema ${outFile}`);
+                await execSync(`${packageExec} prisma format --schema ${outFile}`);
             } catch {
                 warnings.push(`Failed to format Prisma schema file`);
             }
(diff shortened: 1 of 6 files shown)


packages/schema/src/plugins/prisma/schema-generator.ts
```diff
@@ -124,10 +124,12 @@ export default class PrismaSchemaGenerator {
         }
         await writeFile(outFile, this.PRELUDE + prisma.toString());
 
+        const packageManager = process?.versions?.bun ? 'bunx' : 'npx';
+
         if (options.format === true) {
             try {
                 // run 'prisma format'
-                await execSync(`npx prisma format --schema ${outFile}`);
+                await execSync(`${packageManager} prisma format --schema ${outFile}`);
```
ymc9, line 132: Shall we extract an `execPackage` utility to encapsulate the pm detection here? I feel it's going to be needed in other places too in the future.

ymc9:
Thank you for working on this @ErikMCM !

packages/schema/src/plugins/prisma/schema-generator.ts
ErikMCM: Yeah, I thought about that but refrained from doing so because it was used only in that case in the schema package. However, I do agree and can implement this as it would likely ensure better support in the future. Do you have any recommendations about how you see this best done?

I notice there is already a [`getPackageManager` util function](https://github.com/zenstackhq/zenstack/blob/main/packages%2Fschema%2Fsrc%2Futils%2Fpkg-utils.ts) that I could export and adapt, but instead I think I'll create another function for package executors in specific, seeing that they do two very different things. I believe I'll put it in the same file, unless you think exec utils or another file would be a better place.

I'll leave tests using NPX for obvious reasons. I'll also leave other packages using NPX for now, considering I'm not familiar with the environment that they may be executed in.

ymc9: > Yeah, I thought about that but refrained from doing so because it was used only in that case in the schema package. However, I do agree and can implement this as it would likely ensure better support in the future. Do you have any recommendations about how you see this best done?
> 
> I notice there is already a [`getPackageManager` util function](https://github.com/zenstackhq/zenstack/blob/main/packages%2Fschema%2Fsrc%2Futils%2Fpkg-utils.ts) that I could export and adapt, but instead I think I'll create another function for package executors in specific, seeing that they do two very different things. I believe I'll put it in the same file, unless you think exec utils or another file would be a better place.
> 
> I'll leave tests using NPX for obvious reasons. I'll also leave other packages using NPX for now, considering I'm not familiar with the environment that they may be executed in.

Got it. Yes, npx should be the default as long as we're in a Node.js environment. I was just thinking of creating a new `exec...` utility (similar to the current `execSync` one) which encapsulates the bun detection (and npx as default) in it.

For the general task of package manager detection (npm, yarn, pnpm, etc.), I always wanted to migrate to using [@antfu/ni](https://www.npmjs.com/package/@antfu/ni), which Prisma also uses. But that should be a separate change.

ErikMCM: I've implemented the `execPackage` function in the latest commit which should do what you've outlined here. 

Also, it looks like [@antfu/ni](https://www.npmjs.com/package/@antfu/ni) supports getting a package executor (see the [nlx section](https://www.npmjs.com/package/@antfu/ni#nlx---download--execute)). I could work on another PR implementing this package for both the `getPackageManager` and `execPackage` functions after this one is merged.

ymc9: Got it. Thanks! I've approved this PR.

Just a reminder: I noticed that nlx translates to `pnpm dlx` and `yarn dlx`. IIRC, they always download the package to temp folder regardless if it's already installed in the current project. I think always resorting to "npx" should be safe for a Node environment?

ymc9 (approved):

packages/schema/src/plugins/prisma/schema-generator.ts
ymc9: Cool! I'm merging this PR now. Is it blocking you now? The regular ship cadence is bi-weekly, so the 1.10.0 release will be at the end of Feb. If it's blocking, I can also publish a patch release for it.

ErikMCM: I have a workaround that I've been using for the past few months. There isn't any rush, it's just mainly to speed up build times and not copy files in and out of a docker container for one action.

PR #1015, opened 2024-02-20
fix: default auth without user context

Azzerty23:
The withDefaultAuth enhancer should allow actions that do not use this feature. However, it throws an error when there is a @default(auth()) in the schema for an anonymous user, regardless of whether the action involves it.

## Summary by CodeRabbit

- **Refactor**
	- Improved default authentication handling for enhanced security and reliability.
- **Tests**
	- Added new test case to verify authentication behavior in specific scenarios.
- **Bug Fixes**
	- Updated parameter type in `useModelQuery` and `useInfiniteModelQuery` functions to exclude `queryKey` property.

### DIFF
--- a/tests/integration/tests/enhancements/with-policy/auth.test.ts
+++ b/tests/integration/tests/enhancements/with-policy/auth.test.ts
@@ -522,7 +522,7 @@ describe('With Policy: auth() test', () => {
             author User @relation(fields: [authorId], references: [id])
             authorId String @default(auth().id)
 
-            @@allow('all', auth().id != null)
+            @@allow('all', true)
         }
         `
         );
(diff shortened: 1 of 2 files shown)


tests/integration/tests/enhancements/with-policy/auth.test.ts
```diff
@@ -505,4 +505,33 @@ describe('With Policy: auth() test', () => {
...
+            @@allow('all', true)
+        }
+
+        model Post {
+            id String @id @default(uuid())
+            title String
+            author User @relation(fields: [authorId], references: [id])
+            authorId String @default(auth().id)
+
+            @@allow('all', auth().id != null)
```
Azzerty23, line 525: This is confusing, I should have used `@@allow('all', true)`, but it doesn't change the test.

Azzerty23:
As tests failed anyway, I could also improve the error message. What about: "Using \`auth()\` in \`@default\` for the field \`${fieldInfo.name}\` requires a user context" or just "\`${fieldInfo.name}\` requires a user context"?

ymc9 (approved):

Issue #1008, opened 2024-02-19
[Feature Request] extended `import` utilising package manager to provide exportability


TGTGamer:
**Is your feature request related to a problem? Please describe.**
Extension of the `import` utility  to enable user to import from `node_modules` folders, enabling the ability to extend from external packages. 

**Describe the solution you'd like**
A simple way to extend from `node_modules` folders. From the looks of it, an refactor of 

https://github.com/zenstackhq/zenstack/blob/b4579c7f3a5d47e1f61c5268d3ac8efbebc7557d/packages/schema/src/utils/ast-utils.ts#L96-L106

I assume there would be an additional step somewhere within the language file, and probably within the vscode extension, however haven't yet discovered those locations (will update once I have)

**Describe alternatives you've considered**
Keeping all my files in a single project and making endless copies of them into each individual project. Meaning there's no single source of truth, making it easy for one scheme to fall out of sync with another....

PR #1019, opened 2024-02-20
Refactor: find up

TGTGamer:
# Goal
Starts #1008 

Refactor to make better use of this function and ensure multiple directories can be found such as needed for `node_module` directories to ensure support for mono-repos

Details as follows.

# findUp (): String | String[] | undefined

## Summary
The `findUp` function is used to search for file paths by recursively searching parent directories based on a given list of names and the current working directory. It can return a single path or multiple paths, depending on the specified flag. If no paths are found, it returns undefined.

## Example Usage
```javascript
const names = ['node_modules'];
const cwd = '/path/to/parent/child/project/current/directory';
const result = findUp(names, cwd, true);
console.log(result);
// Output: ['/path/to/parent/child/project/current/directory/node_modules', '/path/to/parent/node_modules']
```

### Inputs
- `names` (array of strings): An array of names to search for within the directory.
- `cwd` (string, optional): The current working directory. Defaults to the process's current working directory.
- `multiple` (boolean, optional): A flag indicating whether to search for multiple levels of files. Defaults to false.
- `result` (array of strings, optional): An array to store the accumulated results when searching for multiple paths.
___
### Flow
1. Check if any of the names in the list are empty. If all names are empty, return undefined.
2. Find the first name in the list that exists in the current working directory. If found and multiple is false, return the path as a string.
3. If multiple is true and a name is found, add the path to the result array.
4. Get the parent directory path of the current working directory.
5. If the parent directory is the same as the current working directory, return the result array if multiple is true and it contains paths, otherwise return undefined.
6. Recursively call the `findUp` function with the parent directory as the new current working directory.
7. Return the result of the recursive call.
___
### Outputs
- If multiple is false and a path is found, it returns the path as a string or undefined if no path is found.
- If multiple is true and at least one path is found, it returns an array of paths or undefined if no path is found.
___

### Tests                                                                                                                                                     
findup.spec.js
Total: 7 milliseconds
✓ 9

✓ should return the name of the first matching file when multiple is false and a match is found - 3 milliseconds
✓ should return undefined when no matches are found - 1 millisecond
✓ should search for specified files in parent directories starting from the given current working directory (cwd) - 0 milliseconds
✓ should accumulate results when multiple is true and matches are found - 2 milliseconds
✓ should return undefined when names array is empty - 0 milliseconds
✓ should return undefined when cwd is not a valid directory - 0 milliseconds
✓ should return undefined when no matches are found and multiple is true - 1 millisecond
✓ should return undefined when no matches are found and multiple is false - 0 milliseconds
✓ should return undefined when names array contains an empty string - 0 milliseconds

Tests can be found here: https://bit.cloud/tgtgamer/experiments/findup/~tests
___

## Summary by CodeRabbit

- **Refactor**
	- Enhanced file search functionality within the schema generation process for more efficient file path resolution.
- **Chores**
	- Updated internal utilities to support more flexible file searching capabilities and introduced a deprecation notice for outdated methods.

### DIFF
--- a/packages/schema/src/utils/pkg-utils.ts
+++ b/packages/schema/src/utils/pkg-utils.ts
@@ -107,6 +107,11 @@ export function ensurePackage(
     }
 }
 
+/**
+ * A function that searches for the nearest package.json file starting from the provided search path or the current working directory if no search path is provided. 
+ * It iterates through the directory structure going one level up at a time until it finds a package.json file. If no package.json file is found, it returns undefined.
+ * @deprecated Use findUp instead @see findUp
+ */
 export function findPackageJson(searchPath?: string) {
     let currDir = searchPath ?? process.cwd();
     while (currDir) {
--- a/packages/schema/src/utils/pkg-utils.ts
+++ b/packages/schema/src/utils/pkg-utils.ts
@@ -1,20 +1,42 @@
-import fs from 'fs';
-import path from 'path';
+import fs from 'node:fs';
+import path from 'node:path';
 import { execSync } from './exec-utils';
 
 export type PackageManagers = 'npm' | 'yarn' | 'pnpm';
 
-function findUp(names: string[], cwd: string): string | undefined {
-    let dir = cwd;
-    // eslint-disable-next-line no-constant-condition
-    while (true) {
-        const target = names.find((name) => fs.existsSync(path.join(dir, name)));
-        if (target) return target;
-
-        const up = path.resolve(dir, '..');
-        if (up === dir) return undefined; // it'll fail anyway
-        dir = up;
-    }
+/**
+ * A type named FindUp that takes a type parameter e which extends boolean. 
+ * If e extends true, it returns a union type of string[] or undefined. 
+ * If e does not extend true, it returns a union type of string or undefined.
+ * @author Jonathan Stevens (TGTGamer)
+ *
+ * @export
+ * @template e A type parameter that extends boolean
+ */
+export type FindUp<e extends boolean> = e extends true ? string[] | undefined : string | undefined
+/**
+ * Find and return file paths by searching parent directories based on the given names list and current working directory (cwd) path. 
+ * Optionally return a single path or multiple paths. 
+ * If multiple allowed, return all paths found. 
+ * If no paths are found, return undefined.
+ * @author Jonathan Stevens (TGTGamer)
+ *
+ * @export
+ * @template [e=false]
+ * @param names An array of strings representing names to search for within the directory
+ * @param cwd A string representing the current working directory
+ * @param [multiple=false as e] A boolean flag indicating whether to search for multiple levels. Useful for finding node_modules directories...
+ * @param [result=[]] An array of strings representing the accumulated results used in multiple results
+ * @returns Path(s) to a specific file or folder within the directory or parent directories
+ */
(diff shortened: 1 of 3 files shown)


packages/schema/src/utils/pkg-utils.ts
```diff
@@ -1,20 +1,42 @@
...
-
-        const up = path.resolve(dir, '..');
-        if (up === dir) return undefined; // it'll fail anyway
-        dir = up;
-    }
+/**
+ * A type named FindUp that takes a type parameter e which extends boolean. 
+ * If e extends true, it returns a union type of string[] or undefined. 
+ * If e does not extend true, it returns a union type of string or undefined.
+ * @author Jonathan Stevens (TGTGamer)
```
ymc9, line 11: Shall we remove `@author` tag for consistency? I think people usually have a git extension in the IDE that shows the author.

packages/schema/src/utils/pkg-utils.ts
```diff
@@ -1,20 +1,42 @@
...
+ *
+ * @export
+ * @template [e=false]
+ * @param names An array of strings representing names to search for within the directory
+ * @param cwd A string representing the current working directory
+ * @param [multiple=false as e] A boolean flag indicating whether to search for multiple levels. Useful for finding node_modules directories...
+ * @param [result=[]] An array of strings representing the accumulated results used in multiple results
+ * @returns Path(s) to a specific file or folder within the directory or parent directories
+ */
+export function findUp<e extends boolean = false>(names: string[], cwd: string = process.cwd(), multiple: e = false as e, result: string[] = []): FindUp<e> {
```
ymc9, line 32: The refactor looks good to me. Just to confirm, it's a preparation for the other in-progress PR, right?

TGTGamer: Yeah. This will allow for the node_modules directory to be found using the FindUp function rather than creating a new one, and maintains backwards compatibility with existing functionality

packages/schema/src/utils/pkg-utils.ts
TGTGamer: Can do, I didn't see any other JSDoc comments so I just left it with my default configuration :)

TGTGamer:
```suggestion
```

packages/schema/src/utils/pkg-utils.ts
```diff
@@ -1,20 +1,42 @@
...
+ * @export
+ * @template e A type parameter that extends boolean
+ */
+export type FindUp<e extends boolean> = e extends true ? string[] | undefined : string | undefined
+/**
+ * Find and return file paths by searching parent directories based on the given names list and current working directory (cwd) path. 
+ * Optionally return a single path or multiple paths. 
+ * If multiple allowed, return all paths found. 
+ * If no paths are found, return undefined.
+ * @author Jonathan Stevens (TGTGamer)
```
TGTGamer, line 22:
```suggestion
```

ymc9 (approved):

TGTGamer:
Looks like all the tests for this passed now, I think merge of this is the only think left to do before I can rebase #1021 and mark that ready to review :)

ymc9:
Sure. I'm merging it now.

Issue #1179, opened 2024-03-27
auth() couldn't find the field defined in the abstract base model


jiashengguo:
```prisma
abstract model Base {
  id String @id @default(uuid())
}

model User extends Base {
  email String
  posts Post[]
  @@allow('all', auth() == this)
}

model Post {
  id String @id @default(uuid())

  user User @relation(fields: [userId], references: [id])
  userId String
  @@allow('all', auth().id == userId)
}
```

It will shows error:
Could not resolve reference to DataModelField named 'id'.zmodel(linking-error)

PR #1192, opened 2024-03-30
fix: `auth()` field from a base model is not properly resolved

ymc9:
## Summary by CodeRabbit

- **Refactor**
	- Simplified the linking process in the language server for better performance and reliability.
- **Tests**
	- Added a new test case for regression issue 1179 to ensure data model integrity.


ymc9:
Fixes #1179

PR #1208, opened 2024-04-02
[feature request]: implement shorthand for `@readonly` as `z.literal()`

sakgoyal:


ymc9:
Hi @sakgoyal , thanks for the PR again 😄.

I want to be conservative with introducing new attributes because each of them adds some cognitive cost to the users. May I keep this PR on hold until we get a few more similar ask for the feature? If your main need here is zod schema here, maybe you can also use zod API to extend the generated schemas?

sakgoyal:
that is why I kept it as a draft. I wanted feedback before continuing. 
I will gladly wait since it is not urgent for me.

ymc9:
> that is why I kept it as a draft. I wanted feedback before continuing. I will gladly wait since it is not urgent for me.

Got it. Thanks for understanding!

ymc9:
Closing for now since there's no plan to merge in the near future

Issue #1241, opened 2024-04-11
Two one-to-many relations on the same model generated wrong model meta data


jiashengguo:
Running `zenstack generate` for below model files

```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider   = "postgresql"
  url = env('DATABASE_URL')
}

plugin prisma {
  provider = '@core/prisma'
  output = './prisma/schema.prisma'
}

model User {
  id String @id @default(uuid()) @db.Uuid
  todos Todo[]

  @@auth
  @@allow('all', true)
}

model Todo {
  id String @id @default(uuid()) @db.Uuid

  user_id String @db.Uuid
  user User @relation(fields: [user_id], references: [id])

  images File[] @relation("todo_images")
  documents File[] @relation("todo_documents")

  @@allow('all', true)
}

model File {
  id String @id @default(uuid()) @db.Uuid
  s3_key String @unique
  label String

  todo_image_id String? @db.Uuid
  todo_image Todo? @relation("todo_images", fields: [todo_image_id], references: [id])

  todo_document_id String? @db.Uuid
  todo_document Todo? @relation("todo_documents", fields: [todo_document_id], references: [id])

  @@allow('all', true)
}
```

The `backLink` generated model-meta.js is incorrect:

```ts
const metadata = {
fields:{
     file:{
              ...
             todo_document: {
                name: "todo_document",
                type: "Todo",
                isDataModel: true,
                isOptional: true,
                attributes: [{ "name": "@relation", "args": [{ "value": "todo_documents" }] }],
                backLink: 'images',  // here should be **documents**
                isRelationOwner: true,
                foreignKeyMapping: { "id": "todo_document_id" },
            },
            ...
         }
}

PR #1244, opened 2024-04-12
fix: model meta generator doesn't correctly identify relation names

ymc9:
## Summary by CodeRabbit

- **Refactor**
	- Improved attribute reference checks in data model relationships.

- **Tests**
	- Added a regression test case to ensure reliable updates in data model operations.


ymc9:
Fixes #1241

Issue #1265, opened 2024-04-17
[Zod] Field with `auth()` in `@default()` should be generated as optional in zod schemas


ymc9:
```prisma
model User {
    id String @id @default(uuid())
    posts Post[]              
    @@allow('all', true)
}

model Post {
    id String @id @default(uuid())
    title String @default('xyz')
    userId String @default(auth().id)
    user User @relation(fields: [userId], references: [id])
    @@allow('all', true)
}
```

ymc9:
Fixed in v2.0.0

PR #1266, opened 2024-04-17
fix(zod): field with `auth()` and `@default()` should be generated as optional

ymc9:
## Summary by CodeRabbit

- **New Features**
	- Enhanced CLI descriptions to clarify the usage of core plugins.
	- Introduced new test cases for plugin generation and schema regression scenarios.

- **Refactor**
	- Simplified and clarified internal naming and import paths for better alignment with project standards.
	- Adjusted schema generation logic to handle default values with authentication more effectively.

- **Documentation**
	- Updated terminology from "built-in plugins" to "core plugins" across documentation and tool tips.


ymc9:
Fixes #1265

Issue #1400, opened 2024-05-01
[BUG] zmodel does not support calling @default(dbgenerated()) with no parameters


clementoriol:
**Description and expected behavior**

**Description**
Prisma does not support have first-class support for [Postgres generated columns](https://www.postgresql.org/docs/13/ddl-generated-columns.html) but they still kind of work if you : 

- Write the migration yourself
- Declare them in the prisma schema as optional with `@default(dbgenerated())` [(see the prisma docs)](https://www.prisma.io/docs/orm/reference/prisma-schema-reference#dbgenerated)
```
generatedField    String?                 @default(dbgenerated()) /// note the empty argument
```

**Issue**
However, zmodels don't seem to allow calling dbgenerated() without arguments. Running `zenstack generate` with the previous example will output the following error : 

```
missing argument for parameter "expr"
```

**Expected behavior**
zmodels should allow `@default(dbgenerated())` and generate the prisma schema accordingly

**Environment**

-   ZenStack version: 2.0.2
-   Prisma version: 5.10.2
-   Database type: Postgresql

ymc9:
Thanks for providing the fix @clementoriol !

ymc9:
Fixed in release 2.0.3.

PR #1401, opened 2024-05-01
fix: allow empty expr for dbgenerated() (#1400)

clementoriol:
Hey ! 
The fix for #1400 seemed easy enough (I mean who get the chance to submit a one char PR ? 😄) so here it goes.

The Prisma docs are not very clear on the subject but dbgenerated() can also accept no arguments for the use-case specified in the related ticket. (prisma docs: https://www.prisma.io/docs/orm/reference/prisma-schema-reference#dbgenerated)

However it does NOT supports empty strings but I don't think this can be expressed here.

Looking forward to any feedback.


ymc9:
Thank you for the fix @clementoriol ! Yes, Prisma's doc is a bit fuzzy about function details. I like that we can tweak to provide better flexibility. The change looks good to me and I'll merge it after CI passes. Will include it into a new patch release soon.

ymc9 (approved):

PR #1403, opened 2024-05-03
When using upsert the @default(auth().field) is not been handled

israelins85:


ymc9 (approved):

ymc9:
Thanks for working on this @israelins85 ! The change looks good to me and I'm merging it now.

Issue #1398, opened 2024-05-01
[Feature Request] Please export auth of enhance file


israelins85:
**Is your feature request related to a problem? Please describe.**
The method enhance requires user type that not exported.

**Describe the solution you'd like**
I'm needing define the same type.

**Describe alternatives you've considered**
Exporting the auth namespace can solve this.

**Additional context**
![image](https://github.com/zenstackhq/zenstack/assets/989344/73b1a989-0838-4d4c-8109-d915c168d5ae)
![image](https://github.com/zenstackhq/zenstack/assets/989344/e457846b-f477-4f72-9589-251b7780cd0d)

PR #1412, opened 2024-05-06
Export auth namespace on enhance.d.ts

israelins85:
To make easy use the enhance the correct type.


israelins85:
Solve #1398

packages/runtime/src/enhance.d.ts
```diff
@@ -1,2 +1,2 @@
 // @ts-expect-error stub for re-exporting generated code
-export { enhance } from '.zenstack/enhance';
+export { auth, enhance } from '.zenstack/enhance';
```
ymc9, line 2: Thanks for making this fix @israelins85 !

I believe the generated `auth` namespace also needs to be exported. 

https://github.com/zenstackhq/zenstack/blob/64e0c98acc4cd018c5169f3b325a3d4b617290ef/packages/schema/src/plugins/enhancer/enhance/auth-type-generator.ts#L109

Issue #1278, opened 2024-04-24
[Feature Request] Compatibility with Prisma "omit" feature


ymc9:
https://github.com/prisma/prisma/releases/tag/5.13.0

PR #1432, opened 2024-05-11
fix: compatibility with Prisma's "omit" feature

ymc9:
- Upgrade `@prisma/internals` and `@prisma/generator-helper` to latest.
- Respect `omit` clause when reading entities.
- Conditionally generate `createMany` based on Prisma version.


ymc9:
Fixes #1278

PR #1458, opened 2024-05-18
(Windows) build issues and contributing docs

WimTibackx:
I noticed some Windows-specific build failures and some gaps in the `CONTRIBUTING.md` docs.

I have the PR currently in draft mode because I am still doing some more checks on the tests, and I have yet to check the publishing. Felt it was sufficient to start opening it up to review, though.

I'd be happy to provide more details as needed.

This PR addresses the following:

- Note that PNPM version currently used is version 8.x, not last version

I haven't verified yet whether [PNPM 9.x](https://github.com/pnpm/pnpm/releases/tag/v9.0.0) works, but I figured it might be useful to specify the version reflected by the lockFileVersion. This part isn't really Windows specific, I hadn't really caught on to the Windows-specificness of my issues yet at this point :) .

- For windows users, document that git setting `core.symlinks` should be true

The symlinks in use in the repo don't work out of the box on Windows, but require a bit of Windows settings fiddling and git settings fiddling.

- Enforce the line endings as they are currently in use in the repo

This avoids Windows users executing `pnpm build` or adding new files and thereby introducing inconsistent line endings. Current settings in `.gitattributes` indicate that everything but `.bat` files use LF. I'm not sure whether the `.bat` file specifically _has_ to be CRLF, but given that it's currently the case I assumed it would have to be.
I ran `git add --update --normalize` to verify line endings, but all current files match this configuration.

- Make script arguments in package.json files play more nicely with Windows

Single quotes are a bit of an issue, unfortunately.  
In glob-style arguments I replaced the single quotes with escaped double quotes - which are already in use in some glob-style arguments.  
For the pack-destination param I removed the quotes, as some package.jsons already have that param without quotes - so I presumed it to be safe.

- Execute the correct `gradlew` script for the current OS

I added a dependency on `run-script-os` for this. This seems like a fairly sensible and lightweight way to handle it.

- Non-Windows specific: document that a postgres database is used in some tests and should be available for contributors executing tests.  

- Non-Windows specific: introduce environment variables to override default postgres connection details in integration tests.

In my case, I wanted to use a user other than the default `postgres` user, so I could have a bit more fine-grained control to the integration tests user. And honestly, so I wouldn't inadvertedly change the password to my root user and break my tests setup.  
This would also allow development on a local system, with postgres running within docker or setups like that.

- Ensure Windows paths are correctly handled in tests

Added some path normalization, because current processing was resulting in incorrect paths.

### DIFF
--- a/packages/testtools/src/db.ts
+++ b/packages/testtools/src/db.ts
@@ -1,16 +1,29 @@
 import { Pool } from 'pg';
 
-const USERNAME = 'postgres';
-const PASSWORD = 'abc123';
+const USERNAME = process.env.ZENSTACK_TEST_DB_USER || 'postgres';
+const PASSWORD = process.env.ZENSTACK_TEST_DB_PASS || 'abc123';
+const CONNECTION_DB = process.env.ZENSTACK_TEST_DB_NAME || 'postgres';
+const HOST = process.env.ZENSTACK_TEST_DB_HOST || 'localhost';
+const PORT = (process.env.ZENSTACK_TEST_DB_PORT ? parseInt(process.env.ZENSTACK_TEST_DB_PORT) : null) || 5432;
+
+function connect() {
+    return new Pool({ 
+        user: USERNAME, 
+        password: PASSWORD, 
+        database: CONNECTION_DB,
+        host: HOST,
+        port: PORT
+    });
+}
 
 export async function createPostgresDb(db: string) {
-    const pool = new Pool({ user: USERNAME, password: PASSWORD });
+    const pool = connect();
     await pool.query(`DROP DATABASE IF EXISTS "${db}";`);
     await pool.query(`CREATE DATABASE "${db}";`);
-    return `postgresql://${USERNAME}:${PASSWORD}@localhost:5432/${db}`;
+    return `postgresql://${USERNAME}:${PASSWORD}@${HOST}:${PORT}/${db}`;
 }
 
 export async function dropPostgresDb(db: string) {
-    const pool = new Pool({ user: USERNAME, password: PASSWORD });
+    const pool = connect();
     await pool.query(`DROP DATABASE IF EXISTS "${db}";`);
 }
--- a/packages/testtools/src/db.ts
+++ b/packages/testtools/src/db.ts
@@ -20,10 +20,12 @@ export async function createPostgresDb(db: string) {
     const pool = connect();
     await pool.query(`DROP DATABASE IF EXISTS "${db}";`);
     await pool.query(`CREATE DATABASE "${db}";`);
+    await pool.end();
     return `postgresql://${USERNAME}:${PASSWORD}@${HOST}:${PORT}/${db}`;
 }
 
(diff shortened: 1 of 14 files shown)


WimTibackx:
@coderabbitai review

packages/testtools/src/db.ts
```diff
@@ -1,16 +1,29 @@
...
+    });
+}
 
 export async function createPostgresDb(db: string) {
-    const pool = new Pool({ user: USERNAME, password: PASSWORD });
+    const pool = connect();
     await pool.query(`DROP DATABASE IF EXISTS "${db}";`);
     await pool.query(`CREATE DATABASE "${db}";`);
-    return `postgresql://${USERNAME}:${PASSWORD}@localhost:5432/${db}`;
+    return `postgresql://${USERNAME}:${PASSWORD}@${HOST}:${PORT}/${db}`;
```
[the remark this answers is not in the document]
WimTibackx: I was wondering about that indeed! Jest was warning for potentially unclosed resources as well. Will check!

WimTibackx:
As for the local publishing workflow, the suggestion in `CONTRIBUTING.md` to use Verdaccio seemed interesting to me so I tried it out. Two things I noted here:

- The `CONTRIBUTING.md` points to `publish-preview` being set up to publish to a local Verdaccio instance, but this no longer seems to be the case.
- When I manually ran more or less the same script but actually pointing to my local Verdaccio I bumped into the npm policy of not allowing republishes of the same version number. The package.jsons are currently on 2.1.1 which is an actually publically existing version number that I had probably pulled in via Verdaccio already; and so I couldn't publish that same version locally.

What I ended up doing was to first bump the package.json versions locally, and then publish with that:

```
pnpm -r exec pnpm version prerelease --no-git-tag-version && pnpm --filter "./packages/**" -r publish
```

This way, I could then install the prerelease version in my actual project and see my changes.
What's the current workflow you all are using for local development? Am I doing something wrong here, or is everyone using the linking workflow for local development?

packages/testtools/src/schema.ts
```diff
@@ -62,7 +62,7 @@ export function installPackage(pkg: string, dev = false) {
 }
 
 export function normalizePath(p: string) {
-    return p ? p.split(path.sep).join(path.posix.sep) : p;
+    return path.normalize(p);
```
WimTibackx, line 65: Based on some tests of the swr plugin I'm pretty sure this change is incorrect and I should instead ensure the following:

- Paths in zenstack model files use posix separators
- Paths for native filesystem operations are normalized using path.normalize to ensure proper handling of Windows drive prefixes

I have some local changes going in this direction, but I'm still verifying the whole test suite.

WimTibackx: The path handling was fine as is; I was misdirected by some of the expected-error logging and testcases failing for non-Windows specific reasons (which I'm purposefully not addressing in this branch)

jiashengguo (approved):
Looks good. Thanks for working on it!

PR #1465, opened 2024-05-25
Attempt at speeding up prisma-generator tests

WimTibackx:
In working on #1458 I noticed that some tests took me a while to execute. A lot of it seems to be centered around npm installs, so I picked out the prisma generator tests to experiment on.

As a baseline, current dev tests for that file took around 400 seconds on my system; with the npm installs taking up between 20 and 40 seconds.

With the `--no-progress --no-audit --no-fund` flags, total time for this file reduced to 300 seconds on my system.

I then looked at adding a test project directory, installing that in test-scaffold and copying from there each test. That reduced execution time to about 100 seconds on my system.

Given that this would seperate definition of the test context (which dependencies etc) from test execution, I wanted to try and avoid that if I could; so I looked at pnpm store. That is the version I finally submitted in this PR:
- Generally, we define a gitignored directory as a custom pnpm-store directory only to be used in the test suite.
- In beforeAll, we use a package.json and .npmrc written into a temp dir to load any missing dependencies into the store.
- In beforeEach, we can then write that same package.json and .npmrc and execute `pnpm install --offline` to quickly install from the store.
On my system this executes at around 55 seconds with an empty store, and at around 45 seconds when I've already populated the store earlier.

To me, this approach seems like a nice middle ground:
- These tests are slightly less isolated then they were before, as they rely on the pnpm store. While that's unfortunate, it isn't that different from what we already have with test-scaffold.
- Given that the project already uses pnpm, I'd expect it shouldn't be a problem to use pnpm commands within tests. I am somewhat wondering whether pnpm-installed-via-npm setups might be an issue?
- Using a pnpm store specific to the complete workspace test suite keeps it insulated from the rest of the system and the actual workings of the project
- The pnpm store seems slighly beneficial to me over the test-scaffold-and-copy I had before, given that the context of the test (package.json) isn't as separated from the actual test.

Happy to hear your thoughts on this! Is this an approach worth taking? Do you see downsides I'm missing here? Are there other factors to consider? Do you see room for further improvement?


script/test-utils.ts
```diff
@@ -0,0 +1,53 @@
...
+  "version":"1.0.0",
+  "dependencies": {
+    ${Object.entries(dependencies).map(([k, v]) => `"${k}": "${v}"`).join(',\n')}
+  },
+  "devDependencies": {
+    ${Object.entries(devDependencies).map(([k, v]) => `"${k}": "${v}"`).join(',\n')}
+  }
+}`;
+
+  initProjectDir(tmpDir, packageJsonContents, false);
```
WimTibackx, line 27: I considered doing a `pnpm store add` here instead of a plain install. While that worked, I decided against it in the end because it's a secondary way of processing the dependencies and I didn't see a significant downside to just installing and throwing the local project away right after.

script/test-utils.ts
```diff
@@ -0,0 +1,53 @@
...
+import { execSync } from 'node:child_process';
+
+export const PNPM_STORE_PATH = path.resolve(__dirname, '../.pnpm-test-store');
+export const NPM_RC_FILE = '.npmrc';
+export const NPM_RC_CONTENTS = `store-dir = ${PNPM_STORE_PATH}`;
+export const PACKAGE_JSON_FILE = 'package.json';
+export const PACKAGE_JSON_CONTENTS = '{"name":"test-project","version":"1.0.0"}';
+
+export function preparePackageJson(dependencies: {[key: string]: string} = {}, devDependencies: {[key: string]: string} = {}): string {
+  const tmpDir = fs.mkdtempSync(path.join(tmpdir(), 'zenstack-test-'));
```
WimTibackx, line 13: Given that this is a loose file included from elsewhere, I couldn't rely on the tmp package here and had to go with built-in node functions. Any downsides to this that I'm not considering?

(My thinking for including it this way was the developer experience of not needing to do a build step inbetween. Thoughts on this?)

packages/schema/tests/generator/prisma-generator.test.ts
```diff
@@ -6,22 +6,29 @@ import path from 'path';
...
 tmp.setGracefulCleanup();
 
 describe('Prisma generator test', () => {
     let origDir: string;
 
+    let packageJsonContents: string;
+
+    beforeAll(async () => {
+        packageJsonContents = preparePackageJson({
+            "prisma": "^5.0.0"
```
WimTibackx, line 21:
```suggestion
            'prisma': '^5.0.0'
```

script/test-utils.ts
```diff
@@ -0,0 +1,53 @@
...
+}`;
+
+  initProjectDir(tmpDir, packageJsonContents, false);
+
+  return packageJsonContents;
+  } finally {
+    fs.rmSync(tmpDir, {recursive: true, force: true});
+    console.log(`Loaded dependencies into store via temp dir ${tmpDir}`);
+  }
+}
```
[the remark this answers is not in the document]
WimTibackx: This feels a bit unnecessary to me, given that we're within tests anyway and that the native error should make it sufficiently clear what's going on.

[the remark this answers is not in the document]
harjotgill: @coderabbitai please respond to @WimTibackx 

(looks like auto chat reply setting is disabled on this repository).

ymc9:
Hi @WimTibackx, I'm sorry for the late response. Thanks for working on improving test speed. I've been bothered by it for quite a while but have been procrastinating on it 😅.

I think using a shared pnpm store is a great idea, which should make the installation process significantly faster, especially for running locally. From another perspective, I'm also wondering if we need to scaffold a npm project for each test at all. The tests were made this way mainly because Prisma and ZenStack both do code generation into node_modules and using disposable projects is the simplest way to isolate. However, since both Prisma and ZenStack allow specifying where to output the generated code, we can probably only emit them into some temp dir while not having a full-scale npm project at all.

What do you think about this direction? It's gonna be a bigger change though.

WimTibackx:
Hi @ymc9! It's about time I respond to this. Sorry for the delayed reply, I got a bit wrapped up in work and life for a bit.

> I think using a shared pnpm store is a great idea, which should make the installation process significantly faster, especially for running locally.

I've been looking into this last weekend, but it requires some further tinkering. My original proof of concept probably used the easiest set of dependencies, unfortunately :) . The bigger dependency trees and the post-install scripts require some further work for this PR. I'm taking another shot at it today, will keep you posted.

> I'm also wondering if we need to scaffold a npm project for each test at all. (...) since both Prisma and ZenStack allow specifying where to output the generated code, we can probably only emit them into some temp dir while not having a full-scale npm project at all.

That's a fair point indeed. As long as there's one full-on npm project integration test per module / variant, having other test cases run based on results emitting to a temp dir could suffice indeed.

For now, I'll see if I can universally apply the pnpm store approach, as that should already help a fair bit.

ymc9:
Closing for now. Will pick it up again as necessary.

PR #1534, opened 2024-06-25
Remove unused dependency `promisify`

benjamintd:
This removes the `promisify` dependency which is unused. 

This dependency itself depends on `"when": ""`, which, not declaring a version, breaks certain package managers like Bun.


ymc9:
Thanks for making the change @benjamintd ! It looks good to me and I'll merge once the CI passes.

ymc9 (approved):

benjamintd:
Thank you @ymc9 for updating the lockfile. I made the edit in the Github UI after checking that the dependency was unused, but did not get to installing the project on my machine to do it.

ymc9:
> Thank you @ymc9 for updating the lockfile. I made the edit in the Github UI after checking that the dependency was unused, but did not get to installing the project on my machine to do it.

No worries. I've also found another issue caused by not pinning Prisma dependency's minor version. Prisma 5.16 introduced a breaking change.

Issue #1537, opened 2024-06-27
[Feature Request] Add prop as optional parameter of getEnhancedPrisma


jasonmacdonald:
**Is your feature request related to a problem? Please describe.**
The `getEnhancedPrisma` call for the NestJS module does not provide any information about the requested property. This limits the ability to decide whether to return the enhanced Prisma client or the original. While you can inject the enhanced or non-enhanced service as separate injections, this feature would allow for a more nuanced approach using the same service.

**Describe the solution you'd like**
Pass the property being called to the `getEnhancedPrisma`.

**Additional context**
I will follow up on this request with a pull request.

PR #1538, opened 2024-06-27
feat(nestjs):  Add prop as optional parameter of getEnhancedPrisma #1537

jasonmacdonald:
Support for receiving the property used to trigger getEnhancedPrisma for the NestJS module. This is in relation to feature request #1537

### DIFF
--- a/packages/server/src/nestjs/zenstack.module.ts
+++ b/packages/server/src/nestjs/zenstack.module.ts
@@ -12,7 +12,7 @@ export interface ZenStackModuleOptions {
     /**
      * A callback for getting an enhanced `PrismaClient`.
      */
-    getEnhancedPrisma: (prop?: string | symbol) => unknown;
+    getEnhancedPrisma: (model?: string ) => unknown;
 }
 
 /**
--- a/packages/server/src/nestjs/zenstack.module.ts
+++ b/packages/server/src/nestjs/zenstack.module.ts
@@ -12,7 +12,7 @@ export interface ZenStackModuleOptions {
     /**
      * A callback for getting an enhanced `PrismaClient`.
      */
-    getEnhancedPrisma: () => unknown;
+    getEnhancedPrisma: (prop?: string | symbol) => unknown;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


jasonmacdonald:
Hmm, I'm not sure why those regressions are failing. I didn't touch anything related to the generated enhance file.

packages/server/src/nestjs/zenstack.module.ts
```diff
@@ -12,7 +12,7 @@ export interface ZenStackModuleOptions {
     /**
      * A callback for getting an enhanced `PrismaClient`.
      */
-    getEnhancedPrisma: () => unknown;
+    getEnhancedPrisma: (prop?: string | symbol) => unknown;
```
ymc9, line 15: Shall we rename `prop` to `model` instead so it's meaning is clearer? Also, I think in this case it's always typed as string. Maybe remove the `| symbol` part so the user can have a simpler callback signature?

ymc9:
Hi @jasonmacdonald , thanks for implementing this FR. The change looks good to me, and I've left a comment about the callback signature. Could you take a look?

I think the test failure is due to the current test scaffolding not pinning Prisma dependencies to their minor versions. Prisma 5.16 introduced a breaking change. I've made a fix to the dev branch. Could you merge it to your branch and push again? The errors should go awai then.

packages/server/src/nestjs/zenstack.module.ts
jasonmacdonald: Sounds good, I'll update this in this tomorrow and push it.

ymc9: Thank you!

jasonmacdonald:
@ymc9  I pushed the requested update. Sorry for the delay.

packages/server/src/nestjs/zenstack.module.ts
jasonmacdonald: So it looks like not including `symbol` throughs a type error. 

```
 Type 'symbol' is not assignable to type 'string'.
 ```
 https://github.com/zenstackhq/zenstack/actions/runs/9877836393/job/27280399900?pr=1538

PR #1629, opened 2024-08-01
Truncate too long delegate aux relation name in runtime

irvinzz:
# Issue
Long relation names cause any db operations failure


tests/integration/tests/enhancements/with-delegate/enhanced-client.test.ts
```diff
@@ -1196,6 +1196,28 @@ describe('Polymorphism Test', () => {
         );
     });
 
+    it('handles very long property names', async () => {
```
ymc9, line 1199: Hey @irvinzz , thanks for making this PR!

I tried this test case and found it actually passes both before and after this fix. Looking at the code, I believe the problem actually happens when the concrete model's name is very long (instead of a field). Is it what you're running into?

ymc9:
Hi @irvinzz , I think the current approach is a little brittle because the name shortening is totally based on a sequential suffix, and it can be unreliable at runtime, depending on what model names are processed first.

I think a more reliable way is to carry over the short name mapping to runtime (record them inside model metadata) so we can directly look it up instead of recomputing.

Do you mind if I make the changes directly on your branch?

irvinzz:
Hi @ymc9. Yes. Agree. It's brittle. I've investigated truncate behaviour in all cases and found that it looks useless to truncate field names. Especially in my issue. Fields of concrete models are 'virtual' (not used to communicate with real db) so don't really need to be truncated i guess. Keeped truncate for relation names and uniq constraints. I hope didn't missed something.

ymc9:
> Hi @ymc9. Yes. Agree. It's brittle. I've investigated truncate behaviour in all cases and found that it looks useless to truncate field names. Especially in my issue. Fields of concrete models are 'virtual' (not used to communicate with real db) so don't really need to be truncated i guess. Keeped truncate for relation names and uniq constraints. I hope didn't missed something.

I understand it's a bit confusion. ZenStack generates additional relation fields and constraints to facilitate polymorphic modeling, and these names can exceed the max length limits. The problem is that some of these fields are also used at runtime, so ideally, the mapping should be carried over from generation time to runtime.

I've made some changes based on your PR to implement the carry-over. It's a bit unfortunate that this introduced some more coupling among plugins, but I don't see a better way to address this without a bigger refactor.

PR #1641, opened 2024-08-08
add path arg to @@validate and gencode

j0rdanba1n:
# Problem

Form-level validation (done with `@@validate`, which zenstack will convert to `schema.refine()` in zod) would not be assigned to any specific field, making fetching the error message tricky, hacky, and less type safe. I had to find these errors using `formState.errors[""]`, since they had no key.

# Solution

Zod already has a solution for this, allowing you to pass a [path option into refine](https://zod.dev/?id=refine) to set the key this error should use / field that this validation error should be associated with.

In this pull request, I make zenstack compatible with this.

1) I add an optional  `_ path: FieldReference?` parameter to `@@validate` attributes
2) I use the values submitted here to generate code, such as `path: ['startDate',]` or `path: ['user', 'name']`
3) I merge the message and path options and pass it into the refine call in the generated code

# Testing

I tested this after running `pnpm zenstack generate` in my own project and it worked!

I can also add to the unit test suite. But wanted to submit this request for feedback first.


packages/plugins/trpc/tests/projects/t3-trpc-v10/package.json
```diff
@@ -20,10 +20,14 @@
     "@trpc/next": "^10.43.6",
     "@trpc/react-query": "^10.43.6",
     "@trpc/server": "^10.43.6",
+    "@zenstackhq/language": "file:../../../../../../.build/zenstackhq-language-2.4.1.tgz",
```
ymc9, line 23: Please exclude this file from the PR. I know it's awkward that whenever the test run this file is updated ... but it should be excluded.

packages/schema/src/res/stdlib.zmodel
```diff
@@ -633,7 +633,7 @@ attribute @lte(_ value: Int, _ message: String?) @@@targetField([IntField, Float
 /**
  * Validates the entity with a complex condition.
  */
-attribute @@validate(_ value: Boolean, _ message: String?) @@@validation
+attribute @@validate(_ value: Boolean, _ message: String?, _ path: FieldReference?) @@@validation
```
ymc9, line 636: I think this parameter should probably be of `String[]` type instead. The `FieldReference` is supposed to refer to a field in the current model (although this is not strictly checked yet); however, in the case of the validation issue path, maybe you'll want to refer to fields in the parent models?

packages/schema/src/plugins/zod/utils/schema-gen.ts
```diff
@@ -221,7 +222,12 @@ export function makeValidationRefinements(model: DataModel) {
             }
 
             const messageArg = getAttributeArgLiteral<string>(attr, 'message');
-            const message = messageArg ? `, { message: ${JSON.stringify(messageArg)} }` : '';
+            const message = messageArg ? `message: ${JSON.stringify(messageArg)},` : '';
+
+            const pathArg = getAttributeArg(attr, 'path');
```
ymc9, line 227: Please see my comments in the stdlib.zmodel part.

ymc9 (changes requested):
Hey @j0rdanba1n , many thanks for making this PR! I agree it's a very useful one.

I've added some comments. Also, if you would like to, the tests related to `@@validate` reside in "tests/integration/tests/enhancements/with-policy/field-validation.test.ts". I think a simple case covering the new parameter should be sufficient.

Thanks!

packages/schema/src/res/stdlib.zmodel
j0rdanba1n: agree, this will add more flexibility

j0rdanba1n:
@ymc9 addressed your feedback - please take another look and thank you :)

j0rdanba1n:
hi @ymc9 - following up :)

ymc9:
> hi @ymc9 - following up :)

Sorry for the late response @j0rdanba1n 
The change looks great to me. I'm merging it. Is it blocking you now? If so, we can make a new patch release.

ymc9 (approved):

j0rdanba1n:
@ymc9 we're not immediately blocked - when do you think the next patch release will be?

Issue #1672, opened 2024-08-30
Error: Dynamic Code Evaluation (e. g. 'eval', 'new Function', 'WebAssembly.compile') not allowed in Edge Runtime 


LilaRest:
**Description and expected behavior**
Hey there! I'm currently implement ZenStack in a NextJS project hosted on Cloudflare Pages (edge-runtime) so my entire application is set with:
```ts
export const runtime = "edge";
```

Development server works well, but as soon as I attempt a build, I obtain the error shown on the below screenshot. I tried to precisely follow the instructions from that [documentation page](https://zenstack.dev/docs/guides/edge), but the error persists.

Does anyone has an idea of what might be happening here?

**Screenshots**
![CleanShot 2024-08-30 at 19 25 35@2x](https://github.com/user-attachments/assets/0d279b22-effd-4471-a63d-7e80dac98f7b)

**Environment (please complete the following information):**

-   ZenStack version: 2.4.1
-   Prisma version: 5.19.0
-   Database type: Postgresql (Neon)

LilaRest:
I managed to silent the error by the following dirty patch adding to my `./src/middleware.ts` file:
```ts
export const config = {
  unstable_allowDynamic: [
    "/node_modules/**",
  ],
};
```

Now the app builds correctly and is deployed by Cloudflare Pages, but the entire website returns an "Error: 500", here is a screenshot of the logs:

![CleanShot 2024-08-30 at 20 32 13@2x](https://github.com/user-attachments/assets/ec25b9a9-d799-4f2b-8799-0e19169f52a7)

LilaRest:
After further investigation, I realize that both errors are the same. They refer to the usage of an evaluation code somewhere in the ZenStack enhancer.
https://nextjs.org/docs/messages/edge-dynamic-code-evaluation

I tried to find any of these in the ZenStack code using the VSCode search bar, but I haven't found anything yet.

LilaRest:
https://github.com/cloudflare/next-on-pages/issues/615

LilaRest:
I tried to run the enhanced ZenStack client in an API route and get an even more explicit error: `EvalError: Code generation from strings disallowed for this context`.

There is definitely a dynamic code evaluation made somewhere in ZenStack runtime or one of its dependencies. Again, using the raw Prisma client works perfectly.

Happy to pay some $$ to prioritize this issue :)

ymc9:
Fixed in 2.5.0

PR #1684, opened 2024-09-03
refactor(runtime): separate node and edge enhancement code for edge-runtime compatibility

ymc9:
The SAT solver "logical-solver" uses MiniSAT which uses "eval" thus breaking edge runtime. This change mutes the `check()` functionality when targeting edge runtime.

A "runtime" option is added to "@core/enhance" plugin. It defaults to "node" and can be set to "edge".


ymc9:
Fixes #1672

Issue #1748, opened 2024-09-27
[BUG] Generated API and openapi spec doesn't handle @@id correctly


thomassnielsen:
When using the openapi plugin (REST variant) with a schema using `@@id`, the attributes that make up the compound id are not added to the attribute list for the model.

Given the model

```prisma
model DashboardTiles extends Base {
  clientId Int
  tileType Int
  visible  Boolean
  Client   Client            @relation(fields: [clientId], references: [id])

  @@id([clientId, tileType])
  @@allow("all", check(Client))
}
```

The expected list of attributes should be

```ts
clientId: number
tileType: number
visible: boolean
```

Currently the list of attributes is only:

```ts
visible: boolean
```

-   ZenStack version: 2.5.1
-   Prisma version: 5.19.1
-   Database type: Postgresql

[Discord thread](https://discordapp.com/channels/1035538056146595961/1289126659932553236)

PR #1754, opened 2024-09-30
RESTApiHandler support compound ids (@@id)

thomassnielsen:
https://github.com/zenstackhq/zenstack/issues/1748

Adds support for the Prisma @@id compound ids to the RESTApiHandler.

[Discord thread](https://discordapp.com/channels/1035538056146595961/1289126659932553236/1289126659932553236)

### DIFF
--- a/packages/server/src/api/rest/index.ts
+++ b/packages/server/src/api/rest/index.ts
@@ -247,20 +247,13 @@ class RequestHandler extends APIHandlerBase {
                     let match = urlPatterns.single.match(path);
                     if (match) {
                         // single resource read
-                        return await this.processSingleRead(prisma, match.type, match.id, query, modelMeta);
+                        return await this.processSingleRead(prisma, match.type, match.id, query);
                     }
 
                     match = urlPatterns.fetchRelationship.match(path);
                     if (match) {
                         // fetch related resource(s)
-                        return await this.processFetchRelated(
-                            prisma,
-                            match.type,
-                            match.id,
-                            match.relationship,
-                            query,
-                            modelMeta
-                        );
+                        return await this.processFetchRelated(prisma, match.type, match.id, match.relationship, query);
                     }
 
                     match = urlPatterns.relationship.match(path);
@@ -271,15 +264,14 @@ class RequestHandler extends APIHandlerBase {
                             match.type,
                             match.id,
                             match.relationship,
-                            query,
-                            modelMeta
+                            query
                         );
                     }
 
                     match = urlPatterns.collection.match(path);
                     if (match) {
                         // collection read
-                        return await this.processCollectionRead(prisma, match.type, query, modelMeta);
+                        return await this.processCollectionRead(prisma, match.type, query);
                     }
 
                     return this.makeError('invalidPath');
@@ -306,8 +298,7 @@ class RequestHandler extends APIHandlerBase {
                             match.id,
                             match.relationship,
                             query,
-                            requestBody,
-                            modelMeta
+                            requestBody
                         );
                     }
 
@@ -345,8 +336,7 @@ class RequestHandler extends APIHandlerBase {
                             match.id,
                             match.relationship as string,
                             query,
-                            requestBody,
-                            modelMeta
+                            requestBody
                         );
                     }
 
@@ -370,8 +360,7 @@ class RequestHandler extends APIHandlerBase {
                             match.id,
                             match.relationship as string,
                             query,
-                            requestBody,
-                            modelMeta
+                            requestBody
                         );
                     }
 
@@ -394,8 +383,7 @@ class RequestHandler extends APIHandlerBase {
         prisma: DbClientContract,
         type: string,
         resourceId: string,
-        query: Record<string, string | string[]> | undefined,
-        modelMeta: ModelMeta
+        query: Record<string, string | string[]> | undefined
     ): Promise<Response> {
         const typeInfo = this.typeMap[type];
         if (!typeInfo) {
@@ -420,16 +408,12 @@ class RequestHandler extends APIHandlerBase {
             include = allIncludes;
         }
 
-        let entity = await prisma[type].findUnique(args);
-
-        if (typeInfo.idFields.length > 1) {
-            entity = { ...entity, [this.makeIdKey(typeInfo.idFields)]: resourceId };
-        }
+        const entity = await prisma[type].findUnique(args);
 
         if (entity) {
             return {
                 status: 200,
-                body: await this.serializeItems(type, entity, modelMeta, { include }),
+                body: await this.serializeItems(type, entity, { include }),
             };
         } else {
             return this.makeError('notFound');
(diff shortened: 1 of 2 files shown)


packages/server/src/api/rest/index.ts
```diff
@@ -32,6 +32,9 @@ const urlPatterns = {
     relationship: new UrlPattern('/:type/:id/relationships/:relationship'),
 };
 
+export const idDivider = '_';
+export const compoundIdKey = 'compoundId';
```
ymc9, line 36: I'm wondering if we actually need this field to be a fixed name. Can we just use Prisma's convention (id1_id2_...) as the compound key field name?

packages/server/src/api/rest/index.ts
```diff
@@ -59,9 +62,13 @@ type RelationshipInfo = {
     isOptional: boolean;
 };
 
+type IdField = {
```
ymc9, line 65: You can use the `FieldInfo` type from `@zenstackhq/runtime`

packages/server/src/api/rest/index.ts
```diff
@@ -1178,18 +1182,31 @@ class RequestHandler extends APIHandlerBase {
...
     }
 
-    private makeIdFilter(idField: string, idFieldType: string, resourceId: string) {
-        return { [idField]: this.coerce(idFieldType, resourceId) };
+    private makeIdFilter(idFields: IdField[], resourceId: string) {
+        if (idFields.length === 1) {
+            return { [idFields[0].name]: this.coerce(idFields[0].type, resourceId) };
+        } else {
+            return {
+                [idFields.map((idf) => idf.name).join('_')]: idFields.reduce(
```
ymc9, line 1190: Prisma actually allows to you name a compound id field like @@id("myCompountId", ...). When used that way, the default compound key formatting actually won't work.

Supporting it will require carrying more information about @@id in model metadata and will require some changes in the code gen part. I can make a separate follow up PR for it later.

packages/server/src/api/rest/index.ts
```diff
@@ -1178,18 +1182,31 @@ class RequestHandler extends APIHandlerBase {
...
+                    (acc, curr, idx) => ({
+                        ...acc,
+                        [curr.name]: this.coerce(curr.type, resourceId.split(idDivider)[idx]),
```
ymc9, line 1193: This means the id values cannot contain the divider.

I like the fact that if we join id fields with "_", the URL will be easy to read and understand. A more reliable approach is to keep it as an object and URL encode it, but it'll look quite ugly.

I'm a bit uncertain about the tradeoff ... what do you think 😄?

packages/server/src/api/rest/index.ts
```diff
@@ -405,7 +408,12 @@ class RequestHandler extends APIHandlerBase {
             include = allIncludes;
         }
 
-        const entity = await prisma[type].findUnique(args);
+        let entity = await prisma[type].findUnique(args);
+
+        if (typeInfo.idFields.length > 1) {
+            entity = { ...entity, [compoundIdKey]: resourceId };
```
ymc9, line 414: It seems the compound id field is only included when reading a single entity. Should it be included when reading a collection or via a relation too?

ymc9 (changes requested):
Hi @thomassnielsen , many thanks for making this PR. It's a great start! I've left some comments there, please help take a look.

I'm not sure how much time you have to continue working on this. Another option is we merge it first, and I can follow up with a separate PR to continue polishing it.

I'd love to include it in the next release!

packages/server/src/api/rest/index.ts
thomassnielsen: Good catch. Should we keep it like this for now in this PR then?

packages/server/src/api/rest/index.ts
thomassnielsen: I had the same thought. I ended up on this setup for two reasons. First it matches the Prisma default, secondly it works without changing the URLPattern config, since `_` is in the [default list of accepted characters](https://github.com/snd/url-pattern?tab=readme-ov-file#named-segments). 

If it turns out to be problematic in practice this may need to be changed. Either like you say an URL encoded object, or maybe making the id divider configurable.

packages/server/src/api/rest/index.ts
thomassnielsen: Yes it should always be included I think. Will look at that when adding tests for those situations.

ymc9:
> > I'm not sure how much time you have to continue working on this. Another option is we merge it first, and I can follow up with a separate PR to continue polishing it.
> 
> 
> 
> I have some time today and tomorrow so I'll try to fix at least what's been discussed in this PR and add some more tests. When's the next release cutoff point?

Thank you! Current goal of v2.7 is 10/15. Please take your time 😄.

packages/server/src/api/rest/index.ts
ymc9: Yes, I think we can keep it like this for now.

packages/server/src/api/rest/index.ts
ymc9: I don't have a strong opinion about it. I like keeping the URL clean and readable. Maybe as you said making the id divider configurable (in the future) is a good solution. Let's keep it as is for now.

thomassnielsen:
Marking this ready for review now and moving on to looking at the openapi plugin.

packages/server/src/api/rest/index.ts
```diff
@@ -1178,18 +1214,35 @@ class RequestHandler extends APIHandlerBase {
...
+            return {
+                [idFields.map((idf) => idf.name).join('_')]: idFields.reduce(
+                    (acc, curr, idx) => ({
+                        ...acc,
+                        [curr.name]: this.coerce(curr.type, resourceId.split(idDivider)[idx]),
+                    }),
+                    {}
+                ),
+            };
+        }
```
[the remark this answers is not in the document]
thomassnielsen: 🤖🐰  This is straight up wrong. Compound ids need to be nested under the name given to the compound id, which by default is the field names joined by `'_'`. Noticed that I forgot to use the divider const here though, so fixing that.

thomassnielsen:
Testing with an updated openapi plugin it looks like there's something still missing here, so converting it back to a draft.

packages/server/tests/api/test.zmodel
ymc9, line 1: Should this file be deleted?

packages/server/tests/api/rest.test.ts
```diff
@@ -1546,6 +1637,54 @@ describe('REST server tests', () => {
...
+                                attributes: { userId: 'user1', postId: 1, superLike: false },
+                            },
+                        },
+                        prisma,
+                    });
+
+                    expect(r.status).toBe(201);
+                });
+
+                describe('compound id', () => {
```
ymc9, line 1662: Is this new level of nesting intentional?

ymc9 (changes requested):
Thanks for continuing to work on this @thomassnielsen ! I've left two minor comments and the code overall looks very good to me now.

packages/server/tests/api/rest.test.ts
thomassnielsen: Nope, forgot to remove as some tests ended up elsewhere.

ymc9 (approved):
Looking good now! I'll merge it after CI passes. Thank you!

PR #1773, opened 2024-10-14
feat: create Zod's `default` function for entity's booleans

Arkanii:
I'm using [Superforms](https://superforms.rocks/default-values#default-values) in my Svelte project and it use the `default` function from the Zod schema to fulfill the default value of the form.

It was not implemented in the Zod plugin, so here is my PR !


ymc9:
Thank you for the fix @Arkanii ! It looks to me and I'll merge it after the CI passes. I'll be part of the upcoming 2.7 release.

ymc9 (approved):

PR #1791, opened 2024-10-21
URI decode resourceId 

thomassnielsen:
Decode resourceId so encoded characters (e.g. `:`) can be user as id divider.


ymc9 (approved):
Thanks! Approved. I'll make a new patch release later today.

PR #1798, opened 2024-10-22
Include relationships in REST result from creates and updates

thomassnielsen:
Here are two test cases: For GET requests, we do get the expected empty data array in the relationship response. However, in the response of a PUT request, the data array is missing and the test fails before the included change.


ymc9 (approved):
Good catch! Merging and will make a release today. Thanks!

PR #1801, opened 2024-10-23
Improve working with entities related to entities that uses compound ids

thomassnielsen:
1. Fixes an issue where adding entities related to entities with compound id (see `makeIdConnect`)
1. Fixes validation of entities with compound id, given that zod schemas expect the attributes separately
  1. This is not super elegant, happy to get suggestions on improvements
1. Update openAPI spec generation to match the previous point, making generate clients behave as expected

### DIFF
--- a/packages/server/src/api/rest/index.ts
+++ b/packages/server/src/api/rest/index.ts
@@ -723,7 +723,7 @@ class RequestHandler extends APIHandlerBase {
         if (parsed.data.relationships) {
             for (const [key, data] of Object.entries<any>(parsed.data.relationships)) {
                 const typeInfo = this.typeMap[key];
-                if (typeInfo.idFields.length > 1) {
+                if (typeInfo && typeInfo.idFields.length > 1) {
                     typeInfo.idFields.forEach((field, index) => {
                         attributes[field.name] = this.coerce(field.type, data.data.id.split(this.idDivider)[index]);
                     });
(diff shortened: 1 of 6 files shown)


packages/server/src/api/rest/index.ts
```diff
@@ -719,6 +719,18 @@ class RequestHandler extends APIHandlerBase {
         const parsed = this.createUpdatePayloadSchema.parse(body);
         const attributes: any = parsed.data.attributes;
 
+        // Map in the compound id relationships as attributes, as they are expected by the zod schema
```
ymc9, line 722: Hi @thomassnielsen , I think we should probably validate using a zod schema that doesn't include fk fields here. The purpose should be validating rules declared on the data fields.

Shall I make some tweaks directly?

thomassnielsen: That makes sense. Yes, go ahead and make the changes. I'll read them to learn afterwards 😄

ymc9:
Hi @thomassnielsen , I've changed to use the "scalar" schema to validate which doesn't include FK fields. Also had to make a fix in the zod generation part to make sure ID fields are always included in the schema even they're also FK fields. May different corner combinations 😄.

Please take a look, thanks!

thomassnielsen:
Nice! I'm not familiar with the zod generator, but this looks reasonable to me 👍

ymc9:
> Nice! I'm not familiar with the zod generator, but this looks reasonable to me 👍

Cool. I'm merging it now.

ymc9 (approved):

PR #1806, opened 2024-10-24
Fix for filtering by compound id

thomassnielsen:
In some cases Prisma doesn’t expect these to be nested under the compound id key. I don't have a full overview of where this is (just the example I hit myself), so I added it just where I can confirm it's needed.

### DIFF
--- a/packages/server/src/api/rest/index.ts
+++ b/packages/server/src/api/rest/index.ts
@@ -1602,8 +1602,8 @@ class RequestHandler extends APIHandlerBase {
                 const values = value.split(',').filter((i) => i);
                 const filterValue =
                     values.length > 1
-                        ? { OR: values.map((v) => this.makePrismaIdFilter(info.idFields, v)) }
-                        : this.makePrismaIdFilter(info.idFields, value);
+                        ? { OR: values.map((v) => this.makePrismaIdFilter(info.idFields, v, false)) }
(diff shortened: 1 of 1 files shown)


packages/server/src/api/rest/index.ts
```diff
@@ -1598,7 +1606,7 @@ class RequestHandler extends APIHandlerBase {
                         : this.makePrismaIdFilter(info.idFields, value);
                 return { some: filterValue };
             } else {
-                return { is: this.makePrismaIdFilter(info.idFields, value) };
+                return { is: this.makePrismaIdFilter(info.idFields, value, false) };
```
ymc9, line 1609: Good catch.

I believe Prisma's rule is that the `id1_id2: { ... }` form of filter is only required in a "unique where clause" (used in `findUnique`, `update`, etc).

The filter made by the `makeFilterValue` function are not used in the unique context, so I believe all `makePrismaIdFilter` call inside this function should have `false` as the second parameter.

ymc9 (changes requested):

packages/server/src/api/rest/index.ts
thomassnielsen: That makes sense. Made this change now.

ymc9 (approved):

Issue #784, opened 2023-10-23
[Feature Request] Strong-typed JSON fields


ymc9:
## Background

Related Pisma issue: https://github.com/prisma/prisma/issues/3219

## Proposed Solution

1. Add language-level "type" declaration support (not mapped to database tables) #783
2. Enforce schema checking at runtime for the strongly-typed fields through enhanced PrismaClient

ymc9:
Fixed in [2.8.0](https://github.com/zenstackhq/zenstack/releases/tag/v2.8.0)

Issue #783, opened 2023-10-23
[Feature Request] Support for non-model types


ymc9:
We can introduce a pure "type" construct into ZModel to complement the "model" construct. Contrary to "model", "type"s are not mapped to database tables and are purely for type-checking purposes.

Here are some scenarios that can be enabled by type:

1. Strong-typed JSON field

    ```prisma
    type Profile {
        bio String
        avatar String
    }

    model User {
        ...
        profile Profile @json
    }
    ```

2. Type-checking for plugin options

    Today, plugin options are just plain arbitrary properties, and there's no parse-time checking at all. With the "type" construct, plugins can define their own options type, and users can enjoy parse-time error reporting and IDE auto-completion.

3. More flexible `auth()` typing

    Today, the typing of `auth()` function is confined to the `User` model. To use something out of it, you'll have to use a hack: define a field in `User` but mark it as `@ignore`. With "type" construct, one can use a type (completely decoupled from database tables) to back `auth()`.

    ```prisma
    type MyAuth {
        ...
        @@auth()
    }
    ```

Other notes:
- Prisma already uses the "type" keyword, but it's for MongoDB only. This feature will repurpose this keyword for relational databases.

AmruthPillai:
Wouldn't it be wise to use another keyword other than `type` as Prisma might have plans on using the same keyword for relational databases, but for a different construct? Something like `abstract auth { }` or even `auth User { }` comes to mind.

PR #1813, opened 2024-10-29
feat(zmodel): "type" construct and strongly-typed Json fields

ymc9:
Initial implementation of #784 and #783 

- `type` declaration
- Zod generation with validation attributes support like `@email`
- Create/update payload validation
- Proper return type from enhanced PrismaClient
- Support for `@default` (including using `auth() in it)

Out of scope:

- Inheritance
- Type-level validation via `@@validate`


packages/schema/src/plugins/enhancer/enhance/index.ts
```diff
@@ -493,10 +504,72 @@ export function enhance(prisma: any, context?: EnhancementContext<${authTypePara
...
+            return source.replace(
+                new RegExp(`(${field.name}\\??\\s*):[^\\n]+`),
+                `$1: ${field.type.reference!.$refText}${field.type.array ? '[]' : ''}${
+                    field.type.optional ? ' | null' : ''
+                }`
+            );
+        };
+
+        // fix "$[Model]Payload" type
+        const payloadModelMatch = modelsWithTypeField.find((m) => `$${m.name}Payload` === typeName);
```
nwidynski, line 534: Instead of replacement, could we slightly alter the type construction, to support type augmentation?

https://github.com/prisma/prisma/issues/3219#issuecomment-2400772675

ymc9: Hi @nwidynski , I'm glad you found the work here 😄.

I think your proposal is a good approach to making the model types easily customizable. Can it be simplified to the following (without the nested `$user`)?

```ts
export interface $UserExtensionScalar {}

export interface $UserScalar {
    // original scalar fields
    ...
}

export type $UserPayload<ExtArgs extends $Extensions.InternalArgs = $Extensions.DefaultArgs> = {
    ...
    scalars: $Extensions.GetPayloadResult<
      Omit<$UserScalar, keyof $UserExtensionScalar> & $UserExtensionScalar,
      ExtArgs["result"]["user"]>
}


// user land
import '@prisma/client';

declare module '@prisma/client' {
    namespace Prisma {
        export interface $UserExtensionScalar {
            metadata: { ... }
        }
    }
}
```

Curious about your use case: when ZenStack's native support for typed-json-field is available, do you still need this augmentation-based extension?

nwidynski: Hey,

unfortunately the nested property is necessary as typescript can’t override existing declarations but only patch them. In order to make conditionals on those patches, a known property must be patched. It can have any name though, it doesnt have to be $user.

As for your second question, yes I would still have a use case for them in order to get strongly typed model results while they are separated in different extensions. This currently doesnt work well, especially for nested includes.

ymc9: Got it. For the field-type patching, I was thinking of "omit then patch": `Omit<$UserScalar, keyof $UserExtensionScalar> & $UserExtensionScalar`. I haven't tried it yet though.

My current goal is to make the "type" declaration a preview feature in v2.8.0 to get more feedback about the direction. I think we can incorporate your suggestion after the general approach is confirmed viable.

nwidynski: Yes, that delivers additional convenience, although the $UserScalar could also be reused in user land to build the extension.

PR #1846, opened 2024-11-12
fix: ignore file from typechecking

genu:
Generating the zod schemas in a project causes the schemas to be subject to typechecking.

There can be cases where the zod plugin adds an import which doens't end up being used. This raises typechecking errors.

I don't see a reason why these files need to be typechecked.

### DIFF
--- a/packages/schema/src/plugins/zod/generator.ts
+++ b/packages/schema/src/plugins/zod/generator.ts
@@ -326,6 +326,7 @@ export class ZodSchemaGenerator {
 
     private addPreludeAndImports(decl: DataModel | TypeDef, writer: CodeBlockWriter, output: string) {
         writer.writeLine('/* eslint-disable */');
+        writer.writeLine('// @ts-nocheck');
         writer.writeLine(`import { z } from 'zod';`);
 
         // import user-defined enums from Prisma as they might be referenced in the expressions


ymc9:
Hi @genu , thanks for making the PR. Is this the only error you see with the generated code? I'm thinking if we should actually prevent the error from being there.

genu:
So far in my testing, that is the only error I see with the generated code.

Fixing the error would be resolve it, true.

Does type checking in these files have any benefit though?

ymc9:
> So far in my testing, that is the only error I see with the generated code.
> 
> Fixing the error would be resolve it, true.
> 
> Does type checking in these files have any benefit though?

The generated code is supposed to always pass typechecking 😄, but just in case there is a real problem, I think it's still nice for typescript compiler to catch it before failing at runtime ... what do you think?

genu:
Hmm, well, my thought process is that this is generated code, so even if there are type issues, there is nothing that I would be  able to do about it.

The type checks would maybe make more sense during testing on the zenstack side? 🤔

ymc9:
> Hmm, well, my thought process is that this is generated code, so even if there are type issues, there is nothing that I would be able to do about it.
> 
> The type checks would maybe make more sense during testing on the zenstack side? 🤔

Right, if it fails it's a zenstack bug. With a second thought, I think we can proceed with adding ts-nocheck first.

ymc9:
Merging it now. Will include it in the next patch release. Thank you!

PR #1855, opened 2024-11-14
fix(openapi): bugfix for compound id create and update

thomassnielsen:
When creating a record with a compound id, Prisma expects the parts of the compound id to be passed as attributes. This change reflects that.

### DIFF
--- a/packages/plugins/openapi/src/rest-generator.ts
+++ b/packages/plugins/openapi/src/rest-generator.ts
@@ -860,7 +860,7 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
         // For compound ids each component is also exposed as a separate fields for read operations,
         // but not required for write operations
         const fields =
-            idFields.length > 1 && mode === 'read' ? model.fields : model.fields.filter((f) => !isIdField(f));
+            idFields.length > 1 && mode !== 'update' ? model.fields : model.fields.filter((f) => !isIdField(f));
 
         const attributes: Record<string, OAPI.SchemaObject> = {};
         const relationships: Record<string, OAPI.ReferenceObject | OAPI.SchemaObject> = {};
@@ -907,7 +907,7 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
         if (mode === 'create') {
             // 'id' is required if there's no default value
             const idFields = model.fields.filter((f) => isIdField(f));
-            if (idFields.length && idFields.every((f) => !hasAttribute(f, '@default'))) {
+            if (idFields.length === 1 && idFields.every((f) => !hasAttribute(f, '@default'))) {
                 properties = { id: { type: 'string' }, ...properties };
                 toplevelRequired.unshift('id');
             }
--- a/packages/plugins/openapi/src/rest-generator.ts
+++ b/packages/plugins/openapi/src/rest-generator.ts
@@ -905,7 +905,7 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
         if (mode === 'create') {
             // 'id' is required if there's no default value
             const idFields = model.fields.filter((f) => isIdField(f));
-            if (idFields.length === 1 && idFields.every((f) => !hasAttribute(f, '@default'))) {
+            if (idFields.length === 1 && !hasAttribute(idFields[0], '@default')) {
                 properties = { id: { type: 'string' }, ...properties };
                 toplevelRequired.unshift('id');
             }
--- a/packages/plugins/openapi/src/rest-generator.ts
+++ b/packages/plugins/openapi/src/rest-generator.ts
@@ -857,10 +857,8 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
 
     private generateModelEntity(model: DataModel, mode: 'read' | 'create' | 'update'): OAPI.SchemaObject {
         const idFields = model.fields.filter((f) => isIdField(f));
-        // For compound ids each component is also exposed as a separate fields for read operations,
-        // but not required for write operations
-        const fields =
-            idFields.length > 1 && mode !== 'update' ? model.fields : model.fields.filter((f) => !isIdField(f));
+        // For compound ids each component is also exposed as a separate fields.
+        const fields = idFields.length > 1 ? model.fields : model.fields.filter((f) => !isIdField(f));
 
         const attributes: Record<string, OAPI.SchemaObject> = {};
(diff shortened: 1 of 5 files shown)


packages/plugins/openapi/src/rest-generator.ts
```diff
@@ -907,7 +907,7 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
         if (mode === 'create') {
             // 'id' is required if there's no default value
             const idFields = model.fields.filter((f) => isIdField(f));
-            if (idFields.length && idFields.every((f) => !hasAttribute(f, '@default'))) {
+            if (idFields.length === 1 && idFields.every((f) => !hasAttribute(f, '@default'))) {
```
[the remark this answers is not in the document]
thomassnielsen: This logic happens elsewhere.

ymc9, line 910: shall we simplify this to `hasAttribute(idFields[0], '@default')`?

packages/plugins/openapi/src/rest-generator.ts
```diff
@@ -860,7 +860,7 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
         // For compound ids each component is also exposed as a separate fields for read operations,
         // but not required for write operations
         const fields =
-            idFields.length > 1 && mode === 'read' ? model.fields : model.fields.filter((f) => !isIdField(f));
+            idFields.length > 1 && mode !== 'update' ? model.fields : model.fields.filter((f) => !isIdField(f));
```
ymc9, line 863: I just checked JSON:API spec again, the convention for CREATE with client-generated ID is to include it in the top-level id field. https://jsonapi.org/format/#crud-creating-client-ids

Do you feel we should actually exclude the id fields for create here, and extract it from the top-level id? So we don't need to duplicate them in the payload. Is that actually the current behavior of the RESTAPIHandler?

ymc9: Also I'm starting to think about the "update" case here 😄. You can actually update id fields with an update operation, right?

thomassnielsen: > Do you feel we should actually exclude the id fields for create here, and extract it from the top-level id? So we don't need to duplicate them in the payload. Is that actually the current behavior of the RESTAPIHandler?

This change is made to reflect the implementation of the RESTAPIHandler (and how Prisma works). We could go the other way instead though, and update the RESTAPIHandler. I'll think a bit and discuss with some colleagues who are using this API.

As for the spec I feel like this is a bit ambiguous in this case. Technically the ID is generated on the server based on the fields. The same goes for update - you update the field and the id changes.

packages/plugins/openapi/src/rest-generator.ts
thomassnielsen: That's not the same behaviour though? That would set the to level required 'id' for compound ids as well.

packages/plugins/openapi/src/rest-generator.ts
thomassnielsen: After having a quick chat I think our position is that the compound id is generated server side and as such this proposed change does not break with JSON:API. I'm however very open to be convinced otherwise.

I'll dig a bit further into the behaviour of the RESTAPIHandler and make sure that this and #1856 doesn't cause any mismatches.

packages/plugins/openapi/src/rest-generator.ts
ymc9: Oh, sorry I meant:

```ts
idFields.length === 1 && !hasAttribute(idFields[0], '@default')
```

Just feeling the `.every` call is not needed anymore

packages/plugins/openapi/src/rest-generator.ts
ymc9: I tried the current rest handler implementation , for "create", yes, you'll need to set the compound id fields in the "attributes" object. The top-level "id" field is ignored. I think this is OK (and probably more natural). I can update the docs to reflect this (as it deviates from JSON:API spec). This applies to single-id models as well.

For "update", it behaves similarly. The model id(s) can be updated by setting id fields directly inside the "attributes" object. Top-level "id" is ignored.

So it feels like to keep the spec and runtime consistent, we should unconditionally include id fields (regardless compound or not) for "create" and "update". For "read", only include id fields when it's compound?

To be honest, I feel things will be cleaner if we just always include the raw id fields for read and mutation in the rest handler implementation 😂

What do you think @thomassnielsen ?

thomassnielsen: I'm thinking that id should always be present in reads and required in updates, but may not be required in writes, depending on wether the id is generated (auto increment or compound) or provided by the client (typically UUID).

In particular this part of the spec allows for this.

> Every resource object MUST also contain an id member, except when the resource object originates at the client and represents a new resource to be created on the server.

So basically always include id in responses, and require id except if the model has either `@id @default(autoincrement())` or `@@id`.

ymc9: I think from the response point of view we're good. A top-level "id" is already included, and for the case of compound ids, each component id is also included in "attributes".

For "create" and "update", what I meant was (I believe) currently the rest handler isn't fully compliant with the JSON:API spec, in that:

1. It doesn't check if the top-level "id" exists
2. It doesn't use the top-level "id", instead id value is always fetched from "attributes"

So, if you do:

```
POST 
{
  "id": "1",
  "type": "Post",
  "attributes": { "id": 2 }
 }
```

You'll end up creating a `Post` with id 2 😅.

I think it's probably OK for now. Changing it will be a breaking change.

ymc9: Your current change specifically checks `mode !== 'update'`, which means for compound id models, their update request input will not have the id component fields. There won't be a way to update the id fields then, right?

thomassnielsen: Yes for reads the top level id should work according to the spec for models that don't generate ids. That's a bug but as you say also a breaking change. Should we create an issue to keep track of this?

> Your current change specifically checks mode !== 'update', which means for compound id models, their update request input will not have the id component fields. There won't be a way to update the id fields then, right?

Yes that wasn't the right solution. I removed the mode !== 'update' check in the latest commit.

ymc9 (approved):

thomassnielsen:
@ymc9 could we make sure this makes it into the next release? It's a blocking issue on our end.

ymc9:
Hi @thomassnielsen , v2.9.4 is live now!

PR #1856, opened 2024-11-14
Don't expose foreign keys as attributes in openAPI spec

thomassnielsen:
These attributes are handled as relationships, and as such don't need to be exposed as attributes.

### DIFF
--- a/packages/plugins/openapi/src/rest-generator.ts
+++ b/packages/plugins/openapi/src/rest-generator.ts
@@ -868,7 +868,7 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
         const required: string[] = [];
 
         for (const field of fields) {
-            if (isForeignKeyField(field)) {
+            if (isForeignKeyField(field) && mode !== 'read') {
                 // foreign keys are not exposed as attributes
                 continue;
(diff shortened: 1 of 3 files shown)


packages/plugins/openapi/src/rest-generator.ts
```diff
@@ -868,6 +868,10 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
         const required: string[] = [];
 
         for (const field of fields) {
+            if (isForeignKeyField(field)) {
```
ymc9, line 871: Hi @thomassnielsen , I believe foreign key fields are actually returned by the RESTAPIHandler today, so changing the spec here will cause inconsistency?

Do you find returning the fk fields to cause trouble?

ymc9 (commented):
Hi @thomassnielsen , thanks for continue improving this. Please see my comments.

packages/plugins/openapi/src/rest-generator.ts
thomassnielsen: Looks like this only applies to writes. Added a check for mode now.

ymc9 (approved):

PR #1863, opened 2024-11-15
feat(server): upsert support for rest api handler

thomassnielsen:
Adds support for upsert operations in the RESTAPIHandler via metadata. 

We may also want to add documentation about this to the OpenAPI generator.

[Discussion in Discord](https://discord.com/channels/1035538056146595961/1087761218049167380/1299078053653123165)

### DIFF
--- a/packages/server/src/api/rest/index.ts
+++ b/packages/server/src/api/rest/index.ts
@@ -776,12 +776,25 @@ class RequestHandler extends APIHandlerBase {
         }
 
         let entity: any;
-        if (operation === 'upsert') {
-            entity = await this.runUsert(typeInfo, type, prisma, attributes, relationships, matchFields);
-        } else if (operation === 'create') {
-            entity = await this.runCreate(typeInfo, type, prisma, attributes, relationships);
-        } else {
-            return this.makeError('invalidPayload');
+
+        try {
+            if (operation === 'upsert') {
+                entity = await this.runUpsert(
+                    typeInfo,
+                    type,
+                    prisma,
+                    modelMeta,
+                    attributes,
+                    relationships,
+                    matchFields
+                );
+            } else if (operation === 'create') {
+                entity = await this.runCreate(typeInfo, type, prisma, attributes, relationships);
+            } else {
+                return this.makeError('invalidPayload');
+            }
+        } catch (e) {
+            return e as any;
         }
 
         return {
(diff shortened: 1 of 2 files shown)


packages/server/src/api/rest/index.ts
```diff
@@ -756,12 +764,111 @@ class RequestHandler extends APIHandlerBase {
...
+            entity = await this.runUpsert(typeInfo, type, prisma, modelMeta, attributes, relationships, matchFields);
+        } else if (operation === 'create') {
+            entity = await this.runCreate(typeInfo, type, prisma, attributes, relationships);
+        } else {
+            return this.makeError('invalidPayload');
+        }
+
+        if (entity.status) {
+            return entity;
+        }
```
thomassnielsen, line 790: This is not very elegant. I tried throwing the errors from within each of the upsert and create functions, but that caused some weird regressions with a few tests getting "undefined" from the API handler. I also didn't want to inline the upsert path here since it's quite verbose. I may want to refactor this a bit with fresh eyes another day.

thomassnielsen:
I refactored this to treat upserts as peers to creates etc. I'm much happier with this version.

packages/server/src/api/rest/index.ts
```diff
@@ -809,6 +829,89 @@ class RequestHandler extends APIHandlerBase {
...
+                }
+
+                if (relationInfo.isCollection) {
+                    upsertPayload.create[key] = {
+                        connect: enumerate(data.data).map((item: any) =>
+                            this.makeIdConnect(relationInfo.idFields, item.id)
+                        ),
+                    };
+                    upsertPayload.update[key] = {
+                        connect: enumerate(data.data).map((item: any) =>
```
ymc9, line 886: I'm wondering if the semantic for the "update" case should use "set" instead of "connect" here - replace the entire connection collection instead of adding new ones?

packages/server/src/api/rest/index.ts
```diff
@@ -809,6 +829,89 @@ class RequestHandler extends APIHandlerBase {
...
+            return this.makeUnsupportedModelError(type);
+        }
+
+        const { error, attributes, relationships } = this.processRequestBody(type, requestBody, zodSchemas, 'create');
+
+        if (error) {
+            return error;
+        }
+
+        const matchFields = (requestBody as any).meta.matchFields;
```
ymc9, line 851: `matchFields` should always be `string[]`, right? Shall we zod validate it?

packages/server/tests/api/rest.test.ts
```diff
@@ -6,6 +6,7 @@ import { createPostgresDb, dropPostgresDb, loadSchema, run } from '@zenstackhq/t
 import { Decimal } from 'decimal.js';
 import SuperJSON from 'superjson';
 import makeHandler from '../../src/api/rest';
+import { query } from 'express';
```
ymc9, line 9: is this added by accident?

ymc9 (changes requested):
Hi @thomassnielsen , overall it looks very good to me. Please check the few comments I left. Thanks!

packages/server/src/api/rest/index.ts
thomassnielsen: Yes an optional array of strings. It makes sense to zod validate it. Where would this be added?

packages/server/tests/api/rest.test.ts
thomassnielsen: Yes, removed it now

packages/server/src/api/rest/index.ts
ymc9: I see you currently have a check `body?.meta?.operation === 'upsert'`. Maybe we can define a zod schema and replace that check with zod parse?

thomassnielsen: I agree. Should the zod schema be a part of this file or be a part of the zodSchemas coming in to `handleRequest`? If the latter, where are these defined?

ymc9: Since it's only used locally I think we can just make it a member of the class. There's already one named `createUpdatePayloadSchema` defined 😄

thomassnielsen:
Looking into test failure

thomassnielsen:
> Looking into test failure

Fixed. This is now ready for re-review.

ymc9 (approved):
Thank you @thomassnielsen , looks great to me now!

ymc9:
Hi @thomassnielsen , I'm merging this PR and the other one you made. Will publish a 2.9.4 release shortly. Thanks for your continuous contribution!

Issue #1878, opened 2024-11-20
Bug - mutations with optimisticUpdate are failing since 2.9.1 


roomman:
@ymc9 I've updated to 2.9.1 which includes [your PR #1842](https://github.com/zenstackhq/zenstack/pull/1842), and I am seeing failing mutations, and no network requests being made, when `optimisticUpdate: true`. 

The cases that are failing include optional "belongs-to" relationships where no relationship has yet been formed. For example:

```
model SomeModel {
  id                           String   @id @default(uuid())
  title                        String
  business               Business?          @relation(fields: [businessId], references: [id], onDelete: Cascade)
  businessId            String?

}
```
I am seeing this error:
`Error - there was an error updating SomeModel TypeError: Cannot convert undefined or null to object`

The error is about to the "business" relationship from the above schema, and an evaluation on this line from mutator.ts: https://github.com/zenstackhq/zenstack/blob/a003be9ee46daf791550993cf2b595f4ed9f9214/packages/runtime/src/cross/mutator.ts#L165

Because no business has been related to SomeModel the value of both `business` and `businessId` is `null`. Your change expects either an array or an object. 

If I set `optimisticUpdates: false` for the mutation the error goes away and mutations are successful.

Should L165 be doing a null check?

roomman:
@ymc9 I've created a [minimal reproduction](https://github.com/roomman/zenstack-minimal-repro-uptimistic-update-bug).

**Set up**:
- Clone the repo
- Run `npm install`
- Run `npm run generate`
- Run `npm run seed`
- Run `npm run dev`

Open your browser and you will see that you can click on an "Enter Competition" button for any Competition, including those with no related Business and it works.

Next run `npm run up` to install the latest and greatest. Run `npm run dev` again and retry entering a competition. Check the dev tools and see failed mutations. Under state your should see the following error: `{"name":"TypeError","message":"Cannot convert undefined or null to object"}`

Got to src/app/page.tsx L37 and comment out the optimistic update:

```tsx
  // create a new competition entry
  const { mutate: enterCompetition } = useCreateCompetitionEntry({
    // optimisticUpdate: true,
  });
```

Refresh and retry entering a competition and note that the mutations are now working as expected.

Revert the change to L37. Instead, comment out L14-19 (i.e. don't include the optional relationship):

```tsx
  // fetch all competitions
  const { data: allCompetitions, isPending } = useFindManyCompetition({
    include: {
      // business: {
      //   select: {
[...]

PR #1884, opened 2024-11-25
fix: optimistic updates for nested queries should work then optional relationships have null values

roomman:
This PR addresses the issue raised in #1878. It adds guards to `doApplyMutation` to ensure that `resultData` is an array or an object before it is processed.

I had a bit of trouble getting nock to play nicely in the test suite and needed to be more explicit about which endpoints were being intercepted. I added a constant for `BASE_URL` and added that into `makeUrl` too, for consistency. I hope that's okay?

Either way, the tests are passing and I hope this is a useful contribution to your epic project! 🚀


ymc9 (approved):
LGTM! Thank you for the carefully crafted test cases!

PR #1921, opened 2024-12-20
(WIP) feat: add `list` permission

### LINKED ISSUE #982: [Feature Request] New "list" permission kind
ZenStack currently supports four permission kinds: `create`, `read`, `update`, and `delete`. There are use cases where we want to allow users to read specific entities but want to deny fetching a list of them. This is quite common in products like S3. A separate `list` permission will enable this scenario.

genu:
Resolves #982 

This PR is an initial attempt to add a `list` permission. Putting here for visibility.

@ymc9 If you have some time, could you offer some guidance on the relevant code that would need to be updated, or anything else that would be helpful here.

### DIFF
--- a/packages/runtime/src/enhancements/node/policy/policy-utils.ts
+++ b/packages/runtime/src/enhancements/node/policy/policy-utils.ts
@@ -277,6 +277,7 @@ export class PolicyUtil extends QueryUtils {
             create: { guard: true, inputChecker: true },
             update: { guard: true },
             delete: { guard: true },
+            list: { guard: true },
             postUpdate: { guard: true },
         },
     };
--- a/packages/runtime/src/enhancements/node/types.ts
+++ b/packages/runtime/src/enhancements/node/types.ts
@@ -140,6 +140,7 @@ export type ModelCrudDef = {
     create: ModelCreateDef;
     update: ModelUpdateDef;
     delete: ModelDeleteDef;
+    list: ModelListDef;
     postUpdate: ModelPostUpdateDef;
 };
 
@@ -206,6 +207,11 @@ type ModelUpdateDef = ModelCrudCommon;
  * Policy definition for deleting a model
  */
 type ModelDeleteDef = ModelCrudCommon;
+
+/**
+ * Policy definition for listing a model
+ */
+type ModelListDef = ModelCrudCommon;
 
 /**
  * Policy definition for post-update checking a model
--- a/packages/runtime/src/types.ts
+++ b/packages/runtime/src/types.ts
@@ -35,7 +35,7 @@ export interface DbOperations {
  */
 export type PolicyKind = 'allow' | 'deny';
 
-export type PolicyCrudKind = 'read' | 'create' | 'update' | 'delete';
+export type PolicyCrudKind = 'read' | 'create' | 'update' | 'delete' | 'list';
 
 /**
  * Kinds of operations controlled by access policies
--- a/packages/schema/src/plugins/zod/generator.ts
+++ b/packages/schema/src/plugins/zod/generator.ts
@@ -326,6 +326,7 @@ export class ZodSchemaGenerator {
 
     private addPreludeAndImports(decl: DataModel | TypeDef, writer: CodeBlockWriter, output: string) {
         writer.writeLine('/* eslint-disable */');
+        writer.writeLine('// @ts-nocheck');
         writer.writeLine(`import { z } from 'zod';`);
 
         // import user-defined enums from Prisma as they might be referenced in the expressions
--- a/packages/schema/src/res/stdlib.zmodel
+++ b/packages/schema/src/res/stdlib.zmodel
@@ -504,7 +504,7 @@ attribute @@schema(_ name: String) @@@prisma
  * @param operation: comma-separated list of "create", "read", "update", "delete". Use "all" to denote all operations.
  * @param condition: a boolean expression that controls if the operation should be allowed.
  */
-attribute @@allow(_ operation: String @@@completionHint(["'create'", "'read'", "'update'", "'delete'", "'all'"]), _ condition: Boolean)
+attribute @@allow(_ operation: String @@@completionHint(["'create'", "'read'", "'update'", "'delete'", "'list'", "'all'"]), _ condition: Boolean)
 
 /**
  * Defines an access policy that allows the annotated field to be read or updated.
@@ -522,7 +522,7 @@ attribute @allow(_ operation: String @@@completionHint(["'create'", "'read'", "'
  * @param operation: comma-separated list of "create", "read", "update", "delete". Use "all" to denote all operations.
  * @param condition: a boolean expression that controls if the operation should be denied.
  */
-attribute @@deny(_ operation: String @@@completionHint(["'create'", "'read'", "'update'", "'delete'", "'all'"]), _ condition: Boolean)
+attribute @@deny(_ operation: String @@@completionHint(["'create'", "'read'", "'update'", "'delete'", "'list'", "'all'"]), _ condition: Boolean)
 
 /**
  * Defines an access policy that denies the annotated field to be read or updated.
--- a/packages/sdk/src/policy.ts
+++ b/packages/sdk/src/policy.ts
@@ -12,6 +12,7 @@ export function analyzePolicies(dataModel: DataModel) {
     const read = toStaticPolicy('read', allows, denies);
     const update = toStaticPolicy('update', allows, denies);
     const del = toStaticPolicy('delete', allows, denies);
+    const list = toStaticPolicy('list', allows, denies);
     const hasFieldValidation = hasValidationAttributes(dataModel);
 
     return {
@@ -21,6 +22,7 @@ export function analyzePolicies(dataModel: DataModel) {
         read,
         update,
         delete: del,
+        list,
         allowAll: create === true && read === true && update === true && del === true,
         denyAll: create === false && read === false && update === false && del === false,
         hasFieldValidation,


ymc9:
Thank you for starting this @genu ! It's something wanted by many.

I think the basic implementation should be straightforward. The call path for `findXXX` proxy is:

> PolicyProxyHandler.findXXX -> PolicyProxyHandler.doFind -> PolicyUtils.injectForRead

`PolicyUtils.injectForRead` is the unified function for injecting "read" related policies. Here' you can see currently only "read" policies are injected:

https://github.com/zenstackhq/zenstack/blob/b41fd93977416de1e9d2ab39f941440b38ddc302/packages/runtime/src/enhancements/node/policy/policy-utils.ts#L609-L612

I think what needs to be done is to further inject "list" policies.

Also, the "list" policy should only govern "findFirst" and "findMany", not "findUnique". We should probably introduce a boolean flag to distinguish these two cases.

I guess the proxy code is not entirely easy to read and understand 😅. Let me know if you need more clarifications.

ymc9:
Another complication is, whether "list" should govern `groupBy` call, since `groupBy` is partly "listing" data as well. What do you think?

genu:
> Another complication is, whether "list" should govern `groupBy` call, since `groupBy` is partly "listing" data as well. What do you think?

I'm not exactly sure, but I think I would expect `list` to govern anything that "lists" things.

genu:
@ymc9 Do we need to distinguish between between `findFirst` and `findMany` here?

https://github.com/zenstackhq/zenstack/blob/b41fd93977416de1e9d2ab39f941440b38ddc302/packages/runtime/src/enhancements/node/policy/handler.ts#L133-L151

or should `injectForRead` be modified here to accept a `findMany` parameter to indicate that the `list` permission should apply?

Also, why would `list` apply to a `findFirst`?

ymc9:
> @ymc9 Do we need to distinguish between between `findFirst` and `findMany` here?
> 
> https://github.com/zenstackhq/zenstack/blob/b41fd93977416de1e9d2ab39f941440b38ddc302/packages/runtime/src/enhancements/node/policy/handler.ts#L133-L151
> 
> or should `injectForRead` be modified here to accept a `findMany` parameter to indicate that the `list` permission should apply?
> 
> Also, why would `list` apply to a `findFirst`?

I think 'list' should govern `findFirst` because otherwise you'll be able to iteratively call `findFirst` to simulate `findMany`:

```ts
let r = await db.foo.findFirst();
const knownIds: string[] = [];
while (r) {
  knownIds.push(r.id);
  r = await db.foo.findFirst({ where: { not: { id: { in: knownIds } } } });
}
```

We can probably introduce a boolean parameter to `doFind` to distinguish the two cases where "list" should be enforced or not.

packages/runtime/src/enhancements/node/policy/handler.ts
```diff
@@ -140,6 +140,14 @@ export class PolicyProxyHandler<DbClient extends DbClientContract> implements Pr
...
         }
 
+        if (isList && !this.policyUtils.injectForList(this.prisma, this.model, _args)) {
+            if (this.shouldLogQuery) {
+                this.logger.info(`[policy] \`${actionName}\` ${this.model}: unconditionally denied`);
+            }
+
+            return handleRejection();
+        }
+
```
genu, line 150: I'm assuming this is the kind of check we need to do here?

packages/runtime/src/enhancements/node/policy/policy-utils.ts
```diff
@@ -646,6 +652,14 @@ export class PolicyUtil extends QueryUtils {
         return true;
     }
 
+    /**
+     * Injects auth guard for read operations.
+     */
+    injectForList(_db: CrudContract, _model: string, _args: any) {
+        // make select and include visible to the injection
+        return true;
+    }
```
genu, line 661: @ymc9 
I'm not exactly sure if this is needed. Should the `list` logic go inside of the `injectForRead`?

The `isList` flag is on the handler `doFind` I'm a little confused on how the flow is.

ymc9: Yes, I agree this list logic should go inside `injectForRead`. Maybe we just rename it to `injectFoReadOrList` and pass in the flag there. I think the only difference is the call to `this.injectAuthGuardAsWhere`, for read the arg is `"read"`, and `"list"` for list.

ymc9: https://github.com/zenstackhq/zenstack/blob/4b7d813624b4e73fcf8fd14d2849d5b1aef6aae3/packages/runtime/src/enhancements/node/policy/policy-utils.ts#L609

ymc9: That part of code is not entirely easy to comprehend 😄. Please feel free to pass the PR to me to finish when you reach a point where you feel the basics are working.

genu:
@ymc9 Here are some ideas from thinking through this:

**findMany**
- Controlled by list permission since you're viewing multiple resources
- Would attribute permissions be required to determine what fields are visible in a `list` context? (In the case that `list` is allowed and `read` is denied). Unless `list` implies `read` but that may not necessarily the case.

**findFirst**
- Still governed by list permission even though returning single item
- Because it involves scanning/viewing multiple items to find the match

**Nested to-many relations during read**
- List permission required for the related collection (should a permission allow list, if parent allows a `read`? Similar to how `check` works currently)

**Aggregate** (This is kind of tricky)
- List permission required as you're viewing/organizing multiple items
- May need attribute permissions for the grouping fields, plus permissions for the fields being aggregated within groups

**Count**
- Requires list permission as it reveals information about collection size
- Maybe we have to treat this as a separated permission as count info is sensitive?



thoughts?

ymc9:
Hi @genu , sorry I almost missed this comment. I think my thoughts generally aligns with your description. I'm sharing the differences below.

> @ymc9 Here are some ideas from thinking through this:
> 
> **findMany**
> 
> * Controlled by list permission since you're viewing multiple resources
> * Would attribute permissions be required to determine what fields are visible in a `list` context? (In the case that `list` is allowed and `read` is denied). Unless `list` implies `read` but that may not necessarily the case.

I think field-level and model-level permissions should work independently (as they do today). Model-level "list" determines if you can list entities, and field-level rules determine what fields you can see. You can't define "list" rules at the field level because it's meaningless.

> 
> **findFirst**
> 
> * Still governed by list permission even though returning single item
> * Because it involves scanning/viewing multiple items to find the match
> 
> **Nested to-many relations during read**
> 
> * List permission required for the related collection (should a permission allow list, if parent allows a `read`? Similar to how `check` works currently)

The permission checking of fetching relations doesn't inherit from its parent, instead, for "read", the relation's model's "read" rules are evaluated to filter out items that shouldn't be seen, regardless if the parent is readable. I think "list" can behave the same way.

> 
> **Aggregate** (This is kind of tricky)
> 
> * List permission required as you're viewing/organizing multiple items
> * May need attribute permissions for the grouping fields, plus permissions for the fields being aggregated within groups
> 
> **Count**
> 
> * Requires list permission as it reveals information about collection size
> * Maybe we have to treat this as a separated permission as count info is sensitive?

I agree "count" should require "list" permission. I don't see a separate permission kind is needed for it for now. We can iterate in the future as we gather feedback.

> 
> Maybe for an initial implementation we should stick to the obvious/straight forward approach and polish it later

I totally agree. There're many details to sort out. Just having a single `findMany` to work will be a great start!

> 
> thoughts?

Another major problem is how we add the "list" permission without introducing breaking changes, plus allowing incremental adoption. Here are my thoughts:

- If there're no "list" rules defined for a model, "list" is allowed if "read" is. This will allow people to continue using `findMany` etc. with only "read" rules defined.
- If "list" rules are defined for a model, it then implies "read", and operations like `findMany`, `aggregate`, etc. will require "list" permissions to work. Basically, by starting using "list" permission you opt in the new behavior for that model.

What do you think?

genu:
@ymc9 Do you mind taking over this PR, I'm not exactly sure how to proceed.

I'll keep an eye out for an updates and review/test as needed.

hongkongkiwi:
@ymc9 your comments above make sense on how you expect it to work. I do have a suggestion though.

> If there're no "list" rules defined for a model, "list" is allowed if "read" is. This will allow people to continue using findMany etc. with only "read" rules defined.
> If "list" rules are defined for a model, it then implies "read", and operations like findMany, aggregate, etc. will require "list" permissions to work. Basically, by starting using "list" permission you opt in the new behavior for that model.
> 

What I'm gathering from this is that this will just transparently work and start filtering out entries if it's not specifically defined right but the user has read access? If I'm correct in this understanding then I recommend below.

> The permission checking of fetching relations doesn't inherit from its parent, instead, for "read", the relation's model's "read" rules are evaluated to filter out items that shouldn't be seen, regardless if the parent is readable. I think "list" can behave the same way.

Since this is a breaking change for some users, I would recommend a flag to enable the filtering behaviour and have it off by default. This would allow people to do updates without items unexpectedly disappearing from their lists. E.g. it could be a feature flag in the zmodel as behaviour on or off globally (I don't mean for each model).

I'm really happy with that behaviour so I would switch it on, but since it would affect what shows up when users do a list, it's worth having it opt in imho.

ymc9:
> @ymc9 Do you mind taking over this PR, I'm not exactly sure how to proceed.
> 
> I'll keep an eye out for an updates and review/test as needed.

Sure @genu , and thank you for initiating this effort! I'm currently in the middle of experimenting some larger refactor, so will need to put this PR on hold for a bit. I'll keep you updated when I resume the work.

ymc9:
> @ymc9 your comments above make sense on how you expect it to work. I do have a suggestion though.
> 
> > If there're no "list" rules defined for a model, "list" is allowed if "read" is. This will allow people to continue using findMany etc. with only "read" rules defined.
> > If "list" rules are defined for a model, it then implies "read", and operations like findMany, aggregate, etc. will require "list" permissions to work. Basically, by starting using "list" permission you opt in the new behavior for that model.
> 
> What I'm gathering from this is that this will just transparently work and start filtering out entries if it's not specifically defined right but the user has read access? If I'm correct in this understanding then I recommend below.
> 
> > The permission checking of fetching relations doesn't inherit from its parent, instead, for "read", the relation's model's "read" rules are evaluated to filter out items that shouldn't be seen, regardless if the parent is readable. I think "list" can behave the same way.
> 
> Since this is a breaking change for some users, I would recommend a flag to enable the filtering behaviour and have it off by default. This would allow people to do updates without items unexpectedly disappearing from their lists. E.g. it could be a feature flag in the zmodel as behaviour on or off globally (I don't mean for each model).

Hi @hongkongkiwi , sorry I didn't explain clearly in the previous reply. The intention is to avoid any breaking changes. If you don't write any "list" rules, the "read" rules will behave the same as they do today. The behavior will only change if you start to have "list" rules in a model.

I like the idea of using a runtime flag to control the behavior as well. It's probably cleaner. I'll keep thinking about it. Thanks!

hongkongkiwi:
What happened to this list permission? Is work stopped?

genu:
> What happened to this list permission? Is work stopped?

Correct me if I'm wrong @ymc9 but I believe this feature is being pushed to `v3` which is in the works here: https://github.com/zenstackhq/zenstack-v3

ymc9:
> > What happened to this list permission? Is work stopped?
> 
> 
> 
> Correct me if I'm wrong @ymc9 but I believe this feature is being pushed to `v3` which is in the works here: https://github.com/zenstackhq/zenstack-v3

Yes, that's what I'm thinking. It'll need to be rewritten if implemented in v2.

ymc9:
Closing for now.

PR #1922, opened 2024-12-20
feat: Add `@encrypted` enhancer

### LINKED ISSUE #1643: [Feature Request] Add a `@encrypted` enhancer
**Is your feature request related to a problem? Please describe.**
Relevant discord discussion: https://discord.com/channels/1035538056146595961/1267557057784053924

I'm trying to eliminate the need for this extension: https://github.com/47ng/prisma-field-encryption

It allows marking some fields as encrypted thus the field is never stored in plain text in the db, but only can be read in plain text at runtime through the prisma client.

**Describe the solution you'd like**
A simple `@encrypted` enhancer would suffice. It would encrypted the field when writing, and unencrypted as plain text when reading.

**Describe alternatives you've considered**
Currently using https://github.com/47ng/prisma-field-encryption but it doesn't work well with zenstack generator because it relies on specific annotations which zenstack moves around during `generate`

genu:
Resolves #1643


packages/runtime/src/types.ts
```diff
@@ -166,3 +172,10 @@ export type ZodSchemas = {
      */
     input?: Record<string, Record<string, z.ZodSchema>>;
 };
+
+export type CustomEncryption = {
+    encrypt: (model: string, field: FieldInfo, plain: string) => Promise<string>;
+    decrypt: (model: string, field: FieldInfo, cipher: string) => Promise<string>;
+};
+
+export type SimpleEncryption = { encryptionKey: string };
```
ymc9, line 181: Since we don't need to represent the key in plain text anywhere, is it better to define the key as a UInt8Array instead? So that we don't need to imply any encoding form or limit it to be a utf-8 string.

packages/runtime/src/enhancements/node/encrypted.ts
```diff
@@ -0,0 +1,155 @@
...
+            const shouldDecrypt = fieldInfo.attributes?.find((attr) => attr.name === '@encrypted');
+            if (shouldDecrypt) {
+                entityData[field] = await this.decrypt(fieldInfo, entityData[field]);
+            }
+        }
+    }
+
+    private async preprocessWritePayload(model: string, action: PrismaWriteActionType, args: any) {
+        const visitor = new NestedWriteVisitor(this.options.modelMeta, {
+            field: async (field, _action, data, context) => {
```
ymc9, line 145: Shall we add a nullish check to `data` here?

packages/runtime/src/types.ts
ymc9: If we do so, we can also validate if the given key is 32-byte long.

packages/runtime/src/enhancements/node/create-enhancement.ts
```diff
@@ -120,13 +122,22 @@ export function createEnhancement<DbClient extends object>(
...
-    // password enhancement must be applied prior to policy because it changes then length of the field
+    // password and encrypted enhancement must be applied prior to policy because it changes then length of the field
     // and can break validation rules like `@length`
     if (hasPassword && kinds.includes('password')) {
         // @password proxy
         result = withPassword(result, options);
     }
 
+    if (hasEncrypted && kinds.includes('encrypted')) {
+        if (!options.encryption) {
```
ymc9, line 133: Shall we validate the shape of `options.encryption`? Here or inside the `EncryptedHandler` constructor.

packages/runtime/src/enhancements/node/encrypted.ts
```diff
@@ -0,0 +1,155 @@
...
+        for (const field of getModelFields(entityData)) {
+            const fieldInfo = await resolveField(this.options.modelMeta, realModel, field);
+
+            if (!fieldInfo) {
+                continue;
+            }
+
+            const shouldDecrypt = fieldInfo.attributes?.find((attr) => attr.name === '@encrypted');
+            if (shouldDecrypt) {
+                entityData[field] = await this.decrypt(fieldInfo, entityData[field]);
```
ymc9, line 138: If decryption fails, should we return the original cipher text? I'm thinking this will allow easier adoption: the `@encrypted` attribute can be added and deployed and then a background script is run to migrate the existing plain-text data.

ymc9 (changes requested):
Hi @genu , thanks a lot for contributing this feature! And I appreciate that you incorporated our latest discussions into it 😄.

Overall the changes look great to me. I've left a few comments. Please take a look. Thanks!

packages/runtime/src/enhancements/edge/encrypted.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    private async decrypt(field: FieldInfo, data: string): Promise<string> {
+        if (this.isCustomEncryption(this.options.encryption!)) {
+            return this.options.encryption.decrypt(this.model, field, data);
+        }
+
+        const key = await getKey(this.options.encryption!.encryptionKey);
+
+        // Convert base64 back to bytes
+        const bytes = Uint8Array.from(atob(data), (c) => c.charCodeAt(0));
```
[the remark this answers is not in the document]
genu: I'm not sure why, but doing `const bytes = encoder.encode(atob(data))` doesn't have any type errors, but throws an error on `build`

packages/runtime/src/enhancements/node/create-enhancement.ts
genu: I feel like validating inside of the handler constructor makes better sense, and its together with the rest of the `encrypted` logic

packages/runtime/src/enhancements/node/encrypted.ts
genu: That may be vulnerable to data leaks some edge cases I think. For example, if custom encryption is used and the encrypt function fails resulting in storing the data in plain text, then it would leak during read operations.

I'm thinking that an encrypted field should be treated as all or nothing.

what do you think?

packages/runtime/src/types.ts
genu: I think we still need to keep it a `string` since the user would actually provide it as a `string`.

Maybe keep it as string, but ensure it is a valid `UInt8Array`?

@ymc9  what do you think?

packages/runtime/src/enhancements/edge/encrypted.ts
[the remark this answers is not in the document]
ymc9: Oh, sorry, my bad, I was thinking `atob` returns a Buffer ...

Seems like your code is the right way to do it. Using a Buffer is probably more efficient, but I'm not sure about the compatibility with edge runtimes. Let's keep it this way for now.

packages/runtime/src/enhancements/node/create-enhancement.ts
```diff
@@ -14,13 +14,14 @@ import { withJsonProcessor } from './json-processor';
...
 import { withPassword } from './password';
+import { withEncrypted } from './encrypted';
 import { policyProcessIncludeRelationPayload, withPolicy } from './policy';
 import type { PolicyDef } from './types';
 
 /**
  * All enhancement kinds
  */
-const ALL_ENHANCEMENTS: EnhancementKind[] = ['password', 'omit', 'policy', 'validation', 'delegate'];
+const ALL_ENHANCEMENTS: EnhancementKind[] = ['password', 'omit', 'policy', 'validation', 'delegate', 'encrypted'];
```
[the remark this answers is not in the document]
ymc9: The `withEncrypted` wrapper is only added if any model field has the `@encrypted` attribute. I think it should be fine, what do you think?

packages/runtime/src/types.ts
ymc9: I'm think the user will use it like:

```ts
const getEncryptionKey() {
  return new Uint8Array(Buffer.from(process.env.ENCRYPTION_KEY, 'base64'));
}

const db = enhance(prisma, { user }, { encryption: { encryptionKey: getEncryptionKey() } });
```

He can also store the key as binary form somewhere and then load it directly and pass to the `enhance` call. My main thought is that zenstack can be agnostic about the key storage form.

packages/runtime/src/enhancements/node/encrypted.ts
```diff
@@ -0,0 +1,167 @@
...
+
+            const shouldDecrypt = fieldInfo.attributes?.find((attr) => attr.name === '@encrypted');
+            if (shouldDecrypt) {
+                // Don't decrypt null, undefined or empty string values
+                if (!entityData[field]) return;
+
+                entityData[field] = await this.decrypt(fieldInfo, entityData[field]);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
ymc9: I seems @coderabbitai's comment about "return" -> "continue" is correct 😄 @genu

packages/runtime/src/enhancements/node/encrypted.ts
genu: That may be vulnerable to data leaks for some edge cases I think. For example, if custom encryption is used and the encrypt function fails resulting in storing the data in plain text, then it would leak during read operations.

I'm thinking that an encrypted field should be treated as all or nothing. (e.g. throw an error if encrypt fails, or return a null of decrypt fails?)

what do you think?

packages/runtime/src/types.ts
genu: Would they be able to do simply:

`const db = enhance(prisma, { user }, { encryption: { encryptionKey: "c558Gq0YQK2QcqtkMF9BGXHCQn4dMF8w" } });`

With the field-encryption extension, it allows passing in a key as a string (I believe it validates its length at runtime)

packages/runtime/src/enhancements/node/create-enhancement.ts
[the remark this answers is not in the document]
genu: you're right :)

packages/runtime/src/types.ts
ymc9: Yes, I see that field-encryption uses https://github.com/47ng/cloak to manage encryption keys - which seems to use a special encoding form (base64 + some annotations).

We can improve the options to allow base64 strings as well in the future.

packages/runtime/src/enhancements/node/encrypted.ts
ymc9: I agree encryption should be strict and the write operation should just fail as a whole if encrypting a field fails.

However, I'm worried treating decryption "strict" will make incremental adoption very difficult. If one already has some rows and then turn on `@encrypted`, for a period of time, the column will be mixed with plain and cipher text, unless the system is brought down to maintenance to allow a migration process to complete.

As long as the encryption process's integrity is guaranteed, if decryption fails (e.g., due to incorrect key etc.), the worst case scenario will be cipher text is returned the the caller, right?

packages/runtime/src/enhancements/node/encrypted.ts
```diff
@@ -0,0 +1,175 @@
...
+            key,
+            this.encoder.encode(data)
+        );
+
+        // Combine IV and encrypted data into a single array of bytes
+        const bytes = [...iv, ...new Uint8Array(encrypted)];
+
+        // Convert bytes to base64 string
+        return btoa(String.fromCharCode(...bytes));
+    }
```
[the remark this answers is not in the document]
genu: Is this necessarily since we're already wrapping the outer call in a try/catch?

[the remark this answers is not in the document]
ymc9: Oh, sorry I missed the try/catch part. It looks good then!

ymc9 (approved):
Thank you for working on this great feature and the patience for the review process @genu 😄. I'm approving and merging it. It'll be part of the v2.11.0 release!

PR #1989, opened 2025-02-19
fix: don't set default value in nested writes when set through FK

Gabrola:
```ts
const result = db.post.create({
    data: {
        likes: {
            createMany: {
                data: [{
                    userId: user.id
                }]
            }
        }
    },
    include: {
        likes: true
    }
});

// results in ↓

const result = await db[model].create({
    data: {
        likes: {
            createMany: {
                data: [
                    {
                        userId: "48a96b75-882e-48bc-92eb-2ec266067b6d",
                        tenantId: "fa9c2cd6-ae4f-457a-b556-9089f25d67d4"
                    }
                ]
            }
        },
        tenant: {
            connect: {
                id: "fa9c2cd6-ae4f-457a-b556-9089f25d67d4"
            }
        },
        author: {
            connect: {
                id: "48a96b75-882e-48bc-92eb-2ec266067b6d"
            }
        }
    },
    select: {
        tenantId: true,
        id: true,
        likes: {
            select: {
                tenantId: true,
                id: true
            }
        }
    }
})

// Prisma error: Unknown argument `tenantId`. Available options are marked with ?.]
```

### DIFF
--- a/packages/runtime/src/enhancements/node/default-auth.ts
+++ b/packages/runtime/src/enhancements/node/default-auth.ts
@@ -117,55 +117,75 @@ class DefaultAuthHandler extends DefaultPrismaProxyHandler {
         authDefaultValue: unknown,
         context: NestedWriteVisitorContext
     ) {
-        if (fieldInfo.isForeignKey && fieldInfo.relationField && fieldInfo.relationField in data) {
+        if (fieldInfo.isForeignKey) {
+            // if the field being inspected is a fk field, there are several cases we should not
+            // set the default value or should not set directly
+
             // if the field is a fk, and the relation field is already set, we should not override it
-            return;
-        }
+            if (fieldInfo.relationField && fieldInfo.relationField in data) {
+                return;
+            }
 
-        if (context.field?.backLink) {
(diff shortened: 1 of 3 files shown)


ymc9:
Hi @Gabrola , I checked the change and the location you changed is spot on. However, I think it's safer if we detect such condition by inspecting the nested creation path (and not just checking back link and foreign key mappings), because a model being created (in a nested context) may have multiple relations, and the parent context may not be the one that's related to the foreign key field with default value.

```prisma
model A {
  c C?
}

model B {
  c C?
}

model C {
  a A
  aId String @default(auth().id)
  b B
  bId String @default(auth().id)
}

// "bId" should still be set
db.a.create({ data: { c: {} } });
```

I'm making some changes directly based on your PR. Do you mind taking a look later?

ymc9:
fixes #1997

Gabrola:
@ymc9 this is perfect! I wasn't super confident in my approach, so thank you for making it more robust!

ymc9:
Thanks @Gabrola . Merging it now and will include the fix in the next patch release.

ymc9 (approved):

PR #2018, opened 2025-02-27
feat(runtime): inject enhanced client or tx context so it can be retrieved in extensions

Gabrola:
Prisma already exposes a `$parent` property in the context, but this will point towards the unenhanced client or transaction object

### DIFF
--- a/packages/runtime/src/enhancements/node/proxy.ts
+++ b/packages/runtime/src/enhancements/node/proxy.ts
@@ -289,7 +289,7 @@ export function makeProxy<T extends PrismaProxyHandler>(
                 return propVal;
             }
 
-            return createHandlerProxy(makeHandler(target, prop), propVal, prop, errorTransformer);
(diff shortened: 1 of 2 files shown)


ymc9 (commented):
I didn't know there's actually the `$parent` field in client extensions that refer to the current prisma client instance. I saw it in Prisma code but seems not documented anywhere ...

Do you think we should simply proxy this field directly instead of using a separate `$zenstack_parent` field? Wondering if most people would expect `$parent` to be enhanced if it originated from an enhanced client. What do you think?

Gabrola:
@ymc9 Yeah, actually that makes a lot of sense since the proxy will still allow access to any non-proxied properties; so this shouldn't break any existing usage of `$parent`. I'll push that change 👍

ymc9 (approved):
Thank you! Looking great now!

PR #2022, opened 2025-02-28
Feature request: plugin “preprocessor” option

chunkerchunker:
Would you be open to this feature (or something similar)?  I have a use case where I want to make a large number of modifications to the schema based on some custom attributes, and I want these changes to be made prior to Zenstack's plugins running, so that I still get all of the Zenstack features on top of my modifications.  This pull request is a tiny change that recognizes the "preprocessor" flag in a plugin's config, which causes that plugin to run before Zenstack's plugins.

Happy to make any changes, but wanted to run it by you to see if there are better options.

Thanks!

### DIFF
--- a/packages/schema/src/cli/plugin-runner.ts
+++ b/packages/schema/src/cli/plugin-runner.ts
@@ -108,8 +108,14 @@ export class PluginRunner {
             });
         }
 
+        const preprocessorPlugins = plugins.filter((p) => p.options.preprocessor);
+        const otherPlugins = plugins.filter((p) => !p.options.preprocessor);
+
         // calculate all plugins (including core plugins implicitly enabled)
-        const { corePlugins, userPlugins } = this.calculateAllPlugins(runnerOptions, plugins);
+        const { corePlugins, userPlugins } = this.calculateAllPlugins(
+            runnerOptions,
+            otherPlugins,
+        );
         const allPlugins = [...corePlugins, ...userPlugins];
 
         // check dependencies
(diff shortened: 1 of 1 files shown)


ymc9:
> Would you be open to this feature (or something similar)? I have a use case where I want to make a large number of modifications to the schema based on some custom attributes, and I want these changes to be made prior to Zenstack's plugins running, so that I still get all of the Zenstack features on top of my modifications. This pull request is a tiny change that recognizes the "preprocessor" flag in a plugin's config, which causes that plugin to run before Zenstack's plugins.
> 
> Happy to make any changes, but wanted to run it by you to see if there are better options.
> 
> Thanks!

Hey @chunkerchunker , thanks for making this PR! Do you mind sharing a bit more about what kind of modifications you plan to make with the preprocessors? Is it about minipulating the ZModel AST before it's ingested by the core plugins? The AST nodes have internal linkages (mainly named references to their resolved declarations), so handling them can be tricky or may result in downstream errors.

chunkerchunker:
Hi @ymc9, yes that's exactly the usage.  Specifically, I have a crdt sync engine that I've built for my app (that has some unique properties not satisfied by existing engines I looked at, like vlcn.io and electric-sql).  My plugin looks for models that have been annotated for supporting sync, and it adds corresponding HLC_<field> and updatedAt_<field> for each sync'd field, along with a some other boilerplate fields at the model level.  Having a plugin do this lets me automate everything and make it maintainable.

I did notice when implementing this your point about internal linkages in the AST nodes.  So, the way I got things to work in my plugin was to create a single template model in the original schema, and then I copied and modified fields from the template wherever needed (updating the container linkages on the copies).

I can understand that you wouldn't want to expose much of your implementation details as API, but do you think there's a small enough surface area of the AST that can be relied on to make this work?  Or another option is just to say that any usage is at the user's risk, in a similar way to how Prisma treats the dmmf?

Thanks for the consideration!

ymc9:
Thanks for the explanation @chunkerchunker .

The ZModel AST is a grey area of ZenStack's "API", since the surface is really wide. It's fairly stable now and usually things are added instead of changed/deleted, however I feel it's difficult to let it follow semver. Yes, it's more like Prisma's DMMF. For your use case, if it's mostly about introducing simple fields I guess the risk is low. Thing involving altering relations or inheritance hierarchy can be more subtle.

I think it's fine to introduce the `preprocessor` flag and it adds the flexibility to alter ZModel on the fly as you wish. Just to need to make sure the altered AST is in a full resolved state.

chunkerchunker:
@ymc9 Did you want me to make any changes to this before considering it?  I also don't mind just waiting until 3.0 if there's a cleaner approach you have there.  Thanks.

ymc9:
> @ymc9 Did you want me to make any changes to this before considering it? I also don't mind just waiting until 3.0 if there's a cleaner approach you have there. Thanks.

The change looks cool to me. We can get it merged for v2. I can't promise it won't break for v3 though 😄.

I'm merging and will include it in the next patch release.

chunkerchunker:
Thank you!

PR #2029, opened 2025-03-06
fix(zod): delegate aux fields appearing in WhereUniqueInput schemas

### LINKED ISSUE #2028: delegate aux fields appearing in WhereUniqueInput schema, when aux field is not first in a @@unique constraint
Given the following schema:

```
enum FooType {
    Bar
    Baz
}

model User {
    id          String       @id @default(cuid())
    userFolders UserFolder[]
}

model Foo {
    id          String       @id @default(cuid())
    type        FooType

    userFolders UserFolder[]

    @@delegate(type)
}

model Bar extends Foo {
    name String
}

model Baz extends Foo {
    age Int
}

model UserFolder {
    id     String @id @default(cuid())
    userId String
    fooId  String
    path   String

    user   User   @relation(fields: [userId], references: [id])
    foo    Foo    @relation(fields: [fooId], references: [id])

    @@unique([userId, fooId])
}
```

An invalid Zod schema is generated for `UserFolderWhereUniqueInputObjectSchema`

This schema attempts to import CompoundUniqueInput schemas for `userId_delegate_aux_UserFolder_fooId_Bar/Baz`, but these are removed during schema generation, causing a runtime error when these schemas are used.

**Screenshots**

Image

-   ZenStack version: 12.2.1
-   Prisma version: 6.4.1
-   Database type: Postgresql

**Additional context**

The bug seems to be within the zod schema generator, where `.startsWith(DELEGATE_AUX_RELATION_PREFIX)` is still used instead of `.includes(DELEGATE_AUX_RELATION_PREFIX)` to filter delegate fields on an object. The `CompoundUniqueInput` schemas are filtered out correctly.

diesal11:
Fixes #2028

This also silently fixes the enhanced PrismaClient types by updating that code to use `.includes(DELEGATE_AUX_RELATION_PREFIX)` over `.startsWith(DELEGATE_AUX_RELATION_PREFIX)` to search for delegate aux fields.

Unfortunately I couldn't find an easy way to test the PrismaClient types within the existing test infrastructure, so that regression is untested.

### DIFF
--- a/tests/regression/tests/issue-2028.test.ts
+++ b/tests/regression/tests/issue-2028.test.ts
@@ -1,0 +1,116 @@
+import { createPostgresDb, loadSchema } from '@zenstackhq/testtools';
+
+describe('issue 2028', () => {
+    it('regression', async () => {
+        const dbUrl = await createPostgresDb('issue-2028');
+        const { enhance, zodSchemas } = await loadSchema(
+            `
+enum FooType {
+    Bar
+    Baz
+}
+
+model User {
+    id          String       @id @default(cuid())
+    userFolders UserFolder[]
+    @@allow('all', true)
+}
+
+model Foo {
+    id          String       @id @default(cuid())
+    type        FooType
+
+    userFolders UserFolder[]
+
+    @@delegate(type)
+    @@allow('all', true)
+}
+
+model Bar extends Foo {
+    name String
+}
+
+model Baz extends Foo {
+    age Int
+}
+
+model UserFolder {
+    id     String @id @default(cuid())
+    userId String
+    fooId  String
+
+    user   User   @relation(fields: [userId], references: [id])
+    foo    Foo    @relation(fields: [fooId], references: [id])
+
+    @@unique([userId, fooId])
+    @@allow('all', true)
+}
+            `,
+            {
+                fullZod: true,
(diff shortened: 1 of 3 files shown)


tests/regression/tests/issue-2028.test.ts
```diff
@@ -0,0 +1,116 @@
...
+        expect(
+            zodSchemas.objects.UserFolderWhereUniqueInputObjectSchema.safeParse({
+                userId_delegate_aux_UserFolder_fooId_Baz: {
+                    userId: '1',
+                    fooId: '2',
+                },
+            }).success
+        ).toBeFalsy();
+
+        // Ensure we can query by the CompoundUniqueInput
```
diesal11, line 74: This might be overkill as I don't think it was actually an issue originally?
But I wanted to ensure that no further issues were missed.

ymc9 (dismissed):
Hey @diesal11 , good catch and many thanks for making the fix! Code looks great to me. I'll merge and release a patch.

tests/regression/tests/issue-2028.test.ts
ymc9: Never hurt to have a little more coverage 😄

ymc9 (approved):

PR #2044, opened 2025-03-13
fix(zod): Required fields with a default value should be optional in Typed JSON cases

diesal11:
I believe this is just an edge case with Typed JSON fields again, but it did surface what appears to be a legitimate bug in the zod generator code. So I've tidied that section up slightly to fix the issue and improve readability a touch.

Apologies for not catching this with the last PR.


ymc9 (approved):
Thank you @diesal11 . The change looks good to me. It'll be part of the next release!

PR #2060, opened 2025-03-27
Use prisma edge client for edge target

fredevery:
While attempting to run a zenstack instance in a dockerized cloudflare enviroment,  errors were thrown as the prisma client attempted to require node apis that aren't available in the cloudflare enviroment.

I was able to resolve this by forcing the zenstack edge enhancer to use the prisma edge client.


ymc9 (approved):
Hi @fredevery , thank you for making the fix. The change looks good to me. I'll merge when CI passes.

PR #2064, opened 2025-04-01
Fix(JSON): Allow JSON types to be skipped through Prisma.skip

diesal11:
Another JSON edge case!

When the `strictUndefinedChecks` Prisma Preview feature is enabled, `Prisma.skip` is required to be used over `undefined`. The types generated for `@Json` fields misses this typing in cases where the field is optional in the Prisma type, for example within `UpdateInput` types

Note this doesn't allow using `Prisma.skip` on a field within the JSON type, as that doesn't appear to be supported by Prisma.


ymc9 (approved):
Good catch @diesal11 ! Thanks for making the fix. I'm merging it and will publish a new release.

Issue #2065, opened 2025-04-03
Zod Schema contains invalid import to logical Prisma Client, when @@validate rule references an enum


diesal11:
**Description and expected behavior**
Schema:
```
enum FooType {
    Bar
    Baz
}

// Add JSON type field so a logical prisma client is generated.
// Polymorphic models also trigger this bug 
type Meta {
    test String?
}

model Foo {
    id   String  @id @db.Uuid @default(uuid())
    type FooType
    meta Meta @json

    @@validate(type == Bar, "FooType must be Bar")
}
```

**Screenshots**
The generated Zod Schema file for `Foo` then contains an invalid import. `.zenstack/models` is a `.d.ts` file and contains no value exports.

![Image](https://github.com/user-attachments/assets/f5f2424f-5f7e-4a7e-b730-542aa24f3e42)

**Environment (please complete the following information):**

-   ZenStack version: 2.13.0
-   Prisma version: 6.5.0
-   Database type: Postgresql

**Additional context**

This bug doesn't occur when a logical client isn't generated. I believe the issue lies somewhere in this code, It doesn't really make sense to me that `prismaClientPath` is being mapped to a definitions file?

https://github.com/zenstackhq/zenstack/blob/8a62f631218addd36e09b4c2ca7f2fbdd146bed4/packages/schema/src/plugins/enhancer/index.ts#L29-L45

ymc9:
Hey @diesal11 , I had a look into it, you're right the problem is "./zenstack/models" currently only exports types. However, to import enums, values should be exported there as well.

I think a simple fix would be to generate a "models.js" file which reexports the original prisma client module. The "/models" will behave in a "strange" way that its types correspond to logical schema, however values come from the original schema. But that's how it should work ...

I'll make a PR for it.

PR #2069, opened 2025-04-06
Update Zod Generator - Fix Enum Imports

adnanalbeda:
Since zod generated `models` file are all just types. So `enum` exports from `models` are also types only.

Accordinlgy, enum imports must also be types only.

This fixed issues for generated models when working with `workspaces` of pnpm monorepo and `NextJS`.

### DIFF
--- a/packages/schema/src/plugins/zod/generator.ts
+++ b/packages/schema/src/plugins/zod/generator.ts
@@ -365,7 +365,7 @@ export const ${typeDef.name}Schema = ${refineFuncName}(${noRefineSchema});
         }
         if (importEnums.size > 0) {
             const prismaImport = computePrismaClientImport(path.join(output, 'models'), this.options);
-            writer.writeLine(`import { ${[...importEnums].join(', ')} } from '${prismaImport}';`);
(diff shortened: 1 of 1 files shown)


ymc9:
Hi @adnanalbeda , thank you for making this PR!

Coincidentally, a related issue was filed for the same issue as well #2065. I think the problem is that enums need to be reexported from ".zenstack/models" as values too (instead of just as types), otherwise at runtime, accessing enum values will still fail even if typescript compilation succeeds.

I've made a separate PR to let ".zenstack/models" reexport all values from the original PrismaClient model (while keeping "models.d.ts" reexport ZenStack-modified Prisma types). I think it should be more robust than fixing the enum imports.

https://github.com/zenstackhq/zenstack/pull/2070

adnanalbeda:
Great, I tested your changes and it's working better than mine.

ymc9:
> Great, I tested your changes and it's working better than mine.
> 
> 

Thanks for confirming it! I'll merge and make a patch release today.

adnanalbeda:
#2070 fixed the issue by supporting referencing `enum` as `values` instead of `types`.

PR #2078, opened 2025-04-10
fix(runtime): Refactor logical client exports to be ESM/CJS agnostic

diesal11:
When `zenstack generate` is run in a project with `"type": "module"` in it's package.json, startup fails with the following error:

```sh
 ⚠ ./lib/zenstack/models.js
Specified module format (EcmaScript Modules) is not matching the module format of the source code (CommonJs)
The EcmaScript module format was specified in the package.json that is affecting this source file or by using an special extension, but it looks like that CommonJs syntax is used in the source code.
Exports made by CommonJs syntax will lead to a runtime error, since the module is in EcmaScript mode. Either change the "type" field in the package.json or replace CommonJs syntax with EcmaScript import/export syntax in the source file.
```

Sidenote: Should this simply be inferred by reading package.json? I wasn't certain that would catch some cases in a monorepo.

### DIFF
--- a/packages/schema/src/plugins/enhancer/enhance/index.ts
+++ b/packages/schema/src/plugins/enhancer/enhance/index.ts
@@ -114,21 +114,18 @@ export class EnhancerGenerator {
 
         if (this.needsLogicalClient) {
             prismaTypesFixed = true;
-            resultPrismaTypeImport = `${LOGICAL_CLIENT_GENERATION_PATH}/index-fixed`;
+            resultPrismaTypeImport = LOGICAL_CLIENT_GENERATION_PATH;
             const result = await this.generateLogicalPrisma();
             dmmf = result.dmmf;
         }
 
         // reexport PrismaClient types (original or fixed)
-        const modelsDts = this.project.createSourceFile(
-            path.join(this.outDir, 'models.d.ts'),
+        const modelsTs = this.project.createSourceFile(
+            path.join(this.outDir, 'models.ts'),
             `export * from '${resultPrismaTypeImport}';`,
             { overwrite: true }
         );
-        await modelsDts.save();
-
-        // reexport values from the original PrismaClient (enums, etc.)
-        fs.writeFileSync(path.join(this.outDir, 'models.js'), `module.exports = require('${prismaImport}');`);
+        this.saveSourceFile(modelsTs);
 
         const authDecl = getAuthDecl(getDataModelAndTypeDefs(this.model));
         const authTypes = authDecl ? generateAuthType(this.model, authDecl) : '';
@@ -177,7 +174,7 @@ ${
         return {
             dmmf,
             newPrismaClientDtsPath: prismaTypesFixed
-                ? path.resolve(this.outDir, LOGICAL_CLIENT_GENERATION_PATH, 'index-fixed.d.ts')
+                ? path.resolve(this.outDir, LOGICAL_CLIENT_GENERATION_PATH, 'index.d.ts')
                 : undefined,
         };
     }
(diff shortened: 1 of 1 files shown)


ymc9:
Hi @diesal11 , I feel the entire approach is quite hacky now ... The original idea was to only reexport types from logical prisma client at "@zenstackhq/runtime/models", which was why previously only models.d.ts was generated. Then we found to import enums there you need value exports as well. For better safety, the "models.js" file reexports from the original prisma client, just to avoid accidentally importing duplicated copies of values from different prisma locations ...

I feel maybe it's an overkill. If we simply generate a "models.ts" to just reexport everything from logical prisma, the esm/cjs problem would just go away as it'll be compiled according to user's project settings. What do you think?

diesal11:
Yes I agree this isn't an ideal approach. I've had another look and I believe what you're suggesting is possible, it would also require saving `index-fixed.d.ts` as `index.d.ts` within the logical schema so that types & values can be exported at once within a ts file.

Initially I wasn't confident touching the index-fixed.d.ts file as I wasn't sure about the reason this was saved separately. But I've quickly tested this solution in my side project by manually applying the changes and it appears to work without issue. It looks like most of the value exports from prisma client are just enum like objects that map to strings. So it looks like it should be fine.

Thanks for the feedback! I've updated the PR to implement this, let me chase down test failures :)

ymc9 (commented):
I think the "/index" part can be omitted now.

packages/schema/src/plugins/enhancer/enhance/index.ts
```diff
@@ -114,21 +114,18 @@ export class EnhancerGenerator {
 
         if (this.needsLogicalClient) {
             prismaTypesFixed = true;
-            resultPrismaTypeImport = `${LOGICAL_CLIENT_GENERATION_PATH}/index-fixed`;
+            resultPrismaTypeImport = `${LOGICAL_CLIENT_GENERATION_PATH}/index`;
```
ymc9, line 117: I think the "/index" part can be omitted now.

packages/schema/src/plugins/enhancer/enhance/index.ts
```diff
@@ -114,21 +114,18 @@ export class EnhancerGenerator {
...
+        const modelsTs = this.project.createSourceFile(
+            path.join(this.outDir, 'models.ts'),
             `export * from '${resultPrismaTypeImport}';`,
             { overwrite: true }
         );
-        await modelsDts.save();
-
-        // reexport values from the original PrismaClient (enums, etc.)
-        fs.writeFileSync(path.join(this.outDir, 'models.js'), `module.exports = require('${prismaImport}');`);
+        await modelsTs.save();
```
ymc9, line 128: We can call `this.saveSourceFile(modelTs)` here.

ymc9:
> Yes I agree this isn't an ideal approach. I've had another look and I believe what you're suggesting is possible, it would also require saving `index-fixed.d.ts` as `index.d.ts` within the logical schema so that types & values can be exported at once within a ts file.
> 
> Initially I wasn't confident touching the index-fixed.d.ts file as I wasn't sure about the reason this was saved separately. But I've quickly tested this solution in my side project by manually applying the changes and it appears to work without issue. It looks like most of the value exports from prisma client are just enum like objects that map to strings. So it looks like it should be fine.
> 
> Thanks for the feedback! I've updated the PR to implement this, let me chase down test failures :)

Thanks for making the changes @diesal11 ! I've left two review comments, please help check when you get time.

diesal11:
This should be good to go now, pending green CI :)

ymc9 (approved):
Thanks! Looks good to me. I'll do a bit more testing on my side and then merge. Will be part of 2.14 release.

PR #2081, opened 2025-04-13
Enable JSON fields for SQLite

liontariai:
It's supported by prisma according to the docs https://www.prisma.io/docs/orm/reference/prisma-schema-reference#json
and I've been using it in my private projects for some time now by patching zenstack.

Also, here's the Prisma release announcing the feature: https://github.com/prisma/prisma/releases/tag/6.2.0

I was hoping for this to come to zenstack naturally but it seems to still be overlooked.

It's really just this one line, so I haven't run a testsuite or written tests for it.
I'm open to feedback as this is my first PR here, but I think it's really not worth putting too much overhead-effort into this one.


ymc9:
Hi @liontariai , thanks for making this PR! I've fixed a broken test case and also added test coverage for sqlite+json. Will merge it when CI passes.

ymc9 (approved):

PR #2097, opened 2025-04-24
fix(sdk): correct jsdoc indent

DoctorFTB:
Issue:

The documentation for jsdoc says there should be one space on each new line, currently there is none.
The eslint-disable doesn't work (doesn't miss the error) because the jsdoc is added above.

Solution:

Fix the jsdoc by adding spaces


ymc9:
Good catch! Thanks for making the fix @DoctorFTB !

PR #2099, opened 2025-04-29
feat(server): add ApiHandlerService to handle request with NestJS

ppodds:
When using `@zenstackhq/server` with NestJS, the API handler only works with the Fastify or Express adapter and requires registering middleware during module or app initialization. This approach creates a tight dependency on the framework implementation and makes it difficult to customize routing behavior.

This PR not only restructures the NestJS adapter folder to better align with NestJS conventions, but also introduces an `ApiHandlerService` that users can import and use. With this service, users have the flexibility to apply the handler to specific routes as needed and gain more control over the response generated by the API handler.

## Usage

`post.module.ts`

```typescript
import { Module } from '@nestjs/common';
import { PrismaService } from './prisma.service';
import { ZenStackModule, ApiHandlerService } from '@zenstackhq/server/nestjs';
import { enhance } from '@zenstackhq/runtime';
import { PostController } from './post.controller';
import { ClsService } from 'nestjs-cls';

@Module({
  imports: [ZenStackModule.registerAsync({
    useFactory: (prisma: PrismaService, clsService: ClsService) => {
      return {
        getEnhancedPrisma: () => enhance(prisma, { user: clsService.get('auth') }),
      };
    },
    inject: [PrismaService, ClsService],
    extraProviders: [PrismaService],
  })],
  controllers: [PostController],
  providers: [PrismaService, ApiHandlerService],
})
export class PostModule {}

```

`post.controller.ts`

```typescript
import { Controller, Get, Post } from '@nestjs/common';
import { ApiHandlerService } from '@zenstackhq/server/nestjs';

@Controller('post')
export class PostController {
  constructor(private readonly apiHandlerService: ApiHandlerService) {}

  // should align with the route generated by API handler
  @Get('findMany')
  async findMany() {
    return this.apiHandlerService.handleRequest()
  }

  @Post('create')
  async create() {
    return this.apiHandlerService.handleRequest()
  }
}
```

I’m happy to contribute additional tests or usage examples if needed. Suggestions or improvements are very welcome—I'm open to discussion!


packages/server/src/nestjs/interfaces/api-handler-options.interface.ts
```diff
@@ -0,0 +1,5 @@
+import { AdapterBaseOptions } from "../../types";
+
+export interface ApiHandlerOptions extends AdapterBaseOptions {
+    baseUrl?: string;
```
ymc9, line 4: Shall we add a short documentation for the field?

packages/server/src/nestjs/api-handler.service.ts
```diff
@@ -0,0 +1,56 @@
...
+import { RPCApiHandler } from '../api/rpc';
+import { ENHANCED_PRISMA } from "./zenstack.constants";
+import { ApiHandlerOptions } from './interfaces';
+
+/**
+ * The ZenStack API handler service for NestJS. The service is used to handle API requests
+ * and forward them to the ZenStack API handler. It is platform agnostic and can be used
+ * with any HTTP adapter.
+ */
+@Injectable({ scope: Scope.REQUEST })
```
ymc9, line 14: Is there a specific reason this needs to be Request scoped?

ymc9 (commented):
Hi @ppodds , many thanks for making this contribution! I think it's a good addition, with the service you can even implement a catch all route to automatically forward all requests to the underlying handler, right?

Will you be interested in changing the documentation as well? Basically after the zenstack module is registered, you can have two paths to using it: 1. directly inject the enhanced prisma client and use it in whatever service/controller you want; 2. use the api handler service to automate request handling.

packages/server/src/nestjs/api-handler.service.ts
ppodds: In order to inject the `Request` object in the constructor, we have to set the provider to `Scope.REQUEST`. Alternatively, we can use the default scope and pass the `Request` object when calling  `handleRequest`, but it will produce much code duplication, so I think a request-scoped provider is acceptable.

packages/server/src/nestjs/interfaces/api-handler-options.interface.ts
ppodds: Sure!

ppodds:
> Hi @ppodds , many thanks for making this contribution! I think it's a good addition, with the service you can even implement a catch all route to automatically forward all requests to the underlying handler, right?
> 
> Will you be interested in changing the documentation as well? Basically after the zenstack module is registered, you can have two paths to using it: 1. directly inject the enhanced prisma client and use it in whatever service/controller you want; 2. use the api handler service to automate request handling.

Sure! I will add it after the PR has been merged.

ymc9:
> > Hi @ppodds , many thanks for making this contribution! I think it's a good addition, with the service you can even implement a catch all route to automatically forward all requests to the underlying handler, right?
> > Will you be interested in changing the documentation as well? Basically after the zenstack module is registered, you can have two paths to using it: 1. directly inject the enhanced prisma client and use it in whatever service/controller you want; 2. use the api handler service to automate request handling.
> 
> Sure! I will add it after the PR has been merged.

Awesome! The piece of documentation is here: https://github.com/zenstackhq/zenstack-docs/blob/main/docs/reference/server-adapters/nestjs.mdx

I'm merging this PR.

PR #2114, opened 2025-05-09
feat: add elysiajs

rodrigoburigool:
Add [Elysia](https://elysiajs.com/) adapter

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -48,6 +48,7 @@
         "@types/supertest": "^2.0.12",
         "@zenstackhq/testtools": "workspace:*",
         "body-parser": "^1.20.2",
+        "elysia": "^1.3.1",
         "express": "^4.19.2",
         "fastify": "^4.14.1",
         "fastify-plugin": "^4.5.0",
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -21,7 +21,8 @@
         "express",
         "nextjs",
         "sveltekit",
-        "nuxtjs"
+        "nuxtjs",
+        "elysia"
     ],
     "author": "ZenStack Team",
     "license": "MIT",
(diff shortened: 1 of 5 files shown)


packages/server/src/elysia/handler.ts
```diff
@@ -0,0 +1,70 @@
...
+export interface ElysiaOptions extends AdapterBaseOptions {
+    /**
+     * Callback method for getting a Prisma instance for the given request context.
+     */
+    getPrisma: (context: ElysiaContext) => Promise<unknown> | unknown;
+}
+
+/**
+ * Creates an Elysia middleware handler for ZenStack.
+ * This handler provides RPC API functionality through Elysia's routing system.
```
ymc9, line 19: I think the implementation is not specific to RPC-style API. Should we say "This handler provides automatic CRUD APIs ..." instead?

packages/server/src/elysia/handler.ts
```diff
@@ -0,0 +1,70 @@
...
+            if (!prisma) {
+                set.status = 500;
+                return {
+                    message: 'unable to get prisma from request context'
+                };
+            }
+
+            const url = new URL(request.url);
+            const query = Object.fromEntries(url.searchParams);
+            const path = url.pathname;
```
ymc9, line 37: I'm not familiar with Elysia. Is it possible that the handler is mounted into a subpath (e.g., "/api/myapp/..."? If so, having the prefix in the `path` parameter passed to the API handler will break it. I guess we need to add a "basePath" option and use it to strip the path?

packages/server/package.json
```diff
@@ -48,6 +48,7 @@
         "@types/supertest": "^2.0.12",
         "@zenstackhq/testtools": "workspace:*",
         "body-parser": "^1.20.2",
+        "elysia": "^1.3.1",
```
ymc9, line 51: Please run `pnpm install` and include the lock file in the PR too, otherwise CI will fail.

ymc9 (changes requested):
Hi @rodrigoburigool, many thanks for making this PR!

I've added a few comments. Please help check when you have time.

ymc9 (approved):
Thanks for making the changes @rodrigoburigool ! I've made some small fixes to the tests and will merge when CI passes.

ymc9:
Btw, would you be interested in adding documentation for the new server adapter if you have time? @rodrigoburigool

PR #2144, opened 2025-06-10
chore: replace some unused/small/deprecated/long-updated dependencies and remove some spaces

DoctorFTB:
1. Replaced dependencies:

   * replace-in-file - most of the functionality is not used, its 15 lines code

   * change-case - most of the functionality is not used, v4 (currently used in project) depends on 12 more libraries, used many libraries inside (change-case -> param-case -> dot-case -> no-case -> lower-case), its 15 lines code
   * lower-case-first - deprecated
   * upper-case-first - deprecated
   * tiny-invariant - small library
   * is-plain-object - small library
   * sleep-promise - small library

   * jest-fetch-mock - unused
   * ts-node - unused (calle one once via npx)

2. About packages:
   * pnpm install show `Packages: +16 -189`

3. Move dependencies "decimal.js-light", "superjson" and "uuid" from dependencies to devDependencies in runtime, cuz used only for build

4. Remove spaces on empty lines or before next line.

Feel free to ask questions and discuss :)

P.S. Code was not tested locally. I got stuck after **tests ate 150 GB** of disk space.


DoctorFTB:
@ymc9 Hey. Should you check it? 👀

ymc9:
> @ymc9 Hey. Should you check it? 👀

Absolutely! Will finish reviewing it asap.

packages/runtime/package.json
```diff
@@ -96,22 +100,14 @@
...
         "pluralize": "^8.0.0",
         "safe-json-stringify": "^1.2.0",
         "semver": "^7.5.2",
-        "superjson": "^1.13.0",
-        "tiny-invariant": "^1.3.1",
         "traverse": "^0.6.10",
         "ts-pattern": "^4.3.0",
         "tslib": "^2.4.1",
-        "upper-case-first": "^2.0.2",
-        "uuid": "^9.0.0",
```
ymc9, line 114: "uuid", "decimal.js-light", and "superjson" are actually needed at runtime.

packages/runtime/src/local-helpers.ts
```diff
@@ -0,0 +1 @@
+export * from './local-helpers/index';
```
ymc9: Is this file not needed?

ymc9 (commented):
Hi @DoctorFTB , thanks for making this clean up and my apologies for the delay. Please check the review comments attached. Thanks!

ymc9:
Btw, just curious, on what aspect does having more dependencies hurt you? NPM installation slowness or bundling time?

packages/runtime/package.json
DoctorFTB: Oops, missed it. First time I searched I didn't find any use of import and thought "its uses bundle when compiling because it's weird to use full dependency in browser", now I found import and revert delete

packages/runtime/src/local-helpers.ts
DoctorFTB: This file is used in package json exports, but I remove it and use direct use of index file

DoctorFTB:
> Btw, just curious, on what aspect does having more dependencies hurt you? NPM installation slowness or bundling time?

Both, and also the large Docker image size (we have a layer with full set of node_modules for cache) :(
For example: library traverse using about 60+ additional libraries (checked via dependencies "traverse -> typedarray.prototype.slice -> es-abstract -> 54 dependencies"), but if you only use "traverse(obj).forEach" or "traverse(obj).map" without int8/uint8/int16/etc arrays - this (typedarray.prototype.slice) doesnt used

DoctorFTB:
FYI, I'm also preparing a replacement for "traverse" library, I can push changes here (but I need some time to implement them) or I can make another PR with changes

ymc9:
> FYI, I'm also preparing a replacement for "traverse" library, I can push changes here (but I need some time to implement them) or I can make another PR with changes

Got it. Thanks for the explanation. Maybe we can get this PR merged first? I'm planning a release today or tomorrow.

DoctorFTB:
> Maybe we can get this PR merged first?

Yes, do it! Don't worry, I can make one more PR without any problems :)

PR #2145, opened 2025-06-10
fix(schema): typo "third"

DoctorFTB:


ymc9:
Thank you @DoctorFTB 😄

PR #2163, opened 2025-06-24
.count use same flow as findMany

PodkopovP:
The current .count implementation (and probably other aggregate functions) create a different .where clause to findMany - this means that doing a .count and .where with the same inputs, could (after zenstacks RLS implementation) cause be looking at different results - i.e. you might end up with 4 results from .findMany, but .count says there should be 10 results.

### DIFF
--- a/packages/runtime/src/enhancements/node/policy/handler.ts
+++ b/packages/runtime/src/enhancements/node/policy/handler.ts
@@ -1654,7 +1654,7 @@ export class PolicyProxyHandler<DbClient extends DbClientContract> implements Pr
     }
 
     count(args: any) {
-        return createDeferredPromise<unknown[]>(() => this.doFind(args, 'findMany', () => []));
+        return createDeferredPromise<unknown[]>(() => this.doFind(args, 'count', () => []));
     }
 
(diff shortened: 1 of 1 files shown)


PodkopovP:
The two issues I found so far:

- Including relations in select/include can get funny and can cause a mismatch between .findMany / count - .count doesnt support including relations, so the current implementation doesn't support injecting guards based on relations. The solution I have here is to pass .count through the same funnel .findMany goes through, but override select to true before the final prisma call is made - this lets zenstack inject relevant relational queries to the where clause
- .count currently injects using .injectAuthGuardAsWhere, which always wraps the where clause in 'AND' - this isn't the same process as .findMany goes through (haven't done too much investigation to see what .findMany does, but it seems to more smartly combine the original & new args)

JoeRoddy:
Relevant discord discussion from a similar issue.

https://discord.com/channels/1035538056146595961/1338578707086049372

I was never able to create a simple proof of concept replication.  It was only happening in our large real world project with highly complex rules across tables, including using `check()` across tables.

ymc9:
Closing for now.

PR #2171, opened 2025-07-05
fix: improve support for new `prisma-client` generator

genu:
Improves support for new `prisma-client` generator.

@ymc9 What I think needs some attention, which I wasn't able to figure out. The `enhance.ts` that is generated imports these:

`import type { InternalArgs, DynamicClientExtensionThis } from '../prisma/client/runtime/library';`

But I don't believe these exist anymore with the new `prisma-client` or they are relocated. I'm not exactly sure here.

This impacts the types when we do:

```ts
const db = enhance(prisma) //<-- db is any
```

### DIFF
--- a/packages/schema/src/plugins/enhancer/enhance/index.ts
+++ b/packages/schema/src/plugins/enhancer/enhance/index.ts
@@ -121,9 +121,13 @@ export class EnhancerGenerator {
         }
 
         // reexport PrismaClient types (original or fixed)
+        const modelsTsContent = `export * from '${resultPrismaTypeImport}';${
+            this.isNewPrismaClientGenerator ? '\nexport * from \'./json-fields\';' : ''
+        }`;
+        
         const modelsTs = this.project.createSourceFile(
             path.join(this.outDir, 'models.ts'),
-            `export * from '${resultPrismaTypeImport}';`,
+            modelsTsContent,
             { overwrite: true }
         );
         this.saveSourceFile(modelsTs);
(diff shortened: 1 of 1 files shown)


packages/schema/src/plugins/enhancer/enhance/index.ts
```diff
@@ -502,6 +520,44 @@ export type Enhanced<Client> =
...
+                    if (structure.type && typeof structure.type === 'string' && structure.type.includes('$Types')) {
+                        needsTypesImport = true;
+                    }
+                }
+            });
+
+            // Add $Types import if needed
+            if (needsTypesImport) {
+                sfNew.addStatements('import $Types = runtime.Types;');
+            }
```
genu: Maybe we don't need to do this check, and just add it anyway?

ymc9:
> DynamicClientExtensionThis

Thanks for making the fix @genu ! These types can indeed be imported directly from "@prisma/client/runtime/library" instead. The code changes look good to me, and I'll make some minor tweaks and add a test then merge.

Thank you!

genu:
> > DynamicClientExtensionThis
> 
> Thanks for making the fix @genu ! These types can indeed be imported directly from "@prisma/client/runtime/library" instead. The code changes look good to me, and I'll make some minor tweaks and add a test then merge.
> 
> Thank you!

@ymc9 From my observation, this partly resolved the issue.

type inference is lost when using prisma extensions:

```ts
const prisma = new PrismaClient()
const db = enhance(prisma)

db.user.find() // <-- db.user is properly typed
```

With Enhanced client:

```ts
const prisma = new PrismaClient().$extends({})

const db = enhance(prisma)

db.user.find() // <-- db.user is any
```

PR #2177, opened 2025-07-08
chore: Improve performance for prisma-client generation

mhodgson:
In our fairly large schema with delegate types we saw our `zenstack generate` time balloon up to almost 10 minutes when switching to the new 'prisma-client' generator. I was able to track down the performance issue to the use of ts-morph in overwriting the model/*.ts files. 

This change does two things:
1. Accumulates the model type changes into a temporary list of statements before updating the source file. 
2. Moves the file overwriting outside of the ts-morph usage, making it nearly instantaneous.

It is likely these same changes would speed up the existing "legacy" implementation as well, but probably not as noticeably since there is only one file being moved instead of one for each model.


mhodgson:
@ymc9 Any thoughts here? I'm still tracking down some other issues related to how Zenstack re-exports the logical-prisma-client, but wondering if these changes make sense to you.

ymc9:
I'll finish reviewing it today @mhodgson . Thank you for making the PR!

packages/schema/src/plugins/enhancer/enhance/index.ts
```diff
@@ -492,6 +493,32 @@ export type Enhanced<Client> =
     private async processClientTypesNewPrismaGenerator(prismaClientDir: string, delegateInfo: DelegateInfo) {
         const project = new Project();
 
+        // remove delegate_aux_* fields from the prismaNamespace.ts
```
ymc9: I looked into prismaNamespace.ts and it seems to contain only shared types and top-level operation definitions. I haven't found where it involves `delegate_aux_*` declarations.

Is it a precaution or with your zmodel it happens?

packages/schema/src/plugins/enhancer/enhance/index.ts
```diff
@@ -974,7 +1036,7 @@ export type Enhanced<Client> =
     }
 
     private trimEmptyLines(source: string): string {
-        return source.replace(/^\s*[\r\n]/gm, '');
+        return source.replace(/^\s*\,?[\r\n]/gm, '');
```
ymc9: Linter complains about this unnecessary escape.

packages/schema/src/plugins/enhancer/enhance/index.ts
```diff
@@ -641,6 +677,26 @@ export type Enhanced<Client> =
...
+        const structure = variable.getStructure();
+
+        // remove `delegate_aux_*` fields from the variable's initializer
+        const auxFields = this.findAuxProps(variable);
+        if (auxFields.length > 0) {
+            structure.declarations.forEach((variable) => {
+                if (variable.initializer) {
+                    let source = variable.initializer;
+                    auxFields.forEach((f) => {
+                        source = this.removeFromSource(source, f.getText());
```
ymc9, line 665: It seems a casting to string is needed here.

ymc9:
Hi @mhodgson I guess you've been busy. I've made several changes based on my judgment and will merge the PR. Please take a look when you have time. About to make a new patch release soon.

mhodgson:
@ymc9 yes, sorry, was about to get back to this in the morning. I have some broader issues with the `prisma-client` generator support here that I need to sync with you on. What is boils down to is that the new generator creates two "barrel" style files that are meant for exporting just the models and enums (named `models.ts` and `enums.ts`). These are safe for an application to import and use for type safety in server or client code. 

In the current `model.ts` file that Zenstack is creating, there entire Prisma `client.ts` is re-exported, which contains server-specific code using node module likes path and fs. My recommendation is that Zenstack mirrors the Prisma generated code and simply exports *only* the models in `models.ts` and adds a new `enums.ts` that does the same (re-exporting the `models.ts` and `enums.ts` files from the `logical-prisma-client`). However, this leaves a glaring gap, because the full `Prisma` namespace is then no longer exported anywhere from Zenstack. To address this, we need to decide if we should:
1. re-export the full Prisma object from the `logical-prisma-client/client.ts` file from the `enhance.ts` in Zenstack or...
2. Add a new `client.ts` file in Zenstack that just re-exports the `logical-prisma-client/client.ts`. 

Thoughts? I'm happy to provide another PR, but need your thoughts and guidance here as it is really up to you. As it currently stands, the Zenstack exports are incompatible with any framework trying to use Prisma model or enum types in a client side bundle.

ymc9:
> @ymc9 yes, sorry, was about to get back to this in the morning. I have some broader issues with the `prisma-client` generator support here that I need to sync with you on. What is boils down to is that the new generator creates two "barrel" style files that are meant for exporting just the models and enums (named `models.ts` and `enums.ts`). These are safe for an application to import and use for type safety in server or client code.
> 
> In the current `model.ts` file that Zenstack is creating, there entire Prisma `client.ts` is re-exported, which contains server-specific code using node module likes path and fs. My recommendation is that Zenstack mirrors the Prisma generated code and simply exports _only_ the models in `models.ts` and adds a new `enums.ts` that does the same (re-exporting the `models.ts` and `enums.ts` files from the `logical-prisma-client`). However, this leaves a glaring gap, because the full `Prisma` namespace is then no longer exported anywhere from Zenstack. To address this, we need to decide if we should:
> 
> 1. re-export the full Prisma object from the `logical-prisma-client/client.ts` file from the `enhance.ts` in Zenstack or...
> 2. Add a new `client.ts` file in Zenstack that just re-exports the `logical-prisma-client/client.ts`.
> 
> Thoughts? I'm happy to provide another PR, but need your thoughts and guidance here as it is really up to you. As it currently stands, the Zenstack exports are incompatible with any framework trying to use Prisma model or enum types in a client side bundle.

Hi @mhodgson , thanks for following up. I think this is not specific to the new "prisma-client" generator, right? With the old generator, the `model.ts` file also reexports everything from either the logical client or the original prisma client.

For using the exported model types, does it work if you do a type-only import? I assume bundlers will refrain from bundling js code if everything it seems is type import. But I haven't tried. If so, is it specifically about importing enums (which have to be value imports)?

I agree adding a new `client.ts` is a good idea - to keep all other files (`models.ts`, and a new `enum.ts`) completely safe for frontend import. Just wanted to have a good understanding of the real issue first.

mhodgson:
@ymc9 you are correct. This worked with the `prisma-client-js` generator, but I think it was because the previous Prisma Client implementation was safe to load on the client since the server specific parts were wrapped up in the rust binary. Now, the `client.ts` file starts immediately with using server specific code that cannot be used in the browser. 

You are right that just importing types works fine, but in this case there are also javascript values (namely the enums) that must be loaded, so just importing types leaves a gap. In our Remix (react-router) codebase I didn't discover this issue until running into actual front-end errors where the code was trying to load files from disk. I'm hopeful that having a clear separation of what files are browser safe vs. meant for the server will help. 

I thought about raising this issue with the Prisma folks, but I think loading anything out of Prisma called `client` is safe to assume should not be loaded on the client side, so they are probably correct to assume they can use node specific file access APIs there. 

I like the idea of just mirroring the structure of the Prisma Client as much as possible with the Zenstack files/exports. It should make onboarding of new users clean and clear, and adheres to the principal of least surprise. Then the only new/extra part of using Zenstack is the `enhance.ts` file that is probably only accessed once where the database connection is established.

ymc9:
> @ymc9 you are correct. This worked with the `prisma-client-js` generator, but I think it was because the previous Prisma Client implementation was safe to load on the client since the server specific parts were wrapped up in the rust binary. Now, the `client.ts` file starts immediately with using server specific code that cannot be used in the browser.
> 
> You are right that just importing types works fine, but in this case there are also javascript values (namely the enums) that must be loaded, so just importing types leaves a gap. In our Remix (react-router) codebase I didn't discover this issue until running into actual front-end errors where the code was trying to load files from disk. I'm hopeful that having a clear separation of what files are browser safe vs. meant for the server will help.
> 
> I thought about raising this issue with the Prisma folks, but I think loading anything out of Prisma called `client` is safe to assume should not be loaded on the client side, so they are probably correct to assume they can use node specific file access APIs there.
> 
> I like the idea of just mirroring the structure of the Prisma Client as much as possible with the Zenstack files/exports. It should make onboarding of new users clean and clear, and adheres to the principal of least surprise. Then the only new/extra part of using Zenstack is the `enhance.ts` file that is probably only accessed once where the database connection is established.

Thanks for the clarification. We're on the same page then. A PR is greatly appreciated if you have time!

mhodgson:
@ymc9 PR for discussed changes: https://github.com/zenstackhq/zenstack/pull/2184

mhodgson:
@ymc9 just tested out 2.17.1 and can confirm that I'm still getting problematic fields generated into the Prisma namespace that are causing type errors in the modified models in Zenstack. Here's an excerpt from `logical-prisma-client/internal/prismaNamespace.ts` for our schema:

```
export const TaskStateTransitionScalarFieldEnum = {
  id: 'id',
  taskId: 'taskId',
  fromState: 'fromState',
  toState: 'toState',
  transitionedAt: 'transitionedAt',
  transitionedBy: 'transitionedBy',
  data: 'data',
  delegate_aux_TaskStateTransition_taskI_0: 'delegate_aux_TaskStateTransition_taskI_0',
  delegate_aux_TaskStateTransition_taskI_1: 'delegate_aux_TaskStateTransition_taskI_1',
  delegate_aux_TaskStateTransition_taskI_2: 'delegate_aux_TaskStateTransition_taskI_2',
  delegate_aux_TaskStateTransition_taskI_3: 'delegate_aux_TaskStateTransition_taskI_3',
  delegate_aux_TaskStateTransition_taskI_4: 'delegate_aux_TaskStateTransition_taskI_4',
  delegate_aux_TaskStateTransition_taskI_5: 'delegate_aux_TaskStateTransition_taskI_5',
  delegate_aux_TaskStateTransition_taskI_6: 'delegate_aux_TaskStateTransition_taskI_6',
  delegate_aux_TaskStateTransition_taskI_7: 'delegate_aux_TaskStateTransition_taskI_7',
  delegate_aux_TaskStateTransition_taskI_8: 'delegate_aux_TaskStateTransition_taskI_8',
  delegate_aux_TaskStateTransition_taskI_9: 'delegate_aux_TaskStateTransition_taskI_9',
  delegate_aux_TaskStateTransition_taskI_10: 'delegate_aux_TaskStateTransition_taskI_10'
} as const

export type TaskStateTransitionScalarFieldEnum = (typeof TaskStateTransitionScalarFieldEnum)[keyof typeof TaskStateTransitionScalarFieldEnum]


export const TaskRevisionScalarFieldEnum = {
  deletedAt: 'deletedAt',
  createdAt: 'createdAt',
  updatedAt: 'updatedAt',
  id: 'id',
  taskId: 'taskId',
  note: 'note',
  reason: 'reason',
  status: 'status',
  createdBy: 'createdBy',
  delegate_aux_TaskRevision_taskId_Produ_0: 'delegate_aux_TaskRevision_taskId_Produ_0',
  delegate_aux_TaskRevision_taskId_ListP_0: 'delegate_aux_TaskRevision_taskId_ListP_0',
  delegate_aux_TaskRevision_taskId_Paren_0: 'delegate_aux_TaskRevision_taskId_Paren_0',
  delegate_aux_TaskRevision_taskId_Docum_0: 'delegate_aux_TaskRevision_taskId_Docum_0',
  delegate_aux_TaskRevision_taskId_Contr_0: 'delegate_aux_TaskRevision_taskId_Contr_0',
  delegate_aux_TaskRevision_taskId_MPOFu_0: 'delegate_aux_TaskRevision_taskId_MPOFu_0',
  delegate_aux_TaskRevision_taskId_MPOFu_1: 'delegate_aux_TaskRevision_taskId_MPOFu_1',
  delegate_aux_TaskRevision_taskId_BuyWi_0: 'delegate_aux_TaskRevision_taskId_BuyWi_0',
  delegate_aux_TaskRevision_taskId_Price_0: 'delegate_aux_TaskRevision_taskId_Price_0',
  delegate_aux_TaskRevision_taskId_Prici_0: 'delegate_aux_TaskRevision_taskId_Prici_0',
  delegate_aux_TaskRevision_taskId_OpsTa_0: 'delegate_aux_TaskRevision_taskId_OpsTa_0'
} as const
```

These results in type errors in the modified models like so:
```
src/generated/database/zenstack/logical-prisma-client/models/TaskStateTransition.ts:150:69 - error TS2344: Type 'T["by"]' does not satisfy the constraint 'Enumerable<keyof TaskStateTransitionGroupByOutputType>'.
  Type 'TaskStateTransitionScalarFieldEnum | TaskStateTransitionScalarFieldEnum[]' is not assignable to type 'Enumerable<keyof TaskStateTransitionGroupByOutputType>'.
    Type '"delegate_aux_TaskStateTransition_taskI_0"' is not assignable to type 'Enumerable<keyof TaskStateTransitionGroupByOutputType>'.
      Type 'T["by"]' is not assignable to type '(keyof TaskStateTransitionGroupByOutputType)[]'.
        Type 'TaskStateTransitionScalarFieldEnum | TaskStateTransitionScalarFieldEnum[]' is not assignable to type '(keyof TaskStateTransitionGroupByOutputType)[]'.
          Type 'string' is not assignable to type '(keyof TaskStateTransitionGroupByOutputType)[]'.

150         Prisma.PickEnumerable<TaskStateTransitionGroupByOutputType, T['by']> &
```

Thoughts? I think we need to bring back that chunk of code I had added to address this.

ymc9:
> @ymc9 just tested out 2.17.1 and can confirm that I'm still getting problematic fields generated into the Prisma namespace that are causing type errors in the modified models in Zenstack. Here's an excerpt from `logical-prisma-client/internal/prismaNamespace.ts` for our schema:
> 
> ```
> export const TaskStateTransitionScalarFieldEnum = {
>   id: 'id',
>   taskId: 'taskId',
>   fromState: 'fromState',
>   toState: 'toState',
>   transitionedAt: 'transitionedAt',
>   transitionedBy: 'transitionedBy',
>   data: 'data',
>   delegate_aux_TaskStateTransition_taskI_0: 'delegate_aux_TaskStateTransition_taskI_0',
>   delegate_aux_TaskStateTransition_taskI_1: 'delegate_aux_TaskStateTransition_taskI_1',
>   delegate_aux_TaskStateTransition_taskI_2: 'delegate_aux_TaskStateTransition_taskI_2',
>   delegate_aux_TaskStateTransition_taskI_3: 'delegate_aux_TaskStateTransition_taskI_3',
>   delegate_aux_TaskStateTransition_taskI_4: 'delegate_aux_TaskStateTransition_taskI_4',
>   delegate_aux_TaskStateTransition_taskI_5: 'delegate_aux_TaskStateTransition_taskI_5',
>   delegate_aux_TaskStateTransition_taskI_6: 'delegate_aux_TaskStateTransition_taskI_6',
>   delegate_aux_TaskStateTransition_taskI_7: 'delegate_aux_TaskStateTransition_taskI_7',
>   delegate_aux_TaskStateTransition_taskI_8: 'delegate_aux_TaskStateTransition_taskI_8',
>   delegate_aux_TaskStateTransition_taskI_9: 'delegate_aux_TaskStateTransition_taskI_9',
>   delegate_aux_TaskStateTransition_taskI_10: 'delegate_aux_TaskStateTransition_taskI_10'
> } as const
> 
> export type TaskStateTransitionScalarFieldEnum = (typeof TaskStateTransitionScalarFieldEnum)[keyof typeof TaskStateTransitionScalarFieldEnum]
> 
> 
> export const TaskRevisionScalarFieldEnum = {
>   deletedAt: 'deletedAt',
>   createdAt: 'createdAt',
>   updatedAt: 'updatedAt',
>   id: 'id',
>   taskId: 'taskId',
>   note: 'note',
>   reason: 'reason',
>   status: 'status',
>   createdBy: 'createdBy',
>   delegate_aux_TaskRevision_taskId_Produ_0: 'delegate_aux_TaskRevision_taskId_Produ_0',
>   delegate_aux_TaskRevision_taskId_ListP_0: 'delegate_aux_TaskRevision_taskId_ListP_0',
>   delegate_aux_TaskRevision_taskId_Paren_0: 'delegate_aux_TaskRevision_taskId_Paren_0',
>   delegate_aux_TaskRevision_taskId_Docum_0: 'delegate_aux_TaskRevision_taskId_Docum_0',
>   delegate_aux_TaskRevision_taskId_Contr_0: 'delegate_aux_TaskRevision_taskId_Contr_0',
>   delegate_aux_TaskRevision_taskId_MPOFu_0: 'delegate_aux_TaskRevision_taskId_MPOFu_0',
>   delegate_aux_TaskRevision_taskId_MPOFu_1: 'delegate_aux_TaskRevision_taskId_MPOFu_1',
>   delegate_aux_TaskRevision_taskId_BuyWi_0: 'delegate_aux_TaskRevision_taskId_BuyWi_0',
>   delegate_aux_TaskRevision_taskId_Price_0: 'delegate_aux_TaskRevision_taskId_Price_0',
>   delegate_aux_TaskRevision_taskId_Prici_0: 'delegate_aux_TaskRevision_taskId_Prici_0',
>   delegate_aux_TaskRevision_taskId_OpsTa_0: 'delegate_aux_TaskRevision_taskId_OpsTa_0'
> } as const
> ```
> 
> These results in type errors in the modified models like so:
> 
> ```
> src/generated/database/zenstack/logical-prisma-client/models/TaskStateTransition.ts:150:69 - error TS2344: Type 'T["by"]' does not satisfy the constraint 'Enumerable<keyof TaskStateTransitionGroupByOutputType>'.
>   Type 'TaskStateTransitionScalarFieldEnum | TaskStateTransitionScalarFieldEnum[]' is not assignable to type 'Enumerable<keyof TaskStateTransitionGroupByOutputType>'.
>     Type '"delegate_aux_TaskStateTransition_taskI_0"' is not assignable to type 'Enumerable<keyof TaskStateTransitionGroupByOutputType>'.
>       Type 'T["by"]' is not assignable to type '(keyof TaskStateTransitionGroupByOutputType)[]'.
>         Type 'TaskStateTransitionScalarFieldEnum | TaskStateTransitionScalarFieldEnum[]' is not assignable to type '(keyof TaskStateTransitionGroupByOutputType)[]'.
>           Type 'string' is not assignable to type '(keyof TaskStateTransitionGroupByOutputType)[]'.
> 
> 150         Prisma.PickEnumerable<TaskStateTransitionGroupByOutputType, T['by']> &
> ```
> 
> Thoughts? I think we need to bring back that chunk of code I had added to address this.

Thanks for sharing it! Let me check and get that code back as needed @mhodgson

PR #2184, opened 2025-07-14
fix: Update re-exports for new prisma-client generator

mhodgson:
This should update the re-exporting of the Prisma logical client when using the new `prisma-client` generator, consistent with discussions [here](https://github.com/zenstackhq/zenstack/pull/2177#issuecomment-3069340310)


ymc9:
Thanks for continuing to work on this @mhodgson . I feel this part of logic has become deeply entangled. It started when we introduced the logical client, and now with two flavors of generators 😂.

I've decided to do a local refactor based on your PR. Will make the changes, test and merge then.

ymc9 (approved):

mhodgson:
thanks! This looks great.

Issue #2178, opened 2025-07-10
Support pluralized REST resource names (ie. /users vs. /user)


lsmith77:
**Is your feature request related to a problem? Please describe.**
It is a common pattern to have a model `user` exposed as `/users` and `/users/123`, ie. the URL to the collection is the pluralized model name.

**Describe the solution you'd like**
Some way to define a mapping or automatic pluralization for the purposes of all exposed URLs via the RestHandler (also including the openapi plugin).

As discussed via Discord at least for v2 we will use a mapping:
https://discord.com/channels/1035538056146595961/1390310045031464981

```
I thinking maybe we can make a smaller change, without introducing a zmodel attribute. Instead, just pass a mapping at runtime when creating the api handler:

handler: RestApiHandler({
    endpoint: ...,
    modelNameMapping: { user: 'users', ... }
})


The mapping is simply used right before passing the call to the underlying prismaclient
Something like
let match = this.urlPatterns.single.match(path);
const prismaModel = this.lookupMappedModelName(match.type);
return await this.processSingleRead(prisma, prismaModel, match.id, query);
```

**Describe alternatives you've considered**
Accepting the current status quo, ie. either using `/user` or pluralizing the model name from `User` to `Users`.

**Additional context**
https://restfulapi.net/resource-naming/

lsmith77:
FTR we are working to implement this. I hope to have a PR ready next week.

ymc9:
Implemented in 2.17.0

PR #2187, opened 2025-07-15
feat: add ability to map model names in the URLS/JSON response

lsmith77:
implements https://github.com/zenstackhq/zenstack/issues/2178

this PR also adds support for a `prefix` similar to the openapi plugin. note openapi plugin itself still needs to be updated, to also use this model name mapping.

also we did not add any new tests. would appreciate any hints on that.

finally we will of course also be happy to provide documentation for this.

/cc @tobiasbrugger

### DIFF
--- a/packages/plugins/openapi/src/rest-generator.ts
+++ b/packages/plugins/openapi/src/rest-generator.ts
@@ -44,6 +44,7 @@ type Policies = ReturnType<typeof analyzePolicies>;
  */
 export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
     private warnings: string[] = [];
+    private modelNameMapping: Record<string, string>;
 
     constructor(protected model: Model, protected options: PluginOptions, protected dmmf: DMMF.Document) {
         super(model, options, dmmf);
@@ -51,6 +52,8 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
         if (this.options.omitInputDetails !== undefined) {
             throw new PluginError(name, '"omitInputDetails" option is not supported for "rest" flavor');
         }
+
+        this.modelNameMapping = this.getOption('modelNameMapping', {} as Record<string, string>);
     }
 
     generate() {
@@ -126,6 +129,10 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
         return result;
     }
 
+    private mapModelName(modelName: string): string {
+        return this.modelNameMapping[modelName] ?? modelName;
+    }
+
     private generatePathsForModel(model: DMMF.Model, zmodel: DataModel): OAPI.PathItemObject | undefined {
         const result: Record<string, OAPI.PathItemObject> = {};
 
@@ -139,9 +146,11 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
 
         const resourceMeta = getModelResourceMeta(zmodel);
 
+        const modelName = this.mapModelName(model.name);
+
         // GET /resource
         // POST /resource
-        result[`${prefix}/${lowerCaseFirst(model.name)}`] = {
+        result[`${prefix}/${lowerCaseFirst(modelName)}`] = {
             get: this.makeResourceList(zmodel, policies, resourceMeta),
             post: this.makeResourceCreate(zmodel, policies, resourceMeta),
         };
@@ -150,10 +159,10 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
         // PUT /resource/{id}
         // PATCH /resource/{id}
         // DELETE /resource/{id}
-        result[`${prefix}/${lowerCaseFirst(model.name)}/{id}`] = {
+        result[`${prefix}/${lowerCaseFirst(modelName)}/{id}`] = {
             get: this.makeResourceFetch(zmodel, policies, resourceMeta),
-            put: this.makeResourceUpdate(zmodel, policies, `update-${model.name}-put`, resourceMeta),
-            patch: this.makeResourceUpdate(zmodel, policies, `update-${model.name}-patch`, resourceMeta),
+            put: this.makeResourceUpdate(zmodel, policies, `update-${modelName}-put`, resourceMeta),
+            patch: this.makeResourceUpdate(zmodel, policies, `update-${modelName}-patch`, resourceMeta),
             delete: this.makeResourceDelete(zmodel, policies, resourceMeta),
         };
 
@@ -165,14 +174,14 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
             }
 
             // GET /resource/{id}/{relationship}
-            const relatedDataPath = `${prefix}/${lowerCaseFirst(model.name)}/{id}/${field.name}`;
+            const relatedDataPath = `${prefix}/${lowerCaseFirst(modelName)}/{id}/${field.name}`;
             let container = result[relatedDataPath];
             if (!container) {
                 container = result[relatedDataPath] = {};
             }
             container.get = this.makeRelatedFetch(zmodel, field, relationDecl, resourceMeta);
 
-            const relationshipPath = `${prefix}/${lowerCaseFirst(model.name)}/{id}/relationships/${field.name}`;
+            const relationshipPath = `${prefix}/${lowerCaseFirst(modelName)}/{id}/relationships/${field.name}`;
             container = result[relationshipPath];
             if (!container) {
                 container = result[relationshipPath] = {};
--- a/packages/server/src/api/rest/index.ts
+++ b/packages/server/src/api/rest/index.ts
@@ -276,16 +276,27 @@ class RequestHandler extends APIHandlerBase {
         return this.modelNameMapping[modelName] ?? modelName;
     }
 
-    private matchUrlPattern(path: string, routeType: UrlPatterns): Match {
+    private matchUrlPattern(path: string, routeType: UrlPatterns): Match | undefined {
         const pattern = this.urlPatternMap[routeType];
         if (!pattern) {
             throw new InvalidValueError(`Unknown route type: ${routeType}`);
         }
 
         const match = pattern.match(path);
-        if (match) {
-            match.type = this.reverseModelNameMapping[match.type] ?? match.type;
+        if (!match) {
+            return;
         }
+
+        if (match.type in this.modelNameMapping) {
+            throw new InvalidValueError(
+                `use the mapped model name: ${this.modelNameMapping[match.type]} and not ${match.type}`
+            );
+        }
+
+        if (match.type in this.reverseModelNameMapping) {
+            match.type = this.reverseModelNameMapping[match.type];
+        }
+
         return match;
     }
 
--- a/packages/server/src/api/rest/index.ts
+++ b/packages/server/src/api/rest/index.ts
@@ -245,9 +245,7 @@ class RequestHandler extends APIHandlerBase {
         const segmentCharset = options.urlSegmentCharset ?? 'a-zA-Z0-9-_~ %';
 
         this.modelNameMapping = options.modelNameMapping ?? {};
-        this.reverseModelNameMapping = Object.fromEntries(
-            Object.entries(this.modelNameMapping).map(([k, v]) => [v, k])
-        );
+        this.reverseModelNameMapping = Object.fromEntries(Object.entries(this.modelNameMapping).map(([k, v]) => [v, k]));
         this.urlPatternMap = this.buildUrlPatternMap(segmentCharset);
     }
 
@@ -269,8 +267,8 @@ class RequestHandler extends APIHandlerBase {
         };
     }
 
-    private reverseModelNameMap(type: string): string {
-        return this.reverseModelNameMapping[type] ?? type;
+    private mapModelName(modelName: string): string {
+        return this.modelNameMapping[modelName] ?? modelName;
     }
 
     private matchUrlPattern(path: string, routeType: UrlPatterns): Match {
@@ -281,7 +279,7 @@ class RequestHandler extends APIHandlerBase {
 
         const match = pattern.match(path);
         if (match) {
-            match.type = this.modelNameMapping[match.type] ?? match.type;
+            match.type = this.reverseModelNameMapping[match.type] ?? match.type;
         }
         return match;
     }
@@ -569,7 +567,7 @@ class RequestHandler extends APIHandlerBase {
         }
 
         if (entity?.[relationship]) {
-            const mappedType = this.reverseModelNameMap(type);
+            const mappedType = this.mapModelName(type);
             return {
                 status: 200,
                 body: await this.serializeItems(relationInfo.type, entity[relationship], {
@@ -621,7 +619,7 @@ class RequestHandler extends APIHandlerBase {
         }
 
         const entity: any = await prisma[type].findUnique(args);
-        const mappedType = this.reverseModelNameMap(type);
+        const mappedType = this.mapModelName(type);
 
         if (entity?._count?.[relationship] !== undefined) {
             // build up paginator
@@ -720,7 +718,7 @@ class RequestHandler extends APIHandlerBase {
             ]);
             const total = count as number;
 
-            const mappedType = this.reverseModelNameMap(type);
+            const mappedType = this.mapModelName(type);
             const url = this.makeNormalizedUrl(`/${mappedType}`, query);
             const options: Partial<SerializerOptions> = {
                 include,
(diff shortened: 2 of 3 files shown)


packages/server/src/api/rest/index.ts
```diff
@@ -220,29 +237,60 @@ class RequestHandler extends APIHandlerBase {
...
+            Object.entries(this.modelNameMapping).map(([k, v]) => [v, k])
+        );
+        this.urlPatternMap = this.buildUrlPatternMap(segmentCharset);
     }
 
-    buildUrlPatterns(idDivider: string, urlSegmentNameCharset: string) {
+    private buildUrlPatternMap(urlSegmentNameCharset: string): Record<UrlPatterns, UrlPattern> {
         const options = { segmentValueCharset: urlSegmentNameCharset };
+
+        const buildPath = (segments: string[]) => {
```
lsmith77, line 260: if you prefer we can also drop this change and use `string` rather than `string[]` in the constructor of `UrlPattern`.

packages/server/src/api/rest/index.ts
```diff
@@ -1147,7 +1197,7 @@ class RequestHandler extends APIHandlerBase {
     }
 
     private makeLinkUrl(path: string) {
-        return `${this.options.endpoint}${path}`;
+        return `${this.options.endpoint}${this.prefix}${path}`;
```
ymc9: I think we need to handle `this.prefix` undefined here. Also probably make it more robust with the prefix having/not-having leading/trailing slashes? Maybe we `path.posix.join`?

Of course the prerequisite is this option is needed.

packages/server/src/api/rest/index.ts
```diff
@@ -220,29 +236,60 @@ class RequestHandler extends APIHandlerBase {
...
+
+    private matchUrlPattern(path: string, routeType: UrlPatterns): Match {
+        const pattern = this.urlPatternMap[routeType];
+        if (!pattern) {
+            throw new InvalidValueError(`Unknown route type: ${routeType}`);
+        }
+
+        const match = pattern.match(path);
+        if (match) {
+            match.type = this.modelNameMapping[match.type] ?? match.type;
```
ymc9: When a mapping is provided,  say "users" => "user", should access via unmapped path name like "POST /api/user" fail? I tend to think so - configuring a mapping requires you to only use the mapped name.

What do you think? Currently I believe both are valid.

packages/server/src/api/rest/index.ts
```diff
@@ -220,29 +237,60 @@ class RequestHandler extends APIHandlerBase {
...
+            match.relationship = this.modelNameMapping[match.relationship] ?? match.relationship;
+        }
```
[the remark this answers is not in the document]
ymc9: @lsmith77 I think @coderabbitai 's comment is valid. The "relationship" part in the URI represents model field names instead of model names, so they shouldn't be mapped.

E.g., if we have a mapped model name "myPost" => "post", we'll still be accessing "/api/user/relationships/posts" without altering the "posts" part, since it's a relation field name.

packages/server/src/api/rest/index.ts
```diff
@@ -531,11 +574,13 @@ class RequestHandler extends APIHandlerBase {
         }
 
         if (entity?.[relationship]) {
+            const mappedType = this.reverseModelNameMap(type);
+            const mappedRelationship = this.reverseModelNameMap(relationship);
```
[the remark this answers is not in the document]
ymc9: Same here @lsmith77

packages/server/src/api/rest/index.ts
```diff
@@ -582,11 +627,13 @@ class RequestHandler extends APIHandlerBase {
         }
 
         const entity: any = await prisma[type].findUnique(args);
+        const mappedType = this.reverseModelNameMap(type);
+        const mappedRelationship = this.reverseModelNameMap(relationship);
```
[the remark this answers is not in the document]
ymc9: And here

packages/server/src/api/rest/index.ts
```diff
@@ -50,6 +50,9 @@ export type Options = {
      * it should be included in the charset.
      */
     urlSegmentCharset?: string;
+
+    modelNameMapping?: Record<string, string>;
+    prefix?: string;
```
ymc9: Hi @lsmith77 , I think I'm still not understanding why this is necessary. Isn't configuring a prefix here (e.g., "/v1") equivalent to mounting the API handler directly under the "/v1" route at the framework level?

ymc9 (commented):
Hi @lsmith77, thanks for working on this!

I've put a few review comments there, please take a look when you have time. I've also pushed a sample test case that you can use as a base to expand.

ymc9:
> Hi @lsmith77, thanks for working on this!
> 
> I've put a few review comments there, please take a look when you have time. I've also pushed a sample test case that you can use as a base to expand.

It seems I can't push the test case commit directly. I've put it as a PR here: https://github.com/digital-sustainability/zenstack/pull/1

packages/server/src/api/rest/index.ts
lsmith77: Indeed. It should fail. We only had the case in mind where only some of the models are mapped but not all.

So should we rather throw an exception if a model is not mapped if there are any mappings?

packages/server/src/api/rest/index.ts
lsmith77: In our tests if I just add the `prefix` on the `endpoint`, we would get an error since the regular expressions expect the path to start with model name.

ie. it was trying to match `/api/v1/users` to `/users`

packages/server/src/api/rest/index.ts
ymc9: I think supporting partial mapping is a good feature to have. What I meant is that, for the models that are configured with a mapping, the original name shouldn't be accessible anymore. E.g., if model "user" is mapped to "users", only "/api/users/..." should be valid, and "/api/user/..." should not.

lsmith77: Ok. I guess we will check if the type is included in the `reverseModelNameMapping`. In that case I would return `null` so that it is handled as no match found.

packages/server/src/api/rest/index.ts
ymc9: I see. The server adapter that wraps around the api handler is supposed to strip the url prefix (either via configuration or automatically) and pass the clean result to the api handler.

Which server adapter are you using? Is it NestJS? Have you tried the `baseUrl` option? https://zenstack.dev/docs/reference/server-adapters/nestjs#parameter-options-1

lsmith77: ah indeed, `baseUrl` does the trick. so I will remove the `prefix`. it is a bit unfortunate that in the openapi plugin it is called `prefix` and here it is called `baseUrl`. maybe something to consider for v3

packages/server/src/api/rest/index.ts
lsmith77: I now did some testing. At least with Nest.js it already returns a 404, since there is no `/user` route defined. Now if someone would then also define a `/user` route, indeed without change it would then work but at that point that seems to be intended by the user. So imho there is no need to handle this case inside the `RequestHandler`.

packages/server/src/api/rest/index.ts
lsmith77: FYI the `baseUrl` isn't added to the `endpoint`, so it means you need to add this prefix both as a `baseUrl` option and also add it to the `endpoint`. But I will leave it as is for now. But IMHO it would be better to clean this up, so that it only needs to be set in one location (in addition to also setting in the openapi plugin as well)

lsmith77:
> > Hi @lsmith77, thanks for working on this!
> > I've put a few review comments there, please take a look when you have time. I've also pushed a sample test case that you can use as a base to expand.
> 
> It seems I can't push the test case commit directly. I've put it as a PR here: [digital-sustainability#1](https://github.com/digital-sustainability/zenstack/pull/1)

I added one comment. Shouldn't using the unmapped model name cause a 404 and not a 400?

packages/server/src/api/rest/index.ts
ymc9: Yes, it's a bit awkward now. The sole purpose of `endpoint` is to determine how to generate links in the returned document. Today the server adapters and api handlers are agnostic to each other. We can seek ways to improve in the future.

ymc9:
> > > Hi @lsmith77, thanks for working on this!
> > > I've put a few review comments there, please take a look when you have time. I've also pushed a sample test case that you can use as a base to expand.
> > 
> > 
> > It seems I can't push the test case commit directly. I've put it as a PR here: [digital-sustainability#1](https://github.com/digital-sustainability/zenstack/pull/1)
> 
> I added one comment. Shouldn't using the unmapped model name cause a 404 and not a 400?

I chose 400 originally based on the understanding that you're passing an "invalid" endpoint (not a valid but not-available one). On the hindsight, maybe 404 is better suited, but it'll be an unnecessary breaking change ...

lsmith77:
@ymc9 I have now also updated the openapi plugin. As part of this I realized it would make more sense to have the model mapping look like this:
```
{
  user: 'users'
}
```

rather than

```
{
  users: 'user'
}
```

packages/plugins/openapi/src/rest-generator.ts
```diff
@@ -126,6 +129,10 @@ export class RESTfulOpenAPIGenerator extends OpenAPIGeneratorBase {
         return result;
     }
 
+    private mapModelName(modelName: string): string {
+        return this.modelNameMapping[modelName] ?? modelName;
```
lsmith77, line 133: @ymc9 should we lowercase the `modelName` here?

packages/server/src/api/rest/index.ts
ymc9: My understanding is the mapping (in the latest code) is inverted.

I tested the latest code with this mapping:

```ts
{
    user: 'myUser',
    post: 'myPost',
}
```

And this test is failing because of returning 200:
```ts
            await expect(
                handler({
                    method: 'post',
                    path: '/user',
                    query: {},
                    requestBody: { data: { type: 'user', attributes: { name: 'User1' } } },
                    prisma,
                })
            ).resolves.toMatchObject({
                status: 400,
            });
```

packages/plugins/openapi/src/rest-generator.ts
ymc9: I think it depends on how we want the user to configure the mapping.

I feel it can be quite confusing what the mapping actually means. Should we make it more explicit and non-ambiguous, and let it represent a mapping from model name (without any casing changes) to resource (url path) name?

```ts
// left side is model name as defined in zmodel
// right side is rest resource name as represented in url and payload
modelResourceNameMap: {
    User: 'users',
    Post: 'posts'
}
```

So that the direction of mapping and the meaning of each side is very obvious and clear?

ymc9: And these should be consistent for the rest api handler and the open api plugin

lsmith77: ok agreed. in both places we now map from the model name to the external name represented in the API urls.

lsmith77:
@ymc9 is there anything else you need from us before you can merge this?

packages/server/src/api/rest/index.ts
ymc9: We can get it merged first and I'll make a follow up PR, with some more test cases too.

ymc9:
> @ymc9 is there anything else you need from us before you can merge this?

Hi @lsmith77, it seems the newly added test case is still failing. Also I believe the mapping direction in the test case should be updated based on the latest changes. Could you help fix that, then I think we can merge it. I'll make a follow up PR to add more test cases.

lsmith77:
@ymc9 

> > @ymc9 is there anything else you need from us before you can merge this?
> 
> Hi @lsmith77, it seems the newly added test case is still failing. Also I believe the mapping direction in the test case should be updated based on the latest changes. Could you help fix that, then I think we can merge it. I'll make a follow up PR to add more test cases.

Ah yes, sorry. I didn't take ownership of the test within the PR. I have updated the mapping. I have also now added explicit code to check for the old path (which IMHO should be done by the framework's router and not really here).

But still I am unable to get the tests to pass. Maybe I am running the tests incorrectly.

What I did was:

```
pnpm build
cd packages/server
pnpm test rest.test.ts -- --testNamePattern="REST server tests - model name mapping"
```

Which results in:
```
   - Expected  - 1
    + Received  + 1

      Object {
    -   "status": 400,
    +   "status": 201,
      }

      3057 |                     prisma,
      3058 |                 })
    > 3059 |             ).resolves.toMatchObject({
           |                        ^
      3060 |                 status: 400,
      3061 |             });
      3062 |

      at Object.toMatchObject (../../node_modules/expect/build/index.js:174:22)
      at Object.<anonymous> (tests/api/rest.test.ts:3059:24)
```

ymc9:
> @ymc9
> 
> > > @ymc9 is there anything else you need from us before you can merge this?
> > 
> > 
> > Hi @lsmith77, it seems the newly added test case is still failing. Also I believe the mapping direction in the test case should be updated based on the latest changes. Could you help fix that, then I think we can merge it. I'll make a follow up PR to add more test cases.
> 
> Ah yes, sorry. I didn't take ownership of the test within the PR. I have updated the mapping. I have also now added explicit code to check for the old path (which IMHO should be done by the framework's router and not really here).
> 
> But still I am unable to get the tests to pass. Maybe I am running the tests incorrectly.
> 
> What I did was:
> 
> ```
> pnpm build
> cd packages/server
> pnpm test rest.test.ts -- --testNamePattern="REST server tests - model name mapping"
> ```
> 
> Which results in:
> 
> ```
>    - Expected  - 1
>     + Received  + 1
> 
>       Object {
>     -   "status": 400,
>     +   "status": 201,
>       }
> 
>       3057 |                     prisma,
>       3058 |                 })
>     > 3059 |             ).resolves.toMatchObject({
>            |                        ^
>       3060 |                 status: 400,
>       3061 |             });
>       3062 |
> 
>       at Object.toMatchObject (../../node_modules/expect/build/index.js:174:22)
>       at Object.<anonymous> (tests/api/rest.test.ts:3059:24)
> ```

That's what I meant in a previous comment: even with the mapping "user" -> "myUser", access to route "/user" is still valid given the current implementation, which should actually fail. If a model name is mapped, only the mapped route is accessible. What do you think?

lsmith77:
This aspect I do not understand.

1) There should not be a `/user` route defined within the framework. So the framework should already return a 404. This is at least how it is in Nest.js
2) I have now even added explicit code, which should result in no match being found.

Now for 2), I guess maybe that logic is not sufficient and rather it should trigger an error response directly.

But due to 1), I am not even convinced this case should even be handled explicitly at all.

ymc9:
> This aspect I do not understand.
> 
> 1. There should not be a `/user` route defined within the framework. So the framework should already return a 404. This is at least how it is in Nest.js
> 2. I have now even added explicit code, which should result in no match being found.
> 
> Now for 2), I guess maybe that logic is not sufficient and rather it should trigger an error response directly.
> 
> But due to 1), I am not even convinced this case should even be handled explicitly at all.

I think I understand where the gap is 😄.

The rest api handler is framework agnostic and is supposed to work with different frameworks. For Nest.js, people would explicitly define "myUser" route, however for other frameworks, e.g., Express.js, one would simply mount the api handler to a parent route and let it handle whatever request is sent to it:

```ts
// forward all route under "/api/model" to the underlying request handler
app.use(
    '/api/model',
    ZenStackMiddleware({
        ...
    })
);
```

That's why I think the api handler itself should have a consistent handling of mapped model names, regardless the outer framework rejects an invalid route or not. I.e., if "user" is mapped to "myUser", then "/user" route shouldn't handle any request anymore, at the api handler level.

ymc9:
> > This aspect I do not understand.
> > 
> > 1. There should not be a `/user` route defined within the framework. So the framework should already return a 404. This is at least how it is in Nest.js
> > 2. I have now even added explicit code, which should result in no match being found.
> > 
> > Now for 2), I guess maybe that logic is not sufficient and rather it should trigger an error response directly.
> > But due to 1), I am not even convinced this case should even be handled explicitly at all.
> 
> I think I understand where the gap is 😄.
> 
> The rest api handler is framework agnostic and is supposed to work with different frameworks. For Nest.js, people would explicitly define "users" route, however for other frameworks, e.g., Express.js, one would simply mount the api handler to a parent route and let it handle whatever request sent to it:
> 
> ```ts
> // forward all route under "/api/model" to the underlying request handler
> app.use(
>     '/api/model',
>     ZenStackMiddleware({
>         ...
>     })
> );
> ```
> 
> That's why I think the api handler itself should have a consistent handling of mapped model names, regardless the outer framework rejects an invalid route or not.

The relationship between framework and api handler is depicted here:

https://zenstack.dev/docs/reference/server-adapters/api-handlers/

lsmith77:
OK, I am now throwing an exception in this case and the tests are now passing.

ymc9 (approved):
Thanks @lsmith77 ! I'm merging the change and will publish a release soon.

PR #2188, opened 2025-07-17
feat: support natural keys in the API

lsmith77:
this makes it possible to use a natural key even while also using surrogate primary keys

https://www.prisma.io/dataguide/datamodeling/correctness-constraints#the-primary-key

Using the surrogate key might be sensible since it is more compact for foreign keys. Also the tag name might be editable. But inside the API it makes no sense to use the surrogate key.

ie. `/tags/1` makes no sense. `/tags/blue` is the much better developer experience.

In order to make use of this a unique key needs to be defined and referenced in the RESTApiHandler Options:

```
RESTApiHandler({
      ...
      externalIdMapping: {Post: source__post_id},
    });
  }
```

ie. in the model

```
model Post {
    id Int @id @default(autoincrement()) @deny('all', true)
    post_id Int
    source String
    title String?
    @@unique([source, post_id], name: "source__post_id")
}
```

This will result in a compound id in the API based on the fields in the unique index similar to when using a [compound `@id`](https://zenstack.dev/docs/reference/server-adapters/api-handlers/rest#compound-id-fields).

Note as shown in the above example, if the surrogate key should be hidden in the API response, it is possible to do this via `@deny('all', true)`

### DIFF
--- a/packages/server/src/api/rest/index.ts
+++ b/packages/server/src/api/rest/index.ts
@@ -7,6 +7,7 @@ import {
     PrismaErrorCode,
     clone,
     enumerate,
+    requireField,
     getIdFields,
     isPrismaClientKnownRequestError,
 } from '@zenstackhq/runtime';
@@ -52,6 +53,8 @@ export type Options = {
     urlSegmentCharset?: string;
 
     modelNameMapping?: Record<string, string>;
+
+    externalIdMapping?: Record<string, string>;
 };
 
 type RelationshipInfo = {
@@ -238,6 +241,7 @@ class RequestHandler extends APIHandlerBase {
     private urlPatternMap: Record<UrlPatterns, UrlPattern>;
     private modelNameMapping: Record<string, string>;
     private reverseModelNameMapping: Record<string, string>;
+    private externalIdMapping: Record<string, string>;
 
     constructor(private readonly options: Options) {
         super();
@@ -251,6 +255,12 @@ class RequestHandler extends APIHandlerBase {
         this.reverseModelNameMapping = Object.fromEntries(
             Object.entries(this.modelNameMapping).map(([k, v]) => [v, k])
         );
+
+        this.externalIdMapping = options.externalIdMapping ?? {};
+        this.externalIdMapping = Object.fromEntries(
+            Object.entries(this.externalIdMapping).map(([k, v]) => [lowerCaseFirst(k), v])
+        );
+
         this.urlPatternMap = this.buildUrlPatternMap(segmentCharset);
     }
 
@@ -1166,11 +1176,28 @@ class RequestHandler extends APIHandlerBase {
     }
 
     //#region utilities
+    private getIdFields(modelMeta: ModelMeta, model: string): FieldInfo[] {
+        const modelLower = lowerCaseFirst(model);
+        if (!(modelLower in this.externalIdMapping)) {
+            return getIdFields(modelMeta, model);
+        }
+
+        const metaData = modelMeta.models[modelLower] ?? {};
+        const externalIdName = this.externalIdMapping[modelLower];
+        const uniqueConstraints = metaData.uniqueConstraints ?? {};
+        for (const [name, constraint] of Object.entries(uniqueConstraints)) {
+            if (name === externalIdName) {
+                return constraint.fields.map((f) => requireField(modelMeta, model, f));
+            }
+        }
+
+        throw new Error(`Model ${model} does not have unique key ${externalIdName}`);
+    }
 
     private buildTypeMap(logger: LoggerConfig | undefined, modelMeta: ModelMeta): void {
         this.typeMap = {};
         for (const [model, { fields }] of Object.entries(modelMeta.models)) {
-            const idFields = getIdFields(modelMeta, model);
+            const idFields = this.getIdFields(modelMeta, model);
             if (idFields.length === 0) {
                 logWarning(logger, `Not including model ${model} in the API because it has no ID field`);
                 continue;
@@ -1186,7 +1213,7 @@ class RequestHandler extends APIHandlerBase {
                 if (!fieldInfo.isDataModel) {
                     continue;
                 }
-                const fieldTypeIdFields = getIdFields(modelMeta, fieldInfo.type);
+                const fieldTypeIdFields = this.getIdFields(modelMeta, fieldInfo.type);
                 if (fieldTypeIdFields.length === 0) {
                     logWarning(
                         logger,
@@ -1214,7 +1241,7 @@ class RequestHandler extends APIHandlerBase {
         const linkers: Record<string, Linker<any>> = {};
 
         for (const model of Object.keys(modelMeta.models)) {
-            const ids = getIdFields(modelMeta, model);
+            const ids = this.getIdFields(modelMeta, model);
             const mappedModel = this.mapModelName(model);
 
             if (ids.length < 1) {
@@ -1266,7 +1293,7 @@ class RequestHandler extends APIHandlerBase {
                 if (!fieldSerializer) {
                     continue;
                 }
-                const fieldIds = getIdFields(modelMeta, fieldMeta.type);
+                const fieldIds = this.getIdFields(modelMeta, fieldMeta.type);
                 if (fieldIds.length > 0) {
                     const mappedModel = this.mapModelName(model);
 
@@ -1306,7 +1333,7 @@ class RequestHandler extends APIHandlerBase {
         if (!data) {
             return undefined;
         }
-        const ids = getIdFields(modelMeta, model);
+        const ids = this.getIdFields(modelMeta, model);
         if (ids.length === 0) {
             return undefined;
         } else {
--- a/packages/server/tests/api/rest.test.ts
+++ b/packages/server/tests/api/rest.test.ts
@@ -3111,4 +3111,87 @@ describe('REST server tests', () => {
             });
         });
     });
+
+    describe('REST server tests - external id mapping', () => {
+        const schema = `
+    model User {
+        id Int @id @default(autoincrement())
+        name String
+        source String
+        posts Post[]
+
+        @@unique([name, source])
+    }
+
+    model Post {
+        id Int @id @default(autoincrement())
+        title String
+        author User? @relation(fields: [authorId], references: [id])
+        authorId Int?
+    }
+    `;
+        beforeAll(async () => {
+            const params = await loadSchema(schema);
+            prisma = params.prisma;
+            zodSchemas = params.zodSchemas;
+            modelMeta = params.modelMeta;
+
+            const _handler = makeHandler({
+                endpoint: 'http://localhost/api',
+                externalIdMapping: {
+                    User: 'name_source',
+                },
+            });
+            handler = (args) =>
+                _handler({ ...args, zodSchemas, modelMeta, url: new URL(`http://localhost/${args.path}`) });
+        });
+
+        it('works with id mapping', async () => {
+            await prisma.user.create({
+                data: { id: 1, name: 'User1', source: 'a' },
+            });
+
+            // user is no longer exposed using the `id` field
+            let r = await handler({
+                method: 'get',
+                path: '/user/1',
+                query: {},
+                prisma,
+            });
+
+            expect(r.status).toBe(400);
+
+            // user is exposed using the fields from the `name__source` multi-column unique index
+            r = await handler({
+                method: 'get',
+                path: '/user/User1_a',
+                query: {},
+                prisma,
+            });
+
+            expect(r.status).toBe(200);
+            expect(r.body.data.attributes.source).toBe('a');
+            expect(r.body.data.attributes.name).toBe('User1');
+
+            await prisma.post.create({
+                data: { id: 1, title: 'Title1', authorId: 1 },
+            });
+
+            // post is exposed using the `id` field
+            r = await handler({
+                method: 'get',
+                path: '/post/1',
+                query: { include: 'author' },
+                prisma,
+            });
+
+            expect(r.status).toBe(200);
+            expect(r.body.data.attributes.title).toBe('Title1');
+            // Verify author relationship contains the external ID
+            expect(r.body.data.relationships.author.data).toMatchObject({
+                type: 'user',
+                id: 'User1_a',
+            });
+        });
+    });
 });


lsmith77:
I can certainly refactor it to be a Handler option like the model name mapping. That being said, the argument you bring forth to some degree also applies to the `@@allow`/`@@deny` annotations, that do not really apply to data modelling but rather to the API only.

As for changing the semantics of `getIdFields()`. That is very hard for me to ascertain the risks involved given that I still have a very limited overview of the codebase.

Your proposal would rather be we do not call this method but rather receive a list of fields from the Handler config, which is then only checked at runtime against the defined fields (in order to map the model attributes to RDBMS field names to handle if the model used `@map`)?

I guess this would end up duplicating a fair bit of code which comes with its own set of risks. The advantage is of course it should prevent unwanted side-effects and whatever bugs it introduces would be isolated to this feature. If you feel more comfortable with this approach I am happy to implement it.

lsmith77:
FTR I just tried to quickly implement this but it appears that it is not enough to override all calls to `getIdFields()` in the Rest Handler. In fact just using GET calls, the above changes seem to have no impact at all:
```
diff --git a/packages/server/src/api/rest/index.ts b/packages/server/src/api/rest/index.ts
index 75196b1a..8c5b4461 100644
--- a/packages/server/src/api/rest/index.ts
+++ b/packages/server/src/api/rest/index.ts
@@ -8,6 +8,7 @@ import {
     clone,
     enumerate,
     getIdFields,
+    requireField,
     isPrismaClientKnownRequestError,
 } from '@zenstackhq/runtime';
 import { lowerCaseFirst, upperCaseFirst, paramCase } from '@zenstackhq/runtime/local-helpers';
@@ -52,6 +53,8 @@ export type Options = {
     urlSegmentCharset?: string;
 
     modelNameMapping?: Record<string, string>;
+
+    externalIdMapping: Record<string, string>;
 };
 
 type RelationshipInfo = {
@@ -238,6 +241,7 @@ class RequestHandler extends APIHandlerBase {
     private urlPatternMap: Record<UrlPatterns, UrlPattern>;
     private modelNameMapping: Record<string, string>;
     private reverseModelNameMapping: Record<string, string>;
+    private externalIdMapping: Record<string, string>;
 
     constructor(private readonly options: Options) {
         super();
@@ -251,6 +255,7 @@ class RequestHandler extends APIHandlerBase {
         this.reverseModelNameMapping = Object.fromEntries(
             Object.entries(this.modelNameMapping).map(([k, v]) => [v, k])
         );
+        this.externalIdMapping = options.externalIdMapping ?? {};
         this.urlPatternMap = this.buildUrlPatternMap(segmentCharset);
     }
 
@@ -1166,11 +1171,27 @@ class RequestHandler extends APIHandlerBase {
     }
 
     //#region utilities
+    private getIdFields(modelMeta: ModelMeta, model: string): FieldInfo[] {
+        if (!(model in this.externalIdMapping)) {
+            return getIdFields(modelMeta, model);
+        }
+
+        const metaData = modelMeta.models[lowerCaseFirst(model)] ?? {};
+        const uniqueConstraints = metaData.uniqueConstraints ?? {};
+        const externalIdName = this.externalIdMapping[model];
+        for (const [name, constraint] of Object.entries(uniqueConstraints)) {
+            if (name === externalIdName) {
+                return constraint.fields.map((f) => requireField(modelMeta, model, f));
+            }
+        }
+
+        return [];
+    }
 
     private buildTypeMap(logger: LoggerConfig | undefined, modelMeta: ModelMeta): void {
         this.typeMap = {};
         for (const [model, { fields }] of Object.entries(modelMeta.models)) {
-            const idFields = getIdFields(modelMeta, model);
+            const idFields = this.getIdFields(modelMeta, model);
             if (idFields.length === 0) {
                 logWarning(logger, `Not including model ${model} in the API because it has no ID field`);
                 continue;
@@ -1186,7 +1207,7 @@ class RequestHandler extends APIHandlerBase {
                 if (!fieldInfo.isDataModel) {
                     continue;
                 }
-                const fieldTypeIdFields = getIdFields(modelMeta, fieldInfo.type);
+                const fieldTypeIdFields = this.getIdFields(modelMeta, fieldInfo.type);
                 if (fieldTypeIdFields.length === 0) {
                     logWarning(
                         logger,
@@ -1214,7 +1235,7 @@ class RequestHandler extends APIHandlerBase {
         const linkers: Record<string, Linker<any>> = {};
 
         for (const model of Object.keys(modelMeta.models)) {
-            const ids = getIdFields(modelMeta, model);
+            const ids = this.getIdFields(modelMeta, model);
             const mappedModel = this.mapModelName(model);
 
             if (ids.length < 1) {
@@ -1266,7 +1287,7 @@ class RequestHandler extends APIHandlerBase {
                 if (!fieldSerializer) {
                     continue;
                 }
-                const fieldIds = getIdFields(modelMeta, fieldMeta.type);
+                const fieldIds = this.getIdFields(modelMeta, fieldMeta.type);
                 if (fieldIds.length > 0) {
                     const mappedModel = this.mapModelName(model);
 
@@ -1306,7 +1327,7 @@ class RequestHandler extends APIHandlerBase {
         if (!data) {
             return undefined;
         }
-        const ids = getIdFields(modelMeta, model);
+        const ids = this.getIdFields(modelMeta, model);
         if (ids.length === 0) {
             return undefined;
         } else {
```

So it seems like further investigation is needed where all getIdFields() needs to be overridden to get the desired change in behavior. Which however then might lead to additional complexity introduced in even more places :-/

ymc9:
> Author

Hi @lsmith77 , sorry I didn't put it clearly in my previous comments. What I meant was `@@externalId`'s effect is only at the HTTP API layer (specifically about the REST flavor API). On the contrary, `@@allow` and `@@deny` are built-in features at the data access layer. The enhanced PrismaClient recognizes these rules and uses them to enforce access control, regardless the access originates from hand-coded ORM calls or requests to the automatic HTTP CRUD endpoints.

The rules are effective in  manually written backend code:
```ts
const db = enhance(prisma, { user: getAuthUser() });
// only posts allowed to be read are returned
const posts = await db.post.findMany();
```

, and when requested from automatic http api:

```ts
// requesting "/api/model/post" will only return readable posts
app.use(
    '/api/model',
    ZenStackMiddleware({
        handler: RestApiHandler(...),
        getPrisma: (request) => enhance(prisma, { user: getAuthUser(request) }),
    })
);
```

lsmith77:
OK thx for this additional explanation.
I am off this week but will look into this further next week but as my above explanation shows, it will require changes beyond the Rest Handler to get this to work. In ligt of this, are you still open to such a change?

ymc9:
> OK thx for this additional explanation. I am off this week but will look into this further next week but as my above explanation shows, it will require changes beyond the Rest Handler to get this to work. In ligt of this, are you still open to such a change?

Yes, an option in the rest handler sounds good to me.

lsmith77:
@ymc9 ok got it to work. the issue came from `model` sometimes not having `lowerCaseFirst` applied already.

the second commit is not necessary but felt like a sensible optimization but I am happy to revert if you prefer.

lsmith77:
I will now look into updating the openapi plugin so that `filter[id]` sets the appropriate string type where necessary. Or should I rather just hide this filter?

Actually upon further investigation it appears that right now you just list all the compound id fields as separate filters, so I guess there isn't really anything to do here.

We will likely just add our own logic to remove `filter[id]` entirely since we do not want people to be able to interact with the surrogate key in our API. But I guess this will remain custom code.

lsmith77:
Hmm we would however like to not include the `id` in the output. But I cannot use `@ignore` on the field unless we change all our foreign keys to use the natural keys as well, which we obviously do not want to do.

What would you advise here? How could be prevent exposing `id` while still being able to use it as a surrogate key in the model?

```
model Project {
    id                       Int                   @id @default(autoincrement())
    ..
```

image

lsmith77:
ok found a solution using `@deny`:
```
id                       Int                   @id @default(autoincrement()) @deny('all', true)
```

lsmith77:
@ymc9 could you have another look if this is now the right direction and what is left to do to get this merged?

packages/server/src/api/rest/index.ts
```diff
@@ -1166,11 +1176,33 @@ class RequestHandler extends APIHandlerBase {
     }
 
     //#region utilities
+    private getIdFields(modelMeta: ModelMeta, model: string): FieldInfo[] {
+        const modelLower = lowerCaseFirst(model);
+
+        if (this.typeMap && modelLower in this.typeMap) {
```
lsmith77: this is a performance optimization but not sure if its really needed

ymc9: I believe after typeMap is built, every model with proper id fields will have an entry in it. Will this code block then completely mute the "externalIdMapping" check below?

packages/server/src/api/rest/index.ts
```diff
@@ -1166,11 +1176,33 @@ class RequestHandler extends APIHandlerBase {
...
+        }
+
+        if (!(modelLower in this.externalIdMapping)) {
+            return getIdFields(modelMeta, model);
+        }
+
+        const metaData = modelMeta.models[modelLower] ?? {};
+        const externalIdName = this.externalIdMapping[modelLower];
+        const uniqueConstraints = metaData.uniqueConstraints ?? {};
+        for (const [name, constraint] of Object.entries(uniqueConstraints)) {
```
ymc9, line 1188: There's some complexity here due to how Prisma handles multi-field id fields or unique fields.

If a model has id like:

```prisma
model Model {
  ...
  @@id([x, y])
}
```

The model meta's uniqueConstraint for this model will have an entry like:
```ts
{
  ...
  x_y: { name: 'x_y', fields: ['x', 'y']
}
```

, and when querying, the synthetic compound field "x_y" must be used as a unique filter.

```ts
db.model.findUnique({ where: { x_y: { x: 1, y: 2 } });
```

For this new feature, do you plan to support mapping multi-field id/unique fields? That'll result in some more complexity. However, if not, we can have a check in the loop and only match unique constraint that has a single field.

I don't personally see too much value to support multi field.

packages/server/src/api/rest/index.ts
```diff
@@ -251,6 +255,12 @@ class RequestHandler extends APIHandlerBase {
         this.reverseModelNameMapping = Object.fromEntries(
             Object.entries(this.modelNameMapping).map(([k, v]) => [v, k])
         );
+
+        this.externalIdMapping = options.externalIdMapping ?? {};
```
ymc9, line 259: Could you also help add a test case for this new feature in "packages/server/tests/api/rest.test.ts"? Thank you!

ymc9:
Hi @lsmith77 , I've left a few review comments. Please take a look when you have time. Thanks!

packages/server/src/api/rest/index.ts
lsmith77: For `@id` fields, this feature is not needed, since the current code already handles this. so I am only supporting model (not field level) `@unique`. I have added an example at the very top of this PR.

packages/server/src/api/rest/index.ts
lsmith77: Yes that was the idea as my assumption was that there is no way things could change at run time.

packages/server/src/api/rest/index.ts
lsmith77: I have added a test, though tests seem to be failing but not sure if this is related to my new test :-/

lsmith77: something is weird. I am investigating.

packages/server/tests/api/rest.test.ts
```diff
@@ -3111,4 +3111,87 @@ describe('REST server tests', () => {
...
+
+            // user is no longer exposed using the `id` field
+            let r = await handler({
+                method: 'get',
+                path: '/user/1',
+                query: {},
+                prisma,
+            });
+
+            expect(r.status).toBe(400);
```
lsmith77, line 3162: this here imho should be a 404 but it causes a 400 due to `"detail": "\nInvalid `prisma.post.findUnique()` invocation` as it expects a `_` separated string and therefore lacks the number of parameters for this method call.

that being said, the same happens when using the existing compound id feature. so this issue be addressed in another PR, though I am not entirely sure where. if you give me a hint I can look into this in another PR.

packages/server/src/api/rest/index.ts
lsmith77: OK, there is still an issue here.

If I map the unique index name to something arbitrary, it doesn't work as the current code calling `findUnqiue()` apparently always then uses the `fieldA_fieldB` in the `where` rather than the unique index name, causing an error.

But digging into the source code I found that this is a known limitation:
https://github.com/zenstackhq/zenstack/blob/main/packages/server/src/api/rest/index.ts#L1415

packages/server/src/api/rest/index.ts
lsmith77: ok, locally the tests are now passing.

ymc9:
Hi @lsmith77 , could you merge from the latest code from the branch? I've fixed some environment issues that were causing test failures.

packages/server/src/api/rest/index.ts
lsmith77: After some more reflection, this change does not prevent the computation being done multiple times for the same model if a schema contains relations. As such I guess I will remove it since it only seems to affect the very first call anyway.

A proper approach would be using LRU cache on the method call but again not sure if its even necessary.

lsmith77:
@ymc9 ok looks better now!

3 topics remain:
1) performance optimization in `getIdFields()` [should be reverted](https://github.com/zenstackhq/zenstack/pull/2188#discussion_r2259228436)
2) only unique indexes without a custom name work -> out of scope as its a [known limitation](https://github.com/zenstackhq/zenstack/pull/2188#discussion_r2257085757)
3) passing an ID that doesn't include the appropriate number of "chunks" causes a [400 and not 404](https://github.com/zenstackhq/zenstack/pull/2188#discussion_r2257053650) -> out of scope but should be fixed in another PR

I have taken care of 1) and if you provide a hint for 3) I will work on this in a followup PR.

lsmith77:
Ok figured out 3)

First up, it seems like you consider an invalid resource ID to be a 400 anyway, so maybe there is nothing to fix here:
```
            invalidId: {
                status: 400,
                title: 'Resource ID is invalid',
            },
```

I see two possible fixes.

1) Look for `undefined` for any field in `where` of the `args`.

2) Catch the `PrismaClientValidationError` exception

```
diff --git a/packages/server/src/api/rest/index.ts b/packages/server/src/api/rest/index.ts
index 0bf02762..c4d2fb64 100644
--- a/packages/server/src/api/rest/index.ts
+++ b/packages/server/src/api/rest/index.ts
@@ -521,16 +521,24 @@ class RequestHandler extends APIHandlerBase {
             include = allIncludes;
         }
 
-        const entity = await prisma[type].findUnique(args);
+        try {
+            const entity = await prisma[type].findUnique(args);
 
-        if (entity) {
+            if (entity) {
+                return {
+                    status: 200,
+                    body: await this.serializeItems(type, entity, { include }),
+                };
+            }
+            // how to only catch PrismaClientValidationError ?
+        } catch (e) {
             return {
-                status: 200,
-                body: await this.serializeItems(type, entity, { include }),
+                status: 404,
+                body: 'notFound due to invalid resource ID',
             };
-        } else {
-            return this.makeError('notFound');
         }
+
+        return this.makeError('notFound');
     }
 
     private async processFetchRelated(
```

packages/server/tests/api/rest.test.ts
ymc9: Yes, I agree it should be a 404. The current implementation of the rest handler doesn't do strict input validation so some of the invalid input leaks to Prisma. We can leave a fix to the future.

packages/server/src/api/rest/index.ts
ymc9: You're right. Custom index name is not respected yet. I think we can also leave it for the future if it doesn't hurt you now.

ymc9: If you configure the external id mapping and the target constraint has a custom index name, I guess you can just configure the mapping directly to that cusotm name, right?

lsmith77: Unfortunately the current code for compound IDs only receives the field names, not the name of the constraint. It then assumes the constraint name is simply a joined with underscores. So in order to make this all work, in addition to the field names, we need to keep track of the constraint name. But that should be done in a follow up PR IMHO.

packages/server/tests/api/rest.test.ts
lsmith77: see here for two possible approaches to fix this:
https://github.com/zenstackhq/zenstack/pull/2188#issuecomment-3163705895

note the issue also affects the current compound ID feature. so this is an existing issue, not one introduced with this PR.

ymc9:
> Resource ID is invalid

I searched for references and it seems the "invalidId" error is never used ... not sure why but maybe the usage was removed with a previous refactor.

If the reason comes from Prisma rejecting an invalid unique filter, I believe the fix is larger, because a real fix should really be systematically validate the entire input object that's passed to Prisma.

I suggest we can get this PR merged and worry about the larger fix when it hurts.

Btw, do you want to work on the documentation for this new feature too? It'd be great if you have time!

lsmith77:
Ok great. Yeah I will add a doc PR next once this is merged, ok?

ymc9:
> Ok great. Yeah I will add a doc PR next once this is merged, ok?

Sounds good. Thank you! The related doc is here: https://github.com/zenstackhq/zenstack-docs/blob/main/docs/reference/server-adapters/api-handlers/rest.mdx

I'm merging this PR @lsmith77

lsmith77:
I have created https://github.com/zenstackhq/zenstack/issues/2211

lsmith77:
here is the doc PR https://github.com/zenstackhq/zenstack-docs/pull/480

Issue #1959, opened 2025-01-13
[Feature Request] Encapsulation and Reusability of Permission Rules in ZModel


alex-zwingli:
Currently, when defining permissions in ZModel, it is challenging to encapsulate and reuse permission conditions. For example, repeating complex permission conditions like `auth().memberships?[permissions?[resource == currentModel() && action == currentOperation()]]` across models or operations can lead to redundancy and reduced maintainability. Additionally, there is no straightforward way to ensure operations like `create`, `update`, and `delete` are allowed only when `read` is permitted.

1. Introduce a "function-like" construct in ZModel, such as `rule`, to encapsulate reusable permission conditions. For example:  
   ```zmodel
   rule hasResourceActionPermission() {
     auth().memberships?[permissions?[resource == currentModel() && action == currentOperation()]]
   }
   ```
   This would allow defining reusable rules that can be called across models and operations.

2. Support parameters for rules to make them flexible and applicable to specific model types or conditions. For example:
   ```zmodel
   rule hasPermissionForAction(action) {
     auth().memberships?[permissions?[resource == currentModel() && action == action]]
   }
   
3. Allow rules to ensure granular permissions for operations (e.g., `create`, `update`, `delete`) while enforcing dependencies between them, such as requiring read to be allowed.

**Alternatives considered**
A current workaround is to define permissions as if `read` is allowed and then use `@@deny` with the inverse policy of read to deny actions when `read` is not permitted. However, this approach is not intuitive and can become cumbersome in complex models.

**Additional context**
This feature would significantly improve maintainability, reduce redundancy, and simplify permission management in ZModel. For example, using rules like the one below would enable concise and reusable permission logic:
```zmodel
@@allow('create,update,delete', check(self(), 'read') && hasResourceActionPermission())
```

zgeorge22:
@ymc9 @jiashengguo  Any update on this (ideally for V2 as well, but at least for V3)? Would make writing complex access policies much easier and more readable.

PR #2196, opened 2025-07-20
feat: enable alias declarations

Azzerty23:
This is a first attempt at implementing **`alias` declarations**. The goal is to enable code reuse for both policy rules and default functions.

Example of intended usage:

```zmodel
alias allowAll() {
    true
}

alias defaultTitle() {
    'Default Title'
}

alias currentUser() {
    auth().id 
}

alias ownPublishedPosts() {
    currentUser() == authorId && published
}

model Post {
    id          Int      @id @default(autoincrement())
    title       String   @default(defaultTitle())
    published   Boolean  @default(allowAll())

    author      User     @relation(fields: [authorId], references: [id])
    authorId    String   @default(auth().id)

    @@allow('read', allowAll())
    @@allow('create,update,delete', currentUser() == authorId && published)
    @@deny('all', !ownPublishedPosts())
}

model User {
    id          String    @id @default(cuid())
    name        String?
    posts       Post[]

    @@allow('all', allowAll())
}

```

However, during implementation, I realized that having a *shared generic alias mechanism* for both policy rules and default values might not be the best idea: policy rule expressions and default value expressions are not interchangeable, and it becomes tricky to handle validation rules properly.

I initially started working on `rule` ([https://github.com/zenstackhq/zenstack/issues/1959](https://github.com/zenstackhq/zenstack/issues/1959)), but got sidetracked along the way... 😅

Also, I might have done something completely off with the SDL (Langium) changes — I have zero experience in that area.

Currently, I'm generating functions from aliases inside the policy file. But I’m thinking of reverting to my original approach of inlining the expression instead, as it would make handling relations easier. For example, this:

```ts
// generated policy.ts
function allowAll(): any {
  return true;
}

function Post_read(context: QueryContext, db: CrudContract): any {
  if (allowAll()) {
    return { AND: [] };
  }
  return { OR: [] };
}
```

Would become something like:

```ts
// generated policy.ts
function Post_read(context: QueryContext, db: CrudContract): any {
  if (true) {
    return { AND: [] };
  }
  return { OR: [] };
}
```

I’m also facing scoping issues where the expression doesn't allow referencing model fields:

```zmodel
alias ownPublishedPosts() {
    currentUser() == authorId && published
}
// Could not resolve reference to ReferenceTarget named 'authorId'.
// Could not resolve reference to ReferenceTarget named 'published'.
// invalid operand type for "&&" operator

```

@ymc9, thanks a lot in advance for your help!
If this approach doesn’t make sense at all, please don’t hesitate to say so ;)

### DIFF
--- a/packages/schema/src/language-server/validator/attribute-application-validator.ts
+++ b/packages/schema/src/language-server/validator/attribute-application-validator.ts
@@ -30,12 +30,7 @@ import {
 import { ValidationAcceptor, streamAllContents, streamAst } from 'langium';
 import pluralize from 'pluralize';
 import { AstValidator } from '../types';
-import {
-    getStringLiteral,
-    mapBuiltinTypeToExpressionType,
-    translateExpressionTypeToResolve,
-    typeAssignable,
-} from './utils';
+import { getStringLiteral, mapBuiltinTypeToExpressionType, typeAssignable } from './utils';
 
 // a registry of function handlers marked with @check
 const attributeCheckers = new Map<string, PropertyDescriptor>();
@@ -424,10 +419,8 @@ function isAliasAssignableToType(alias: AliasDecl, dstType: string, attr: Attrib
         return false;
     }
 
-    const mappedAliasResolvedType = translateExpressionTypeToResolve(alias.expression.$type);
-    return (
-        effectiveDstType === mappedAliasResolvedType || effectiveDstType === 'Any' || mappedAliasResolvedType === 'Any'
-    );
+    const aliasExpressionType = alias.expression.$resolvedType?.decl;
+    return effectiveDstType === aliasExpressionType || effectiveDstType === 'Any' || aliasExpressionType === 'Any';
 }
 
 function resolveEffectiveDestinationType(dstType: string, attr: AttributeApplication): string | null {
--- a/packages/schema/src/language-server/validator/utils.ts
+++ b/packages/schema/src/language-server/validator/utils.ts
@@ -99,26 +99,6 @@ export function mapBuiltinTypeToExpressionType(
     }
 }
 
-/**
- * Map an expression $type (e.g. StringLiteral) to a resolved ExpressionType (e.g. String)
- */
-export function translateExpressionTypeToResolve(expressionType: Expression['$type']): ExpressionType {
-    switch (expressionType) {
-        case 'StringLiteral':
-            return 'String';
-        case 'NumberLiteral':
-            return 'Int';
-        case 'BooleanLiteral':
-            return 'Boolean';
-        case 'ObjectExpr':
-            return 'Object';
-        case 'NullExpr':
-            return 'Null';
-        default:
-            return 'Any';
-    }
-}
-
 export function isAuthOrAuthMemberAccess(expr: Expression): boolean {
     return isAuthInvocation(expr) || (isMemberAccessExpr(expr) && isAuthOrAuthMemberAccess(expr.operand));
 }


Azzerty23:
It’s conceptually more of a fragment than an alias...
Regarding my reference issue, I understand that it's not possible to include an expression inside the body of the alias —references can't be resolved statically outside the scope of a model.
And if we use a string instead of an Expression in the body of the declaration, the references won’t be resolved in the abstract syntax tree. I’m wondering if this "fragment" idea can actually be implemented.

Azzerty23:
I partially resolve my issue by linking references inside alias to the first matching model field (this could be improved). Type resolution appears to be working well, and both Prisma model generation and policy rule generation are functioning as expected (tested on simple use cases). I will continue later this week. 

Before going further, I’d like to get the ZenStack team’s approval to make sure I’m not heading in the wrong direction. A few specific points I’d like to clarify:

- Which keyword should I keep (`rule`, `fragment`, etc.)?
- Should we keep this generic fragment system, or specialize it for policy rules (the original FR)? I feel it might make more sense to have a different keyword (if needed) to define stored values, e.g., `const myValue = ....`
- Should we keep the parentheses in declarations like `rule myRule() { ... }`? I’m not sure I fully understand the benefit of parameters, considering references are supposed to be resolved in the scope of the model where the rule is invoked.

ymc9:
Hi @Azzerty23 , thanks a lot for driving this effort. A sharing mechanism of policy rules is a long-desired feature.

I think we have two main things to discuss:

1. Is this mechanism more a "fragment" (its content literally copied over to the place where it's applied) or a "function" (with well-defined input and output)
2. The keyword to use

For #1, my tendency is to conceptualize it as a function, because the concept already exists today, and a function can have its input's types defined so we can keep the ZModel a statically typed language as it is today). So it's basically:

```ts
[keyword] isCurrentUser(user: User) {
  user.id == auth().id
}

...

model User {
  ...
  @@allow('all', isCurrentUser(this))
}

model Post {
  ...
  @@allow('all', isCurrentUser(author))
}
```

Love to hear your thoughts about this.

The benefit is that the resolution is very clean and straightforward. The downside is you can only call the function where the type matches. It doesn't apply to other models that also have an "id" field. My feeling is that this should generally be fine, thanks to abstract inheritance and polymorphism. But please also verify if it works for your use cases.

For #2, we can choose to use a new keyword like `rule`, etc., so it becomes explicit that this construct is specifically used for policy rules. Or we can just reuse the `function` keyword, keeping the option open that such a construct can potentially be used in other context as well (e.g., `@default`). My current tendency is to reuse `function` and avoid introducing a new concept.

Azzerty23:
Hi @ymc9, thank you for your response! While implementing this, I realized what I created was more like a fragment, but that wasn’t driven by any specific use case. I agree that a function, as you described, makes much more sense. Fragments feel too restrictive for code reusability since their references rely on having the same names. 

I also agree with your second point about reusing the existing `function` keyword instead of introducing something new. The type-safe aspect you mentioned is definitely important, so I’ll experiment with the function-based approach and share my findings.

One question: do you think it would make sense to support parameters with default values for extra flexibility? (maybe as a future enhancement)

ymc9:
> Hi @ymc9, thank you for your response! While implementing this, I realized what I created was more like a fragment, but that wasn’t driven by any specific use case. I agree that a function, as you described, makes much more sense. Fragments feel too restrictive for code reusability since their references rely on having the same names.
> 
> I also agree with your second point about reusing the existing `function` keyword instead of introducing something new. The type-safe aspect you mentioned is definitely important, so I’ll experiment with the function-based approach and share my findings.
> 
> One question: do you think it would make sense to support parameters with default values for extra flexibility? (maybe as a future enhancement)

Awesome! So we're very much on the same page. I suggest we keep parameter default out of the scope first to keep it simple.

Additionally, please note that a significant portion of the implementation will need to be rewritten in v3 due to the entirely different mechanism for access policies.

PR #2202, opened 2025-07-25
fix: Restore processing of prismaNamespace.ts

mhodgson:
For prisma-client generator there are some `delegate_aux_*` properties that, if not removed, cause type errors with the models since those fields have been removed from the model types. 
 
 Previous discussion here: https://github.com/zenstackhq/zenstack/pull/2177#issuecomment-3101979507
 
 cc/ @ymc9


ymc9 (approved):
Thanks, @mhodgson . I still couldn't figure out what triggers generation of such content in prismaNamespace.ts, but the change LGTM. I'm merging it and will publish a patch release.

PR #2206, opened 2025-08-04
Angular tanstack query

ScriptType:
Second try (before I did not branch from dev ...)

Heyho, pretty new to the open-source world.
As I use Angular and NestJS in my tech stack and found this interesting library I was pretty sad that the plugin does not work with Angular, so I created an implementation.
Sadly, I can't get the test running on my system. (even without any changes to the code) (It searches the package.json in a very wrong directory. I don't know why, and I didn't want to change anything on the setup/config files.)

### DIFF
--- a/packages/plugins/tanstack-query/package.json
+++ b/packages/plugins/tanstack-query/package.json
@@ -58,6 +58,12 @@
             "import": "./runtime-v5/svelte.mjs",
             "require": "./runtime-v5/svelte.js",
             "default": "./runtime-v5/svelte.js"
+        },
+        "./runtime-v5/angular": {
+            "types": "./runtime-v5/angular.d.ts",
+            "import": "./runtime-v5/angular.mjs",
+            "require": "./runtime-v5/angular.js",
+            "default": "./runtime-v5/angular.js"
         }
     },
     "repository": {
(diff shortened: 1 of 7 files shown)


packages/plugins/tanstack-query/package.json
```diff
@@ -87,7 +93,14 @@
         "ts-morph": "^16.0.0",
         "ts-pattern": "^4.3.0"
     },
+    "peerDependencies": {
+        "@tanstack/angular-query-experimental": "^5.0.0"
```
ymc9: Hi @ScriptType , previously this package didn't have "@tanstack/*-query" as peer dependencies, and assumed users installed the proper version.

I think the correct approach is to have them as peer dependencies but mark them optional with "peerDependenciesMeta" (so that angular users don't install react-query). However, I'd suggest we keep it as is for now (meaning to remove this peer dependency here) to keep things consistent and avoid unnecessary potential issues.

ymc9 (commented):
Hi @ScriptType ,

Thanks for making this PR! I've never used Angular or tanstack-angular-query, but the change looks very good to me. I'll merge it with the dev branch where I made a few fixes related to the observed test failures. If the issues are resolved, I'll approve the PR and get it merged.

Will you be interested in adding documentation for it too?

ScriptType:
Removed the Peer-Dependencie.
The Documentation would be a separate PR in the other Repo ? Sure i can do it.
But my English is not the best, so I would let me AI help correct the text and stuff if this is ok.

Is there anything else I can help you with inside the project?

ymc9:
> Removed the Peer-Dependencie. The Documentation would be a separate PR in the other Repo ? Sure i can do it. But my English is not the best, so I would let me AI help correct the text and stuff if this is ok.
> 
> Is there anything else I can help you with inside the project?

Awesome! Yes, docs are in a separate repo here:

https://github.com/zenstackhq/zenstack-docs/blob/main/docs/reference/plugins/tanstack-query.mdx

I think simply mirroring current content about react/vue would be sufficient. I can help with some touchup if needed. Thanks! @ScriptType

PR #2252, opened 2025-09-25
Rest api sparse fieldsets

sw34:
implements sparse fieldsets as shown in the specification https://jsonapi.org/format/#fetching-sparse-fieldsets

### DIFF
--- a/packages/server/src/api/rest/index.ts
+++ b/packages/server/src/api/rest/index.ts
@@ -585,7 +585,14 @@ class RequestHandler extends APIHandlerBase {
             select = relationSelect;
         }
 
-        select = select ?? { [relationship]: true };
+        // handle partial results for requested type
+        if (!select) {
+            const { select: partialFields, error } = this.buildPartialSelect(lowerCaseFirst(relationInfo.type), query);
+            if (error) return error;
+
+            select = partialFields ? { [relationship]: { select: { ...partialFields } } } : { [relationship]: true };
+        }
+
         const args: any = {
             where: this.makePrismaIdFilter(typeInfo.idFields, resourceId),
             select,
@@ -771,8 +778,6 @@ class RequestHandler extends APIHandlerBase {
             };
         } else {
             args.take = limit;
-
-            console.log('args', JSON.stringify(args));
 
             const [entities, count] = await Promise.all([
                 prisma[type].findMany(args),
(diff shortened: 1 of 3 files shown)


packages/server/src/api/rest/index.ts
```diff
@@ -566,7 +585,14 @@ class RequestHandler extends APIHandlerBase {
...
 
-        select = select ?? { [relationship]: true };
+        // handle partial results for requested type
+        if (!select) {
+            const { select: partialFields, error } = this.buildPartialSelect(lowerCaseFirst(relationInfo.type), query);
+            if (error) return error;
+
+            select = partialFields ? { [relationship]: { select: { ...partialFields } } } : { [relationship]: true };
+        }
+
```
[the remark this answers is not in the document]
sw34: relationship was already vetted on lines 567-570 to be a valid model.

```
const relationInfo = typeInfo.relationships[relationship];
if (!relationInfo) {
  return this.makeUnsupportedRelationshipError(type, relationship, 404);
}
```

packages/server/src/api/rest/index.ts
```diff
@@ -1820,11 +1892,19 @@ class RequestHandler extends APIHandlerBase {
                         return { select: undefined, error: this.makeUnsupportedModelError(relationInfo.type) };
                     }
 
+                    // handle partial results for requested type
+                    const { select, error } = this.buildPartialSelect(lowerCaseFirst(relationInfo.type), query);
+                    if (error) return { select: undefined, error };
+
                     if (i !== parts.length - 1) {
                         currPayload[relation] = { include: { ...currPayload[relation]?.include } };
```
ymc9: This branch handles intermediate levels of deep selection like `include=profile.address.street`. I think we should also respect fieldset if the caller provided one instead of selecting all fields?

ymc9 (commented):
Hi @sw34 , thanks a lot for making this PR and it overall looks great to me. Please check my review comments. Would be like to change the docs as well? Or I can take care of it if you're busy.

packages/server/src/api/rest/index.ts
sw34: updated for deep includes, forgot to go back to that

sw34: I will work on the documentation later today or tomorrow.

Issue #1532, opened 2024-06-24
[Feature Request] Multiple entry zmodels in VS Code extension


platon-ivanov:
**Is your feature request related to a problem? Please describe.**
Currently using a structure with multiple entry zmodel schemas in different directories with their own `User` models or models with `@@auth` flags only respects one of the auth models in the VS Code extension.

```
- dbs
  - db1
    - schema.zmodel
  - db2
    - schema.zmodel
```

```prisma
model User {
  @@allow('read', auth() == this) // works fine in `db1` directory. "incompatible operand types" when in `db2` directory.
}
```

**Describe the solution you'd like**
As @jiashengguo pointed out, a configuration setting could be introduced to support multiple schemas:

> I think it's related to how the VSCode plugin works.  From VSCode point of view, there is no entry schema as in CLI,  so it just  loads all the ZModel file could find in the project and try to compile them together.   That's why you will see the error because there are two User models now.
> 
> I think in order to support multiple schema set in the same project, we probably need to introduce some configuration setting like 
>
> ```json
> {
>   "zmodelEntries": ["**/dbs/db1/schema.zmodel",  "**/dbs/db2/schema.zmodel"]  
> } 
> ```

jasonmacdonald:
I just hit this as well and couldn't figure out why my schemas were now throwing errors. We use a mono repo, and I added another schema to a package we are building for a Zenstack plugin as an artifact for testing. 

Thanks for finding this. It was causing me some undue stress trying to determine why I was getting this error. At least now I know why and that it's just VSCode, but it would be good to resolve this issue as, like me, others may hit this error and not realize the root cause.

ymc9:
Hi @jasonmacdonald , thanks for offering to fix it! Since it's only for IDE we probably should configure it inside vscode settings. Let me check that part of code and share more thoughts back here soon.

ymc9:
Hey @jasonmacdonald , thanks for the wait. I checked the code and here're my thoughts:

- First we can introduce a VSCode extension setting to specify the "entry point" zmodel patterns. Maybe just use glob patterns? Right now the only format ZenStack has is the one about formatting. You can use this code as a reference: https://github.com/zenstackhq/zenstack/blob/df9b0ff504feeb1db071bd1dee9f95d1e2acc619/packages/schema/src/language-server/zmodel-formatter.ts#L81C1-L82C1

[...]

PR #2295, opened 2025-11-05
fix: monorepo's wrong auth error

cqh963852:
# Change

- add getCurrentNodeAndReachableDataModelsAndTypeDefs

When performing auth() expression validation, the current approach is to retrieve the relevant decl data from all documents. 

However, this is prone to errors in a monorepo, as described in the issue. 

Therefore, my improvement is to start from the current document, retrieve all documents corresponding to the imports, and then parse them.

related issue https://github.com/zenstackhq/zenstack/issues/1532

### DIFF
--- a/packages/schema/src/language-server/zmodel-linker.ts
+++ b/packages/schema/src/language-server/zmodel-linker.ts
@@ -55,7 +55,7 @@ import {
 } from 'langium';
 import { match } from 'ts-pattern';
 import { CancellationToken } from 'vscode-jsonrpc';
-import { getAllLoadedAndReachableDataModelsAndTypeDefs, getContainingDataModel } from '../utils/ast-utils';
+import { getContainingDataModel, getCurrentNodeAndReachableDataModelsAndTypeDefs } from '../utils/ast-utils';
 import { isMemberContainer } from './utils';
 import { mapBuiltinTypeToExpressionType } from './validator/utils';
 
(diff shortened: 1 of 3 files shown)


cqh963852:
@ymc9 @jasonmacdonald 

Excuse me, could you please review my changes? Do you think they are reasonable, and would such changes cause any disruptive changes?

ymc9:
Hi @cqh963852 , thanks for making this PR!

The original design was for a convenience concern. Consider the following example:

```zmodel
model Post {
    ...
    @@allow('all', auth().role == 'ADMIN')
}
```

The `Post` model has no relation to `User`, thus doesn't import the zmodel that contains `User`, the current design still resolves `auth()` by checking all loaded models, even those not imported. With a design change, the user would need to explicitly import it in this case, if I understand it correctly.

For your specific problem, does the resolution issue only happen inside IDE, or with the `zenstack generate` as well?

cqh963852:
> For your specific problem, does the resolution issue only happen inside IDE, or with the zenstack generate as well?

Only happened inside vscode. ZenStack Generate works correctly.

cqh963852:
> the current design still resolves auth() by checking all loaded models

Could we adjust this design? maybe

1. If the user has specified entries, Load only documents with intersecting paths; otherwise, load from all documents. 

```json
{
    "zmodel.entries":["package/abc",  "package/def"]
}
```

package/abc/post.zmodel ,only load package/abc's documents
```
model Post {
    ...
    @@allow('all', auth().role == 'ADMIN')
}
```

package/def/post.zmodel ,only load package/def's documents
```
model Post {
    ...
    @@allow('all', auth().role == 'ADMIN')
}
```

ymc9:
> > the current design still resolves auth() by checking all loaded models
> 
> Could we adjust this design? maybe
> 
> 1. If the user has specified entries, Load only documents with intersecting paths; otherwise, load from all documents.
> 
> ```json
> {
>     "zmodel.entries":["package/abc",  "package/def"]
> }
> ```
> 
> package/abc/post.zmodel ,only load package/abc's documents
> 
> ```
> model Post {
>     ...
>     @@allow('all', auth().role == 'ADMIN')
> }
> ```
> 
> package/def/post.zmodel ,only load package/def's documents
> 
> ```
> model Post {
>     ...
>     @@allow('all', auth().role == 'ADMIN')
> }
> ```

Absolutely! I think similar ideas were brought up before but never got implemented. The setting only needs to be recognized by the language server to limit the transitive closure built for resolving `auth()`.

ymc9:
Closing for now.

PR #2305, opened 2025-11-19
fix: make datasource url optional

ymc9:


jamiter:
FYI, for Prisma 7 the url has been deprecated completely and this change is required to upgrade to v7.

ymc9:
Thanks @jamiter . The current plan is to not support v7 for ZenStack v2. I'll share this information with the next release and collect feedback.

PR #2369, opened 2026-02-11
fix: enhance delegate model interaction with afterEntityMutation plugin support in child updates

### LINKED ISSUE #2367: [Bug]: Update on delegate child model fails with "missing FROM-clause entry" when entity mutation plugins are present
**Description**
Updating a child model in a delegate hierarchy fails with a SQL error when an entity mutation plugin with an `afterEntityMutation` hook is registered. The error is:

- **PostgreSQL**: `missing FROM-clause entry for table "Base"`
- **SQLite**: `no such column: Base.field`

**Additional context**

When any plugin registers an `afterEntityMutation` hook, the query executor globally replaces `RETURNING "id"` with `RETURNING *` on all UPDATE/INSERT queries (`zenstack-query-executor.ts:224-234`). This causes the following chain of events during a child model update:

**Steps to Reproduce**
1. Define a delegate hierarchy where the base model has non-ID fields:

```
model Asset {
    id Int @id @default(autoincrement())
    organizationId String
    assetType String
    @@delegate(assetType)
}

model Product extends Asset {
    name String
}
```

2. Register any entity mutation plugin with an afterEntityMutation hook:

```
client.$use({
    id: 'my-plugin',
    onEntityMutation: {
        afterEntityMutation() {
            // even a noop triggers the bug
        },
    },
});
```

3. Create and then update a child entity with data that touches both base and child fields:

```
const group = await client.Post.create({
    data: { organizationId: 'org1', name: 'My Group' },
});

// This fails:
await client.Post.update({
    where: { id: group.id },
    data: { organizationId: 'org2', name: 'Updated Group' },
});
```

**Notes**
- The bug only triggers when the update includes child-specific fields (otherwise the child UPDATE is skipped entirely)
- The afterEntityMutation hook doesn't need to do anything — its mere presence causes RETURNING * globally via hasEntityMutationPluginsWithAfterMutationHooks
- Policy plugins are not required to reproduce the issue, though in practice users typically have both policy and entity mutation plugins active

genu:
Fixes #2367

## Summary by CodeRabbit

* **Bug Fixes**
  * Child record updates for inherited models now use strict ID-based filters to prevent incorrect update scopes and improve reliability.
  * Narrower update filters reduce risk of accidental multi-record updates when base-model data is involved.

* **Tests**
  * Added end-to-end tests covering delegate model interactions and entity mutation hooks to validate create/update flows and field propagation.

### DIFF
--- a/tests/e2e/orm/plugin-infra/entity-mutation-hooks.test.ts
+++ b/tests/e2e/orm/plugin-infra/entity-mutation-hooks.test.ts
@@ -691,11 +691,13 @@ describe('Entity mutation hooks tests', () => {
             await expect(client.user.findMany()).toResolveWithLength(0);
         });
     });
+});
 
-    describe('Delegate model interaction', () => {
-        it('update on child model succeeds with afterEntityMutation plugin', async () => {
-            const client = await createTestClient(
-                `
+describe('Entity mutation hooks - delegate model interaction', () => {
+    it('update on child model succeeds with afterEntityMutation plugin', async () => {
+        console.log('RUNNING!!!!');
+        const client = await createTestClient(
+            `
 model Base {
     id Int @id @default(autoincrement())
     orgId Int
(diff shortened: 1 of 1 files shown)


ymc9 (dismissed):
Hi @genu , thanks for making the fix, and I think it's a good safety measure. A deeper problem is, when mutation hooks are used in plugins, the underlying query executor "forces injecting" a `return *` clause, resulting in more fields returned and breaking some assumptions on the upper layer. I'll see I can make a separate PR on the root cause.

I've merged your branch with the latest "dev" and refined the comments a bit. Will merge once CI passes

PR #2377, opened 2026-02-12
feat: `customId` function

### LINKED ISSUE #2348: nanoid  support  custom alphabet
Most of the time, we need to define a special business ID field for models, such as `post_id` or `user_id`. We often choose UUID or nanoid for generating these IDs. However, when using nanoid, we found that it cannot meet our requirements because it defaults to generating IDs with uppercase letters and numbers. Sometimes we prefer IDs in lowercase letters and numbers, which nanoid cannot handle. As a result, we are forced to set the ID field as non‑nullable and generate a compliant ID in the code layer when creating records.

This is not an ideal practice. Providing a custom alphabet feature for nanoid would help reduce this mental overhead. Moreover, I believe the optimal API design would be:

- Keep the usage as is:  
  `@default(nanoid(16))`

- Add an optional parameter when creating the client:  
```typescript
export const db = new ZenStackClient(schema, {
    dialect: new SqliteDialect({
        database: new SQLite(':memory:'),
    }),
    nanoidCustomAlphabet: "1234567890abcdef"
});
```

sanny-io:
Allows developers to specify a custom implementation of an ID generation function.

## Sample Usage

```prisma
model User {
    id String @id @default(customId()) // length chosen by implementation
}

model Post {
    id String @id @default(customId(8)) // length chosen by schema
}
```

```typescript
export const db = new ZenStackClient(schema, {
    dialect: new SqliteDialect({
        database: new SQLite(':memory:'),
    }),

    customId: ({ model, length, client }) => {
        return ...
    },
});
```

Closes #2348
Prisma issues: prisma/prisma#6719

## Summary by CodeRabbit

* **New Features**
  * Configurable custom ID generation via a client-supplied function, with optional length.
  * Default ID generation now receives per-entity and per-field context so IDs can vary by entity/field.
  * Runtime checks ensure a provided length is a positive number and generated IDs are non-empty strings.

* **Tests**
  * Added unit tests for length validation.
  * Added end-to-end tests covering custom ID outputs, nested creations, and missing/invalid implementation cases.

### DIFF
--- a/packages/orm/src/client/crud/operations/base.ts
+++ b/packages/orm/src/client/crud/operations/base.ts
@@ -25,6 +25,7 @@ import { NUMERIC_FIELD_TYPES } from '../../constants';
 import { TransactionIsolationLevel, type ClientContract, type CRUD } from '../../contract';
 import type { FindArgs, SelectIncludeOmit, WhereInput } from '../../crud-types';
 import {
+    createConfigError,
     createDBQueryError,
     createInternalError,
     createInvalidInputError,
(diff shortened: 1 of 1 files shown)


sanny-io:
@ymc9 only non-empty strings are allowed at the moment. Should the restrictions be relaxed, or maybe tightened?

Not sure if, for example, a string full of only whitespaces could cause issues.

ymc9:
Hi @sanny-io , thanks for working on this.

I think defining custom functions is useful for several use cases, like custom id generation in this PR, and maybe non-id default values, and access policies. Maybe we should plan it as a more generic feature instead?

sanny-io:
@ymc9 yeah, if custom functions would cover this use case we can close this 👍

oriole-pub:
So, in what form will this feature appear? Will it be in future plans?
@sanny-io @ymc9

sanny-io:
@oriole-pub I'm assuming it will be similar to procedures in that they are declared in ZModel and defined in the client.

```prisma
model User {
    id String @id @default(generateId())
}

function generateId(length: Int?): String
```

```typescript
export const db = new ZenStackClient(schema, {
    dialect: new SqliteDialect({
        database: new SQLite(':memory:'),
    }),

    functions: {
        generateId: (...) => {
            ...
        },
    },
});
```

PR #2396, opened 2026-02-22
fix: cast auth() string values to uuid on PostgreSQL when field has @db.Uuid

### LINKED ISSUE #2394: PostgreSQL: policy check fails with "operator does not exist: text = uuid" when auth() field has @db.Uuid
## Description

When using PostgreSQL with UUID primary keys (`@db.Uuid`) and access policies that reference `auth().userId`, the policy-check SQL fails at runtime with:

```
operator does not exist: text = uuid
```

### ZModel schema

```zmodel
model User {
  userId String @id @db.Uuid @default(uuid())
  @@auth()
}

model Client {
  clientId     String     @id @db.Uuid @default(uuid())
  dealershipId String     @db.Uuid
  Dealership   Dealership @relation(fields: [dealershipId], references: [dealershipId])

  @@allow('all', Dealership.UserAccess?[userId == auth().userId])
}

model DealershipUserAccess {
  dealershipId String @db.Uuid
  userId       String @db.Uuid
  // ...relations
}
```

### What happens

1. `auth().userId` is resolved as a JavaScript string value (e.g., `"550e8400-e29b-41d4-a716-446655440000"`)
2. The `valueMemberAccess` method in `expression-transformer.ts` calls `transformValue(value, 'String')` — because the ZModel type is `String`
3. `transformValue` passes the string through to the PostgreSQL dialect's `transformInput`, which is a no-op for strings
4. Kysely creates a parameterized query binding the value as `$1` with PostgreSQL type `text`
5. The generated SQL compares this `text` parameter against a `uuid` column: `"DealershipUserAccess"."userId" = $1`
6. PostgreSQL raises `operator does not exist: text = uuid` because there is no implicit cast from `text` to `uuid`

### Root cause

In `packages/plugins/policy/src/expression-transformer.ts`, the `transformValue` method receives only the `BuiltinType` (e.g., `'String'`). It does not have access to the field's native database type attribute (`@db.Uuid`), which is available in the `FieldDef.attributes` array as `{ name: "@db.Uuid" }`.

The PostgreSQL dialect's `transformInput` method (in `packages/orm/src/client/crud/dialects/postgresql.ts`) has no handling for UUID values — it falls through to the default pass-through for strings.

### Expected behavior

When a `String` field has `@db.Uuid` (or other native type attributes), the generated SQL should cast the parameter value appropriately. For example:

```sql
-- Current (broken):
"DealershipUserAccess"."userId" = $1   -- $1 is text, column is uuid → ERROR

-- Expected (working):
"DealershipUserAccess"."userId" = $1::uuid   -- explicit cast
```

## Potential fix approaches

1. **Extend `transformValue` / `transformInput` to accept native type info**: Pass the `@db.Uuid` attribute information from `FieldDef.attributes` through to the value transformation, and have the PostgreSQL dialect wrap the value in a `::uuid` cast using `sql.raw` or `eb.cast()`.

2. **Handle it at the binary comparison level**: In `_binary` or `transformAuthBinary`, when one side is a column reference to a `@db.Uuid` field and the other is a value, wrap the value with a cast.

## Current workaround

Users can work around this by adding an implicit cast in a Prisma migration:

```sql
CREATE CAST (text AS uuid) WITH INOUT AS IMPLICIT;
```

However, this is a global database-level change that may have unintended side effects.

## Environment

- ZenStack: v3.3.3
- PostgreSQL: 16
- Prisma: 6.x
- Node.js: 22

niehaus1301:
## Problem

When using `auth()` in access policies with PostgreSQL and the auth model has fields annotated with `@db.Uuid`, the policy check fails with:

```
operator does not exist: text = uuid
```

This happens because parameterized string values from `auth()` are sent as `text` type, but the corresponding database columns are `uuid` type. PostgreSQL does not implicitly cast between these types.

### Reproduction

```zmodel
model User {
  userId String @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  email  String @unique
  // ...
  @@auth()
}

model Client {
  clientId     String @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  dealershipId String @db.Uuid
  // ...
  @@allow('read', auth().dealershipId == dealershipId)
}
```

The generated SQL compares `auth().dealershipId` (a `text` parameter) against the `dealershipId` column (`uuid` type), which PostgreSQL rejects.

## Root Cause

In `expression-transformer.ts`, the `valueMemberAccess` method resolves `auth()` member fields and calls `transformValue(curr, currType)` where `currType` is the ZModel type (`'String'`). The PostgreSQL dialect's `transformInput` for `String` is a no-op, so the value remains a plain text parameter. When Kysely parameterizes this as `$1`, PostgreSQL infers `text` type, causing the type mismatch.

The `@db.Uuid` native type attribute *is* available in the runtime schema via `FieldDef.attributes`, but was not being utilized during value transformation.

## Fix

Added an `applyNativeTypeCast` method that inspects the field's native type attributes and wraps the value node with an explicit `::uuid` SQL cast when:
1. The provider is PostgreSQL, AND
2. The field has the `@db.Uuid` attribute

The cast is applied in two code paths:
- **`valueMemberAccess`**: For `auth().field` member access in policies (the primary case)
- **`_field` with `contextValue`**: For field evaluation in collection predicates with value objects

### Changes

Only one file is modified: `packages/plugins/policy/src/expression-transformer.ts`
- Added `sql` to Kysely imports
- Modified `valueMemberAccess` to store the full `FieldDef` and apply native type cast after `transformValue`
- Modified `_field` with `contextValue` to also apply native type cast
- Added `applyNativeTypeCast` helper with JSDoc documentation

Fixes #2394

## Bug Fixes

* Fixed policy expression evaluation to correctly handle UUID column comparisons in PostgreSQL databases. Policy expressions now properly cast values when evaluating against UUID-typed fields.

### DIFF
--- a/packages/plugins/policy/src/expression-transformer.ts
+++ b/packages/plugins/policy/src/expression-transformer.ts
@@ -40,6 +40,7 @@ import {
     ReferenceNode,
     SelectionNode,
     SelectQueryNode,
+    sql,
     TableNode,
     ValueListNode,
     ValueNode,
@@ -187,7 +188,8 @@ export class ExpressionTransformer<Schema extends SchemaDef> {
         if (context.contextValue) {
             // if we're transforming against a value object, fields should be evaluated directly
             const fieldDef = QueryUtils.requireField(this.schema, context.modelOrType, expr.field);
-            return this.transformValue(context.contextValue[expr.field], fieldDef.type as BuiltinType);
+            const node = this.transformValue(context.contextValue[expr.field], fieldDef.type as BuiltinType);
+            return this.applyNativeTypeCast(node, fieldDef);
         }
 
         const fieldDef = QueryUtils.requireField(this.schema, context.modelOrType, expr.field);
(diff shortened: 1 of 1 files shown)


lsmith77:
https://github.com/zenstackhq/zenstack/pull/2473

ymc9:
Closing since the issue has been fixed with https://github.com/zenstackhq/zenstack/commit/46867201783642d57b9e2b49fbbe60d19fdae38c

PR #2399, opened 2026-02-23
feat(transaction): implement transaction handling with sequential operations and validation

jiashengguo:
## Summary by CodeRabbit

* **New Features**
  * Added a transactional RPC endpoint for batched, sequential operations with atomic execution. Requests are POST-only, each operation is validated, results are serialized with metadata, structured errors are returned for failures, and the entire batch rolls back on error.

* **Tests**
  * Added comprehensive tests covering multi-operation success, method and payload validation, per-operation serialization/deserialization, ORM error mapping, and rollback behavior.

### DIFF
--- a/packages/server/src/api/rpc/index.ts
+++ b/packages/server/src/api/rpc/index.ts
@@ -240,7 +240,7 @@ export class RPCApiHandler<Schema extends SchemaDef = SchemaDef> implements ApiH
             if (!this.isValidModel(client, lowerCaseFirst(itemModel))) {
                 return this.makeBadInputErrorResponse(`operation at index ${i} has unknown model: ${itemModel}`);
             }
-            if (itemArgs !== undefined && itemArgs !== null && typeof itemArgs !== 'object') {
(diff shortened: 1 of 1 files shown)


packages/server/src/api/rpc/index.ts
```diff
@@ -11,6 +11,26 @@ import { loggerSchema } from '../common/schemas';
...
+    'aggregate',
+    'groupBy',
+    'count',
+    'exists',
+    'update',
+    'updateMany',
+    'updateManyAndReturn',
+    'delete',
+    'deleteMany',
+]);
```
sanny-io: Should be able to use the `CoreCrudOperations` array that is already defined in `packages/orm/src/client/crud/operations/base.ts`

jiashengguo: Good point! You know the code better than claude! 😄

ymc9 (approved):
I think we can add transaction support to the tanstack hooks too. Shall we create a feature request first?

jiashengguo:
> I think we can add transaction support to the tanstack hooks too. Shall we create a feature request first?

ok.  @copilot  would you please help create a feature request for it?

PR #2425, opened 2026-03-01
fix(orm): fallback to compact temp aliases for overlong names

### LINKED ISSUE #2424: PolicyPlugin deep nested include generates malformed PostgreSQL alias reference (42703)
## Description and expected behavior

  With PolicyPlugin enabled, a deep nested include query generates broken SQL alias references on PostgreSQL and fails with SQLSTATE 42703 (column ...productSku does not exist).

  This reproduces on 3.4.1 (PostgreSQL, useCompactAliasNames: false).

  Minimal schema:
```
  plugin policy {
    provider = '@zenstackhq/plugin-policy'
  }

  plugin prisma {
    provider = '@core/prisma'
    output = 'prisma/schema.prisma'
  }

  model Store {
    id                  String @id
    customerOrders      CustomerOrder[]
    productCatalogItems ProductCatalogItem[]
    @@allow('all', true)
  }

  model CustomerOrder {
    id                          String @id
    storeId                     String
    store                       Store @relation(fields: [storeId], references: [id], onDelete: Cascade)
    customerOrderPaymentSummary CustomerOrderPaymentSummary[]
    @@allow('all', true)
  }

  model CustomerOrderPaymentSummary {
    id                              String @id
    customerOrderId                 String
    customerOrder                   CustomerOrder @relation(fields: [customerOrderId], references: [id], onDelete: Cascade)
    customerOrderPaymentSummaryLine CustomerOrderPaymentSummaryLine[]
    @@allow('all', true)
  }

  model PaymentTransaction {
    id                              String @id
    customerOrderPaymentSummaryLine CustomerOrderPaymentSummaryLine[]
    paymentTransactionLineItem      PaymentTransactionLineItem[]
    @@allow('all', true)
  }

  model CustomerOrderPaymentSummaryLine {
    customerOrderPaymentSummaryId String
    lineIndex                    Int
    paymentTransactionId         String

    customerOrderPaymentSummary  CustomerOrderPaymentSummary @relation(fields: [customerOrderPaymentSummaryId], references: [id], onDelete: Cascade)
    paymentTransaction           PaymentTransaction @relation(fields: [paymentTransactionId], references: [id], onDelete: Cascade)

    @@id([customerOrderPaymentSummaryId, lineIndex])
    @@allow('all', true)
  }

  model ProductCatalogItem {
    storeId                    String
    sku                        String

    store                      Store @relation(fields: [storeId], references: [id], onDelete: Cascade)
    paymentTransactionLineItem PaymentTransactionLineItem[]

    @@id([storeId, sku])
    @@allow('all', true)
  }

  model InventoryReservation {
    id                         String @id
    paymentTransactionLineItem PaymentTransactionLineItem[]
    @@allow('all', true)
  }

  model PaymentTransactionLineItem {
    paymentTransactionId   String
    lineNumber             Int
    storeId                String
    productSku             String
    inventoryReservationId String?

    paymentTransaction     PaymentTransaction @relation(fields: [paymentTransactionId], references: [id], onDelete: Cascade)
    productCatalogItem     ProductCatalogItem @relation(fields: [storeId, productSku], references: [storeId, sku])
    inventoryReservation   InventoryReservation? @relation(fields: [inventoryReservationId], references: [id], onDelete: SetNull)

    @@id([paymentTransactionId, lineNumber])
    @@allow('all', true)
  }
```
  Repro query:
```
  await db.customerOrderPaymentSummary.findUnique({
    where: { id: 'summary_1' },
    include: {
      customerOrder: true,
      customerOrderPaymentSummaryLine: {
        include: {
          paymentTransaction: {
            include: {
              paymentTransactionLineItem: {
                include: {
                  productCatalogItem: true,
                  inventoryReservation: true,
                },
              },
            },
          },
        },
      },
    },
  });
```
  Observed error:
```
  error: column $$_CustomerOrderPaymentSummary$customerOrderPaymentSummaryLine$.productSku does not exist
  SQLSTATE: 42703
```
  Expected behavior: the query should execute successfully and return nested data; generated aliases should be valid and consistently referenced.

  ## Screenshots

  N/A

  ## Environment (please complete the following information):

  - ZenStack version: 3.4.1
  - Database type: PostgreSQL (16)
  - Node.js/Bun version: Node.js 22.x
  - Package manager: pnpm

  ## Additional context

  Generated SQL contains very long aliases in this deep nested include path, e.g.:
```
  - "$$_CustomerOrderPaymentSummary$customerOrderPaymentSummaryLine$paymentTransaction$paymentTransactionLineItem$sub"
  - "$$_CustomerOrderPaymentSummary$customerOrderPaymentSummaryLine$paymentTransaction$paymentTransactionLineItem$productCatalogItem$sub"
```
  But later SQL references a malformed alias segment `(...$customerOrderPaymentSummaryLine$.productSku)`, causing the missing-column failure.

  Potentially related areas:

  - packages/orm/src/client/crud/dialects/lateral-join-dialect-base.ts
  - packages/orm/src/client/query-utils.ts (tmpAlias)
  - packages/orm/src/client/executor/zenstack-query-executor.ts (processTempAlias)

pkudinov:
Fixes #2424

  ## Summary

  This PR fixes deep nested include failures caused by overlong temporary relation aliases when `useCompactAliasNames: false`.

  - Keeps existing behavior for `useCompactAliasNames: false` when aliases are short/safe.
  - Adds a safety fallback: if any temp alias exceeds a safe identifier length threshold, ZenStack applies the existing compact alias transformer for that query.
  - Leaves the default path (`useCompactAliasNames: true`) unchanged.

  ## Root Cause

  Deep relation paths generate long temp aliases (via `$$_...`).
  When aliases exceed identifier limits, they can be truncated/collide and break reference resolution in nested joins, resulting in errors like `42703` (`column ... does not exist`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved alias handling to ensure consistent, stable temporary aliases in all modes, preventing incorrect truncation or mismatched nested include results.

* **Documentation**
  * Clarified alias examples and fallback behavior when compact aliasing is disabled and identifier length limits apply.

* **Tests**
  * Added a regression test validating deep nested include behavior with non-compact aliasing.

### DIFF
--- a/packages/orm/src/client/executor/temp-alias-transformer.ts
+++ b/packages/orm/src/client/executor/temp-alias-transformer.ts
@@ -17,15 +17,17 @@ export class TempAliasTransformer extends OperationNodeTransformer {
     private readonly textEncoder = new TextEncoder();
     private readonly mode: TempAliasTransformerMode;
     private readonly maxIdentifierLength: number;
-    private detectOnly = false;
-    private hasOverlongTempAlias = false;
 
     constructor(options: TempAliasTransformerOptions = {}) {
         super();
         this.mode = options.mode ?? 'alwaysCompact';
         // PostgreSQL limits identifier length to 63 bytes and silently truncates overlong aliases.
         const maxIdentifierLength = options.maxIdentifierLength ?? 63;
-        if (!Number.isFinite(maxIdentifierLength) || !Number.isInteger(maxIdentifierLength) || maxIdentifierLength <= 0) {
+        if (
+            !Number.isFinite(maxIdentifierLength) ||
+            !Number.isInteger(maxIdentifierLength) ||
+            maxIdentifierLength <= 0
+        ) {
             throw new RangeError('maxIdentifierLength must be a positive integer');
         }
         this.maxIdentifierLength = maxIdentifierLength;
@@ -33,44 +35,35 @@ export class TempAliasTransformer extends OperationNodeTransformer {
 
     run<T extends OperationNode>(node: T): T {
         this.aliasMap.clear();
-        if (this.mode === 'compactLongNames') {
-            const hasOverlongAliases = this.detectOverlongTempAlias(node);
-            if (!hasOverlongAliases) {
-                return node;
-            }
-        }
         return this.transformNode(node);
     }
 
-    private detectOverlongTempAlias(node: OperationNode) {
-        this.detectOnly = true;
-        this.hasOverlongTempAlias = false;
-        this.transformNode(node);
-        this.detectOnly = false;
-        return this.hasOverlongTempAlias;
-    }
-
     protected override transformIdentifier(node: IdentifierNode, queryId?: QueryId): IdentifierNode {
-        if (this.detectOnly) {
-            if (!node.name.startsWith(TEMP_ALIAS_PREFIX)) {
-                return node;
-            }
-
-            const aliasByteLength = this.textEncoder.encode(node.name).length;
-            if (aliasByteLength > this.maxIdentifierLength) {
-                this.hasOverlongTempAlias = true;
-            }
-            return node;
+        if (!node.name.startsWith(TEMP_ALIAS_PREFIX)) {
+            return super.transformIdentifier(node, queryId);
         }
 
-        if (node.name.startsWith(TEMP_ALIAS_PREFIX)) {
+        let shouldCompact = false;
+        if (this.mode === 'alwaysCompact') {
+            shouldCompact = true;
+        } else {
+            // check if the alias name exceeds the max identifier length, and
+            // if so, compact it
+            const aliasByteLength = this.textEncoder.encode(node.name).length;
+            if (aliasByteLength > this.maxIdentifierLength) {
+                shouldCompact = true;
+            }
+        }
+
+        if (shouldCompact) {
             let mapped = this.aliasMap.get(node.name);
             if (!mapped) {
                 mapped = `$$t${this.aliasMap.size + 1}`;
                 this.aliasMap.set(node.name, mapped);
             }
             return IdentifierNode.create(mapped);
+        } else {
+            return super.transformIdentifier(node, queryId);
         }
-        return super.transformIdentifier(node, queryId);
     }
 }


ymc9:
Hi @pkudinov , with the name compaction, the `useCompactAliasNames` is more a debugging facility now. Any specific reason you set it to true?

pkudinov:
Hi @ymc9, excited to get to contribute to Zenstack. We caught this issue while migrating from V2 to V3.

I did not flip `useCompactAliasNames` to true globally in this PR - `false` is still used by default. Compaction only kicks in as a safety fallback when temp aliases become overlong.

For #2424 specifically, alias collision is reproducible due to Postgres identifier truncation (63 bytes). In the failing SQL, these distinct aliases are generated:

```
  - $$_CustomerOrderPaymentSummary$customerOrderPaymentSummaryLine$sub (66)
  - $$_CustomerOrderPaymentSummary$customerOrderPaymentSummaryLine$paymentTransaction$paymentTransactionLineItem$sub (112)
  - $$_CustomerOrderPaymentSummary$customerOrderPaymentSummaryLine$paymentTransaction$paymentTransactionLineItem$productCatalogItem$sub (131)
```
  All truncate to the same 63-byte prefix:
```
  - $$_CustomerOrderPaymentSummary$customerOrderPaymentSummaryLine$
```
So distinct qualifiers collapse to one identifier, and references like `...paymentTransactionLineItem$sub.productSku` are resolved in the wrong scope / become unresolved, leading to 42703 error.

Is there some better way to fix this?

ymc9:
> Hi @ymc9, excited to get to contribute to Zenstack. We caught this issue while migrating from V2 to V3.
> 
> I did not flip `useCompactAliasNames` to true globally in this PR - `false` is still used by default. Compaction only kicks in as a safety fallback when temp aliases become overlong.
> 
> For #2424 specifically, alias collision is reproducible due to Postgres identifier truncation (63 bytes). In the failing SQL, these distinct aliases are generated:
> 
> ```
>   - $$_CustomerOrderPaymentSummary$customerOrderPaymentSummaryLine$sub (66)
>   - $$_CustomerOrderPaymentSummary$customerOrderPaymentSummaryLine$paymentTransaction$paymentTransactionLineItem$sub (112)
>   - $$_CustomerOrderPaymentSummary$customerOrderPaymentSummaryLine$paymentTransaction$paymentTransactionLineItem$productCatalogItem$sub (131)
> ```
> 
> All truncate to the same 63-byte prefix:
> 
> ```
>   - $$_CustomerOrderPaymentSummary$customerOrderPaymentSummaryLine$
> ```
> 
> So distinct qualifiers collapse to one identifier, and references like `...paymentTransactionLineItem$sub.productSku` are resolved in the wrong scope / become unresolved, leading to 42703 error.
> 
> Is there some better way to fix this?

Thanks for the explanation @pkudinov . I was just wondering if the temp alias compaction logic had any issues that made you set the flag to `false` 😄.

For this PR, I feel it's cleaner to move the name length check into the `TempAliasTransformer` class so we consolidate all alias handling logic there. We can probably add mode to the class like "alwaysCompact" vs "compactLongNames". What do you think?

pkudinov:
@ymc9 please take a look, is that what you had in mind? I refactored accordingly in ff685cdc by moving the check into TempAliasTransformer with alwaysCompact/compactLongNames modes, and followed up in 560c5573 to use UTF-8 byte length for the 63-byte PostgreSQL
  limit.

pkudinov:
Implemented in 6cb00881 — TempAliasLengthChecker now short-circuits traversal after detecting the first overlong temp alias (with an additional guard in transformIdentifier), so deep ASTs avoid unnecessary work.

packages/orm/src/client/executor/temp-alias-transformer.ts
```diff
@@ -1,15 +1,79 @@
...
 import { TEMP_ALIAS_PREFIX } from '../query-utils';
 
+type TempAliasTransformerMode = 'alwaysCompact' | 'compactLongNames';
+
+type TempAliasTransformerOptions = {
+    mode?: TempAliasTransformerMode;
+    maxIdentifierLength?: number;
+};
+
+class TempAliasLengthChecker extends OperationNodeTransformer {
```
ymc9: Thanks for making the revision @pkudinov .

I think we probably don't need this separate transformer to detect length overflow. Instead, we can do the length check in the `transformIdentifier` of `TempAliasTransformer`.

I understand that the separate visitor can stop early and improve performance, but the effect is likely marginal.

pkudinov: @ymc9 fixed. Now it feels a little like between 2 fires - you and the Code Rabbit :-D

ymc9: Hi @pkudinov , I've made some further simplifications and pushed a commit. Will merge it once CI passes. Thank you for making the contribution!

PR #2442, opened 2026-03-04
feat(orm): add result plugin extension point

genu:
## Summary
Implement a new `result` extension point for plugins to declare computed fields on query results with full type safety and select/omit awareness. The feature includes recursive runtime computation, automatic dependency (needs) injection/stripping, and complete support for nested relations.

## Implementation
- Add `ExtResultFieldDef` and `ExtResultBase` types to plugin definitions
- Thread `ExtResult` generic through `ClientContract`, `ModelResult`, and all CRUD return types
- Implement `SelectAwareExtResult` type that respects select/omit clauses
- Add recursive `prepareArgsForExtResult` and `applyExtResult` runtime functions for nested relations
- Extend Zod validation to accept ext result fields in select/omit schemas
- Support multi-plugin composition and all client methods (`$use`, `$unuse`, `$transaction`, etc.)

## Testing
- 26 comprehensive tests covering findMany, findUnique, findFirst, create, update, upsert, delete, createManyAndReturn
- Type-level tests verifying select/omit-aware return types work correctly
- Tests for single-model and nested (include/select) ext result computation
- Tests for both to-one and to-many relations

## Usage Example
```typescript
const db = client.$use(definePlugin({
  id: 'display',
  result: {
    User: {
      fullName: {
        needs: { firstName: true, lastName: true },
        compute: (user) => `${user.firstName} ${user.lastName}`,
      },
    },
  },
}));

const users = await db.user.findMany({ include: { posts: true } });
users[0].fullName;        // "Alice Smith" (type-safe)
users[0].posts[0].author; // nested ext results also computed
```

## Summary by CodeRabbit

* **New Features**
  * Plugins can declare per-model computed "ext" result fields selectable/omittable; required dependencies are auto-injected and results computed post-operation, across nested relations and transactions.
  * Added a typed helper for declaring ext-result fields.

* **Refactor**
  * Ext-result generics and typings now propagate through client, plugin and model CRUD APIs; select/omit schemas and query args handle ext fields end-to-end.

* **Tests**
  * Added comprehensive end-to-end tests covering runtime behavior, typings, nesting, transactions, plugin composition and lifecycle.

### DIFF
--- a/packages/orm/src/client/client-impl.ts
+++ b/packages/orm/src/client/client-impl.ts
@@ -37,7 +37,12 @@ import { ZenStackQueryExecutor } from './executor/zenstack-query-executor';
 import * as BuiltinFunctions from './functions';
 import { SchemaDbPusher } from './helpers/schema-db-pusher';
 import type { ClientOptions, ProceduresOptions } from './options';
-import type { AnyPlugin, ExtResultFieldDef } from './plugin';
+import type { AnyPlugin } from './plugin';
+
+type ExtResultFieldDef = {
+    needs: Record<string, true>;
+    compute: (data: Record<string, any>) => unknown;
+};
 import { getField } from './query-utils';
 import { createZenStackPromise, type ZenStackPromise } from './promise';
 import { ResultProcessor } from './result-processor';
--- a/packages/orm/src/client/plugin.ts
+++ b/packages/orm/src/client/plugin.ts
@@ -21,21 +21,6 @@ export type ExtQueryArgsBase = {
 export type ExtClientMembersBase = Record<string, unknown>;
 
 /**
- * Definition for a single extended result field.
- * When used without type parameters, accepts any field names and untyped compute.
- */
-export type ExtResultFieldDef<Needs extends Record<string, true> = Record<string, true>> = {
-    /**
-     * Fields required to compute this result field.
-     */
-    needs: Needs;
-    /**
-     * Computes the result field value from the query result row.
-     */
-    compute: (data: { [K in keyof Needs]: any }) => unknown;
-};
-
-/**
  * Base shape of plugin-extended result fields.
  * Keyed by model name, each value maps field names to their definitions.
  * `needs` keys are constrained to non-relation fields of the corresponding model.


genu:
@coderabbitai resume

packages/orm/src/client/plugin.ts
```diff
@@ -20,13 +20,34 @@ export type ExtQueryArgsBase = {
...
+     * Computes the result field value from the query result row.
+     */
+    compute: (data: any) => unknown;
+};
+
+/**
+ * Base shape of plugin-extended result fields.
+ * Keyed by model name, each value maps field names to their definitions.
+ */
+export type ExtResultBase = Record<string, Record<string, ExtResultFieldDef>>;
```
ymc9: I think we can have a `Schema` parameter on this type, so that the outer Record can be `Record<GetModels<Schema>, ...>` for better typing.

packages/orm/src/client/plugin.ts
```diff
@@ -20,13 +20,34 @@ export type ExtQueryArgsBase = {
  */
 export type ExtClientMembersBase = Record<string, unknown>;
 
+/**
+ * Definition for a single extended result field.
+ */
+export type ExtResultFieldDef = {
```
ymc9: Same here, we should be able to add a `Schema` and `Model` parameters so as to type the `needs` field and `data` parameter.

packages/orm/src/client/plugin.ts
```diff
@@ -92,10 +119,41 @@ export function definePlugin<
...
+ *             fullName: resultField({
+ *                 needs: { firstName: true, lastName: true },
+ *                 compute: (user) => `${user.firstName} ${user.lastName}`,
+ *             }),
+ *         },
+ *     },
+ * });
+ * ```
+ */
+export function resultField<const N extends Record<string, true>, R>(def: {
```
ymc9, line 145: I think this helper is related to my previous comment. Just wondering if it's more straightforward to express the typing directly on the RuntimePlugin interface side.

packages/orm/src/client/crud-types.ts
```diff
@@ -2409,4 +2433,53 @@ type ExtractExtQueryArgs<ExtQueryArgs, Operation extends CoreCrudOperations> = (
...
+ * Extracts extended result field names as optional boolean keys for use in select/omit inputs.
+ * When ExtResult is `{}`, this resolves to `{}` (no-op for intersection).
+ */
+export type ExtResultSelectOmitFields<ExtResult extends ExtResultBase, Model extends string> = keyof ExtResult extends never
+    ? {}
+    : Model extends keyof ExtResult
+        ? { [K in keyof ExtResult[Model]]?: boolean }
+        : {};
+
+/**
```
ymc9, line 2466: Should this JSDoc comment be moved to above `SelectAwareExtResult`?

packages/orm/src/client/client-impl.ts
```diff
@@ -823,3 +842,274 @@ function createModelCrudHandler(
...
+    }
+
+    const data = row as Record<string, unknown>;
+    const typedArgs = (originalArgs && typeof originalArgs === 'object' ? originalArgs : {}) as Record<string, unknown>;
+    const select = typedArgs['select'] as Record<string, unknown> | undefined;
+    const omit = typedArgs['omit'] as Record<string, unknown> | undefined;
+    const include = typedArgs['include'] as Record<string, unknown> | undefined;
+
+    // Determine which ext result fields were selected/omitted at this level
+    const selectedExtResultFields = select ? new Set<string>() : undefined;
```
ymc9: I'm wondering if the "clean-up" of injected selections can be simpler. Instead of analyzing "needs", can we simply go through select/include/omit and drop all fields that aren't supposed to be there?

I understand only checking those fields that are possibly injected by "needs" is safer, but just thinking maybe the effect will be the same.

packages/orm/src/client/client-impl.ts
```diff
@@ -823,3 +842,274 @@ function createModelCrudHandler(
...
+
+    // Compute ext result fields for the current model
+    for (const [fieldName, fieldDef] of extResultDefs) {
+        if (omittedExtResultFields?.has(fieldName)) {
+            continue;
+        }
+        if (selectedExtResultFields !== undefined && !selectedExtResultFields.has(fieldName)) {
+            continue;
+        }
+        const needsSatisfied = Object.keys(fieldDef.needs).every((needField) => needField in data);
```
ymc9, line 1072: Just thought of an interesting feature interaction here. If field-level access policy caused a "needed" field to be dropped, the evaluation will be muted here. It's probably a reasonable behavior though.

tests/e2e/orm/plugin-infra/ext-result.test.ts
```diff
@@ -0,0 +1,816 @@
...
+    afterEach(async () => {
+        await db?.$disconnect();
+    });
+
+    it('should compute virtual fields on findMany results', async () => {
+        const extDb = db.$use(
+            definePlugin({
+                id: 'greeting',
+                result: {
+                    User: {
```
ymc9: I'm thinking if we should consistently use "lower-case-first" model name ("User" -> "user") in ZenStackClient and plugin settings, which looks more natural.

Currently the "omit" settings in the ORM client use original model names. I regretted it and I'm planning to make a change there.

sanny-io:
Mostly copying my suggested changes from the other PR:

 All fields should maintain their typing as specified in the schema, but non-dependency fields should be optional, and the computed field in question should be absent.
 
 Consider the following:

 ```prisma
 model User {
     id String @id @default(uuid())
     age Int
     birthDate DateTime
     role Role
     firstName String
     lastName String
     fullName String
 }
 
 enum Role {
     USER
     ADMIN
 }
 ```

```typescript
const db = client.$use(definePlugin({
  id: 'display',
  result: {
    User: {
      fullName: {
        needs: { firstName: true, lastName: true },
        compute: (user) => `${user.firstName} ${user.lastName}`,
      },
    },
  },
}));
```
 
 The `user` parameter when defining the `fullName` computed field should be typed as this:
 
 ```typescript
 {
     id?: string
     age?: number
     role?: 'USER' | 'ADMIN'
     birthDate?: Date
     firstName: string
     lastName: string
 }
 ```

Notice how everything that isn't a dependency is allowed to be undefined, dependencies themselves are not, everything still maintains its proper type, and fullName is absent because it doesn't make sense to be able to access a field you are defining how to access.

@ymc9 @genu thoughts? This is the most accurate typing as far as I can tell. Could be reduced if it's too complex to implement.

packages/orm/src/client/client-impl.ts
genu: Good catch — yes, if a policy strips a needed field, `needsSatisfied` will be false and the virtual field is silently omitted from the result. I think this is the right default: computing with missing data would produce garbage, and throwing would be too disruptive. If we ever want to surface this, a dev-mode warning would be the natural place

packages/orm/src/client/plugin.ts
genu: I updated this to `export type ExtResultFieldDef<Needs extends Record<string, true> = Record<string, true>>`.

Did you mean we should pass `Schema`/`Model` through to it? We can add those params to `ExtResultFieldDef` itself, but we can't use them in `ExtResultBase` — doing so forces TypeScript to resolve `NonRelationFields<Schema, M>` while it's still inferring `ExtResult` in `definePlugin`, which creates a circular inference that collapses `ExtResult` to `{}`.

The `needs` validation already works today via the `infer N` + `keyof N extends NonRelationFields` check in `RuntimePlugin.result`.

For the compute data values (which were typed as any), I’ve already updated `RuntimePlugin.result` to provide a typed contextual type using `MapModelFieldType`, so compute’s data parameter now gets proper field types (e.g., user.name is string, user.id is number) without needing to change `ExtResultFieldDef` or `ExtResultBase`.

Unless you mean something else?

packages/orm/src/client/plugin.ts
genu: since `RuntimePlugin.result` now contextually provides strict typing for both needs keys and compute data values (via MapModelFieldType), this helper may no longer be needed — it actually gives weaker types (any values) than what the interface provides.

The only case where it could still be useful is if someone defines result fields in a separate variable outside of `definePlugin`, where the contextual typing wouldn’t apply. But since that’s an uncommon pattern, maybe we can remove it?

tests/e2e/orm/plugin-infra/ext-result.test.ts
genu: Made the change, however, to make `user:` work as a key instead of `User:`, the types had to change too. If I only changed the tests to use user: without updating `ExtResultBase` and the related types, TypeScript would reject user as an invalid key

genu:
@sanny-io

I think this is reasonable since at runtime the full row is passed to compute anyway — those fields may or may not be present depending on the query’s select. Making them optional accurately reflects that. The exclusion of the field being defined also makes sense to prevent circular references.

One consideration: this adds type-level complexity since we’d need to thread the field name being defined into the compute parameter type and map all model fields. But the DX improvement seems worth it. I'm not exactly sure on the level of effort here as it’s a TypeScript inference puzzle. Could take a few hours of experimentation, or could be a rabbit hole.

genu:
@coderabbitai resume

genu:
@ymc9 I’ve addressed the `needs` validation feedback:

Runtime: `collectExtResultFieldDefs` now validates that all needs keys exist as non-relation scalar fields on the model, with clear error messages.

Types: Updated `ExtResultBase` to constrain needs keys to `NonRelationFields<Schema, Model>`, which improves autocomplete and enables compile-time checking via `satisfies ExtResultBase<SchemaType>`.

One limitation I wasn’t able to solve: TypeScript doesn’t enforce the needs constraint through `definePlugin()` due to how generic inference handles excess properties on nested objects. The runtime validation covers this gap. Would appreciate another review when you get a chance!

ymc9:
> @ymc9 I’ve addressed the `needs` validation feedback:
> 
> Runtime: `collectExtResultFieldDefs` now validates that all needs keys exist as non-relation scalar fields on the model, with clear error messages.
> 
> Types: Updated `ExtResultBase` to constrain needs keys to `NonRelationFields<Schema, Model>`, which improves autocomplete and enables compile-time checking via `satisfies ExtResultBase<SchemaType>`.
> 
> One limitation I wasn’t able to solve: TypeScript doesn’t enforce the needs constraint through `definePlugin()` due to how generic inference handles excess properties on nested objects. The runtime validation covers this gap. Would appreciate another review when you get a chance!

Cool! I'm working on several other fixes, and will make another review pass during the weekend. Thanks @genu !

packages/orm/src/client/plugin.ts
```diff
@@ -20,13 +20,44 @@ export type ExtQueryArgsBase = {
...
+/**
+ * Base shape of plugin-extended result fields.
+ * Keyed by model name, each value maps field names to their definitions.
+ * `needs` keys are constrained to non-relation fields of the corresponding model.
+ */
+export type ExtResultBase<Schema extends SchemaDef = SchemaDef> = {
+    [M in Uncapitalize<GetModels<Schema>>]?: Record<
+        string,
+        {
+            needs: Partial<Record<NonRelationFields<Schema, Capitalize<M> & GetModels<Schema>>, true>>;
```
ymc9: I think using `Capitalize` here can break cases where model names start with lower case in ZModel.

Shall we do `M in GetModels<Schema> as Uncapitalize<M>` and keep using `M` in the `needs` field instead?

packages/orm/src/client/plugin.ts
```diff
@@ -20,13 +20,44 @@ export type ExtQueryArgsBase = {
  */
 export type ExtClientMembersBase = Record<string, unknown>;
 
+/**
+ * Definition for a single extended result field.
+ * When used without type parameters, accepts any field names and untyped compute.
+ */
+export type ExtResultFieldDef<Needs extends Record<string, true> = Record<string, true>> = {
```
ymc9: It seems this type is not used on the public typing side, right? Does it still need to exist? It's currently used in client-impl, and that part of code doesn't need strict typing.

packages/orm/src/client/plugin.ts
```diff
@@ -81,9 +112,15 @@ export interface RuntimePlugin<
      * Extended client members (methods and properties).
      */
     client?: ExtClientMembers;
+
+    /**
+     * Extended result fields on query results.
+     * Keyed by model name, each value defines computed fields with `needs` and `compute`.
+     */
+    result?: ExtResult;
```
ymc9, line 105: It seems when a plugin is defined in line when calling `$use`, the `needs`'s typing is still not properly propagated to `compute`. I can try letting claude ponder more on this after we get this initial implementation committed.

genu:
Thanks @ymc9 I addressed your comments. Feel free to merge this initial implementation as you see fit.

packages/orm/src/client/client-impl.ts
```diff
@@ -823,3 +847,269 @@ function createModelCrudHandler(
...
+        }
+        result = { ...result, select: newSelect };
+        changed = true;
+    }
+
+    if (omit && extResultDefs.size > 0) {
+        const newOmit = { ...omit };
+        for (const [fieldName, fieldDef] of extResultDefs) {
+            if (newOmit[fieldName]) {
+                // strip ext result field names from omit (they don't exist in the DB)
```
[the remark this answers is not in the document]
genu: The select and omit objects are typed as `Record<string, boolean>` at the user-facing level, and by the time they reach this internal code, they’ve already passed Zod validation. 

There’s no realistic scenario where a non-boolean `truthy` value sneaks in.

Issue #2453, opened 2026-03-05
[Feature Request] Nested routes in the RestApiHandler


lsmith77:
This relates to [this discussion thread](https://discord.com/channels/1035538056146595961/1087761218049167380/1460605860492476500) on Discord.

**Is your feature request related to a problem? Please describe.**

RestApiHandler only supports flat routes (/post, /post/42). There is no way to express parent-scoped access in the URL structure. To fetch only a user's posts, clients must manually filter by parent ID — which is error-prone and produces URLs that don't reflect the data hierarchy.

** Describe the solution you'd like **
Add explicit nested route configuration to RestApiHandler that enables hierarchical URL patterns (/user/:id/post, /user/:id/post/:postId) for all five HTTP methods, with automatic parent scoping enforced server-side:

```
new RestApiHandler({
    schema: client.$schema,
    endpoint: 'http://localhost/api',
    nestedRoutes: {
        User: {
            Post: {
                relation: 'author',            // relation field on child pointing to parent
                requireOrphanProtection: true, // optional: enforce safe onDelete at startup
            },
        },
    },
});
```

**Key behaviors:**

* 404 if parent does not exist, or child does not belong to the requested parent.
* POST automatically binds the created child to the parent; returns 400 if the payload also sets that relation.
* Flat routes remain fully functional. Non-configured 3-segment paths fall through to fetchRelated unchanged.
* JSON:API self links in responses use the nested URL.
* `requireOrphanProtection`: true throws at startup if the configured relation does not have onDelete: Cascade, Restrict, or NoAction in the schema.
* All configuration errors (unknown models, missing relation fields, mismatched targets) are caught at construction time.

**Describe alternatives you've considered**

* Implicit route generation from schema relations (aka auto discovery of the route nesting) — rejected because it creates unexpected route surface area and is hard to audit. Explicit-only is safer.
* Client-side filtering via filter[authorId]=u1 on flat routes — works but is not server-enforced and doesn't produce hierarchical URLs.

**Additional context**

* `modelNameMapping` and `externalIdMapping` are both respected on nested routes.
* Multi-level nesting (/a/:id/b/:id/c) could be done later.

ymc9:
Supported in 3.5.0

PR #2454, opened 2026-03-05
feat(server): add nested routes support in the RestApiHandler

lsmith77:
fixes https://github.com/zenstackhq/zenstack/issues/2453

Note: this was largely AI developed. I wanted to offer it as a more concrete proposal than just the above referenced ticket. If you agree this feature has merit, I will do a more detailed analysis of the proposed code changes.

# TOCTOU Race Condition in Nested Mutations

TOCTOU = Time-of-Check to Time-of-Use

The nested update/delete operations verify child-parent membership using a merged filter, then perform the mutation using only the child ID. In a concurrent scenario where the parent relationship changes between check and mutation, the operation could target an orphaned or reassigned resource.

## Options Explored:

* Direct mutations with merged filter (ATTEMPTED, FAILED): Have processNestedUpdate/processNestedDelete build and execute mutations directly using the merged filter for atomicity.
  * ❌ Result: Passing the merged filter to the ORM's update() and delete() calls resulted in 422 (Unprocessable Entity) errors. Not viable with current ORM implementation.
* Optional where parameter (NOT ATTEMPTED): Add optional where parameter to processUpdate/processDelete to allow merged filter injection from nested handlers.
  * ⚠️ Trade-off: Would modify widely-used public methods, affecting all CRUD operations globally—higher risk for unintended consequences across the codebase. Still theoretically viable but rejected for minimal impact principle.
* Format adaptation (NOT ATTEMPTED): Investigate merged filter semantics and adapt the format for direct ORM use.
  * ⚠️ Trade-off: Would require deep ORM integration work and understanding internal filter representations. Still theoretically viable but high effort/risk ratio.

## Why Not Fixed:

* Option 1 definitively failed due to ORM incompatibility
* Options 2-3 rejected to keep PR scope minimal and reduce cross-cutting impact
* Can't easily expose the race condition in tests (requires concurrent modifications)

## Mitigation:
The authorization check prevents unauthorized cross-parent access (the primary security concern). The window for this race is extremely narrow and requires concurrent parent-relationship changes—a rare operation pattern.

## Summary by CodeRabbit

* **New Features**
  * Configurable nested REST routes for parent-child resources: scoped reads/writes, nested single resource URLs, path-aware link generation, and enforced parent binding on nested creates/updates.

* **Bug Fixes**
  * Fixed range/filter parsing issue.

* **Tests**
  * Added extensive tests covering nested-route validation, orphan-protection semantics, routing behavior, mappings, and CRUD operations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "zenstack-v3",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "description": "ZenStack",
     "packageManager": "pnpm@10.23.0",
     "type": "module",
--- a/packages/auth-adapters/better-auth/package.json
+++ b/packages/auth-adapters/better-auth/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@zenstackhq/better-auth",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "description": "ZenStack Better Auth Adapter. This adapter is modified from better-auth's Prisma adapter.",
     "type": "module",
     "scripts": {
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -1,0 +1,24 @@
+# @zenstackhq/cli
+
+The command-line interface for ZenStack. Provides commands for initializing projects, generating TypeScript code from ZModel schemas, managing database migrations, and etc.
+
+## Key Commands
+
+- `zenstack init` — Initialize ZenStack in an existing project
+- `zenstack generate` — Compile ZModel schema to TypeScript
+- `zenstack db push` — Sync schema to the database
+- `zenstack db pull` — Pull database schema changes into ZModel
+- `zenstack migrate dev` — Create and apply database migrations
+- `zenstack migrate deploy` — Deploy migrations to production
+- `zenstack format` — Format ZModel schema files
+- `zenstack proxy|studio` — Start a database proxy server for using studio
+
+## Installation
+
+```bash
+npm install -D @zenstackhq/cli
+```
+
+## Learn More
+
+- [ZenStack Documentation](https://zenstack.dev/docs)
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -3,7 +3,7 @@
     "publisher": "zenstack",
     "displayName": "ZenStack CLI",
     "description": "FullStack database toolkit with built-in access control and automatic API generation.",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "type": "module",
     "author": {
         "name": "ZenStack Team"
--- a/packages/clients/client-helpers/package.json
+++ b/packages/clients/client-helpers/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@zenstackhq/client-helpers",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "description": "Helpers for implementing clients that consume ZenStack's CRUD service",
     "type": "module",
     "scripts": {
--- a/packages/clients/tanstack-query/package.json
+++ b/packages/clients/tanstack-query/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@zenstackhq/tanstack-query",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "description": "TanStack Query Client for consuming ZenStack v3's CRUD service",
     "type": "module",
     "scripts": {
--- a/packages/common-helpers/README.md
+++ b/packages/common-helpers/README.md
@@ -1,0 +1,3 @@
+# @zenstackhq/common-helpers
+
+A collection of small, general-purpose utility functions shared across ZenStack packages.
--- a/packages/common-helpers/package.json
+++ b/packages/common-helpers/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@zenstackhq/common-helpers",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "description": "ZenStack Common Helpers",
     "type": "module",
     "scripts": {
--- a/packages/config/eslint-config/package.json
+++ b/packages/config/eslint-config/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@zenstackhq/eslint-config",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "type": "module",
     "private": true,
     "license": "MIT"
--- a/packages/config/typescript-config/package.json
+++ b/packages/config/typescript-config/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@zenstackhq/typescript-config",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "private": true,
     "license": "MIT"
 }
--- a/packages/config/vitest-config/package.json
+++ b/packages/config/vitest-config/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@zenstackhq/vitest-config",
     "type": "module",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "private": true,
     "license": "MIT",
     "exports": {
--- a/packages/create-zenstack/README.md
+++ b/packages/create-zenstack/README.md
@@ -1,0 +1,13 @@
+# create-zenstack
+
+Scaffolding tool to create a new ZenStack project from a template with ZenStack packages installed and a sample ZModel schema generated.
+
+## Usage
+
+```bash
+npm create zenstack@latest my-app
+```
+
+## Learn More
+
+- [ZenStack Documentation](https://zenstack.dev/docs)
--- a/packages/create-zenstack/package.json
+++ b/packages/create-zenstack/package.json
@@ -1,6 +1,6 @@
 {
     "name": "create-zenstack",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "description": "Create a new ZenStack project",
     "type": "module",
     "scripts": {
--- a/packages/ide/vscode/package.json
+++ b/packages/ide/vscode/package.json
@@ -1,7 +1,7 @@
 {
     "name": "zenstack-v3",
     "publisher": "zenstack",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "displayName": "ZenStack V3 Language Tools",
     "description": "VSCode extension for ZenStack (v3) ZModel language",
     "private": true,
--- a/packages/ide/vscode/scripts/post-build.ts
+++ b/packages/ide/vscode/scripts/post-build.ts
@@ -6,13 +6,8 @@ dotenv.config({ path: './.env' });
 
 const telemetryToken = process.env.VSCODE_TELEMETRY_TRACKING_TOKEN;
 if (!telemetryToken) {
-    if (process.env.CI) {
-        console.error('Error: VSCODE_TELEMETRY_TRACKING_TOKEN environment variable is not set');
-        process.exit(1);
-    } else {
-        console.warn('Warning: VSCODE_TELEMETRY_TRACKING_TOKEN environment variable is not set, skipping token injection');
-        process.exit(0);
-    }
+    console.warn('Warning: VSCODE_TELEMETRY_TRACKING_TOKEN environment variable is not set, skipping token injection');
+    process.exit(0);
 }
 const file = 'dist/extension.js';
 let content = fs.readFileSync(file, 'utf-8');
--- a/packages/language/README.md
+++ b/packages/language/README.md
@@ -1,0 +1,7 @@
+# @zenstackhq/language
+
+The ZModel language specification and tooling package, built on [Langium](https://langium.org/).
+
+## Learn More
+
+- [ZModel Language Reference](https://zenstack.dev/docs/reference/zmodel)
--- a/packages/language/package.json
+++ b/packages/language/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@zenstackhq/language",
     "description": "ZenStack ZModel language specification",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "license": "MIT",
     "author": "ZenStack Team",
     "files": [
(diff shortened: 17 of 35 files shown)


ymc9:
Hi @lsmith77 , thanks for making this PR and my apologies for the late follow up.

I can understand it's problematic that resource mutations cannot be scoped (atomically) to a parent. Since it's not part of JSON:API spec, we'll need to deviate and make an extension.

For the URL design, I'm wondering if, instead of having a set of parallel URL patterns for scoped updates, it's simpler to just extend the current ones. We already have today:

- `:type/:id` for single resource
- `:type/:id/:relationship` for relationship data
- `:type/id/relationships/:relationship` for relationship

Can we just extend the second one to allow:
1. POST and DELETE operations (currently the URL is read-only)
    `POST user/1/posts`
    `DELETE user/1/profile`
2. A fourth segment for locating one resource in a one-to-many relation, and allow PATCH/DELETE
    `PATCH user/1/posts/1`
    `DELETE user/1/posts/1`

lsmith77:
Thanks for the feedback! I've reworked the implementation to follow your suggestion. The nested route URLs now extend the existing :type/:id/:relationship pattern rather than introducing a parallel set:

* GET/POST /:type/:id/:relationship — scoped collection read and create (unchanged behavior, same URL)
* DELETE /:type/:id/:relationship — new: deletes the related resource for a to-one relation (e.g. DELETE user/1/profile)
* GET/PATCH/DELETE /:type/:id/:relationship/:childId — scoped single-resource operations, replacing the old /:parentType/:parentId/:childType/:childId pattern

The config also got simplified: the second-level key is now the relation field name on the parent (e.g. posts on User) instead of the child model name, and the explicit relation field is dropped since the reverse relation is derived automatically from the schema's opposite field. So `{ User: { Post: { relation: 'author' } } }` becomes `{ User: { posts: {} } }`.

Three things worth flagging before this is finalised:

1. Config key consistency — the first level of nestedRoutes uses a model name while the second level uses a relation field name, which is a bit inconsistent. An alternative would be to flatten to dotted notation (`{ 'User.posts': {} }`), which is more uniform and scales better with many parent models. Happy to change this if you'd prefer, but wanted to raise it before altering another public API shape.

2. Atomicity of updates and deletes — creates are genuinely atomic (the parent connect is part of a single INSERT). DELETEs have been fixed: both DELETE /:type/:id/:relationship/:childId and DELETE /:type/:id/:relationship now use deleteMany with the merged parent+child filter, so the scope check and the write are a single atomic operation. PATCH however still has a TOCTOU: the parent-scope check (findFirst) and the actual write (update) are separate round-trips, because the ORM's update() requires a unique WHERE clause and rejects the merged filter with a 422. Fixing it atomically would require bypassing processUpdate and duplicating its payload-building and serialization logic, which is out of scope for this PR. Leaving this as a known limitation; happy to revisit in a follow-up.

3. Access control — nested routes add the parent-scope filter on top of whatever ORM client policies are in place. There is no bypass risk, but if no policies are configured, the parent-scope check is the only gate, so applications relying on this feature for access control should be aware that it scopes by parent ownership only, not by the identity of the calling user.

packages/server/src/api/rest/index.ts
```diff
@@ -64,6 +64,28 @@ export type RestApiHandlerOptions<Schema extends SchemaDef = SchemaDef> = {
...
     externalIdMapping?: Record<string, string>;
+
+    /**
+     * Explicit nested route configuration.
+     *
+     * First-level keys are parent model names, second-level keys are relation field names on the parent model
+     * (e.g., `posts` for `User.posts`). This matches the URL segment used in nested routes:
+     * `/:parentType/:parentId/:relationName` and `/:parentType/:parentId/:relationName/:childId`.
+     */
+    nestedRoutes?: Record<
```
ymc9: Maybe we can make the parent-scoped handling always enabled, and don't require a configuration? What do you think?

packages/server/src/api/rest/index.ts
```diff
@@ -335,6 +412,74 @@ export class RestApiHandler<Schema extends SchemaDef = SchemaDef> implements Api
...
+
+        const { reverseRelation } = resolved;
+        const childInfo = this.getModelInfo(resolved.childType);
+        if (!childInfo) {
+            return { filter: undefined, error: this.makeUnsupportedModelError(resolved.childType) };
+        }
+
+        const reverseRelInfo = childInfo.relationships[reverseRelation];
+        const relationFilter = reverseRelInfo?.isCollection
+            ? { [reverseRelation]: { some: this.makeIdFilter(parentInfo.idFields, parentId, false) } }
```
ymc9, line 418: I'm wondering if it's cleaner to limit the parent-scoped read to to-many only. For to-one nested read, you can already do it today like `GET /post/1/author` without requiring a forth segment on the URL.

packages/server/src/api/rest/index.ts
```diff
@@ -375,18 +520,23 @@ export class RestApiHandler<Schema extends SchemaDef = SchemaDef> implements Api
                 case 'GET': {
                     let match = this.matchUrlPattern(path, UrlPatterns.SINGLE);
                     if (match) {
-                        // single resource read
                         return await this.processSingleRead(client, match.type, match.id, query);
                     }
                     match = this.matchUrlPattern(path, UrlPatterns.FETCH_RELATIONSHIP);
                     if (match) {
-                        // fetch related resource(s)
+                        if (this.getNestedRouteConfig(match.type, match.relationship)) {
```
ymc9: I think this branch is not needed,  as it's for fetching a nested resource, scoped by parent, without specifying its own ID. It should already work today.

packages/server/src/api/rest/index.ts
```diff
@@ -395,35 +545,52 @@ export class RestApiHandler<Schema extends SchemaDef = SchemaDef> implements Api
...
+                        if (this.upsertMetaSchema.safeParse(body).success) {
                             return await this.processUpsert(client, match.type, query, requestBody);
                         } else {
-                            // resource creation
                             return await this.processCreate(client, match.type, query, requestBody);
                         }
                     }
                     match = this.matchUrlPattern(path, UrlPatterns.RELATIONSHIP);
                     if (match) {
-                        // relationship creation (collection relationship only)
```
ymc9, line 429: It seems some comments unrelated to this change were removed.

packages/server/src/api/rest/index.ts
```diff
@@ -598,13 +794,15 @@ export class RestApiHandler<Schema extends SchemaDef = SchemaDef> implements Api
         type: string,
         resourceId: string,
         query: Record<string, string | string[]> | undefined,
+        extraFilter?: any,
```
ymc9: These two new parameters never passed?

packages/server/src/api/rest/index.ts
```diff
@@ -639,9 +837,18 @@ export class RestApiHandler<Schema extends SchemaDef = SchemaDef> implements Api
         const entity = await (client as any)[type].findUnique(args);
 
         if (entity) {
+            const serializeOptions: Partial<SerializerOptions> = {
+                include,
+            };
+            if (pathOverride) {
```
ymc9: What is this branch for?

packages/server/src/api/rest/index.ts
```diff
@@ -905,6 +1127,255 @@ export class RestApiHandler<Schema extends SchemaDef = SchemaDef> implements Api
...
+
+        const childInfo = this.getModelInfo(resolved.childType)!;
+        const collectionPath = this.makeNestedLinkUrl(parentType, parentId, parentRelation);
+        const resourceLinker = new tsjapi.Linker((item: any) =>
+            this.makeLinkUrl(`${collectionPath}/${this.getId(childInfo.name, item)}`),
+        );
+        return this.processCollectionRead(client, resolved.childType, query, filter, collectionPath, resourceLinker);
+    }
+
+    private async processNestedSingleRead(
```
ymc9, line 1109: This method seems to have quite a few duplicates with `processSingleRead`. Shall we extract a shared helper method, or maybe even unify them?

packages/server/src/api/rest/index.ts
```diff
@@ -905,6 +1127,255 @@ export class RestApiHandler<Schema extends SchemaDef = SchemaDef> implements Api
...
+        query: Record<string, string | string[]> | undefined,
+        requestBody: unknown,
+    ): Promise<Response> {
+        const resolved = this.resolveNestedRelation(parentType, parentRelation);
+        if (!resolved) {
+            return this.makeError('invalidPath');
+        }
+
+        const parentInfo = this.getModelInfo(parentType)!;
+        const parentExists = await (client as any)[parentType].findFirst({
```
ymc9: There's no need to check parent existence. Nested create should actually be done with a parent update. 

For a query `POST user/1/posts`, the call to ORM should look like:

```
client.user.update({
  where: { id: 1 },
  data: {
    posts: { create: ... }
  }
}
```

If the parent doesn't exist, the ORM will throw a not-found error, and the check&create is atomic.

packages/server/src/api/rest/index.ts
```diff
@@ -905,6 +1127,255 @@ export class RestApiHandler<Schema extends SchemaDef = SchemaDef> implements Api
...
+        }
+
+        const childType = resolved.childType;
+        const typeInfo = this.getModelInfo(childType)!;
+
+        const { filter: nestedFilter, error } = this.buildNestedParentFilter(parentType, parentId, parentRelation);
+        if (error) return error;
+
+        // Verify child belongs to parent before updating
+        const exists = await (client as any)[childType].findFirst({
```
ymc9: Similar to the previous comment, nested update should be done with an ORM `update` call on the parent entity.

packages/server/src/api/rest/index.ts
```diff
@@ -905,6 +1127,255 @@ export class RestApiHandler<Schema extends SchemaDef = SchemaDef> implements Api
...
+            return this.makeError('invalidPath');
+        }
+
+        const childType = resolved.childType;
+        const typeInfo = this.getModelInfo(childType)!;
+
+        const { filter: nestedFilter, error } = this.buildNestedParentFilter(parentType, parentId, parentRelation);
+        if (error) return error;
+
+        // Atomic: delete only if the child still belongs to the parent, eliminating TOCTOU.
```
ymc9: Same for delete.

packages/server/src/api/rest/index.ts
lsmith77: So we would always offer both the flat and nested API? For me that is fine. Just wonder if someone with super high load would have concerns over the TOCTOU race conditions. But this could be handled by a single global config option.

packages/server/src/api/rest/index.ts
lsmith77: Unifying into one method would require threading optional nestedFilter and linkers params through the signature, making processSingleRead harder to read for no benefit to its own callers. So a common helper seems like the better direction.

lsmith77: I am currently doing another review of the core changes, which I will push soon. I will then add this refactoring into a separate follow-up commit into this PR.

packages/server/src/api/rest/index.ts
lsmith77: Removed

packages/server/src/api/rest/index.ts
lsmith77: I restored these comments and reverted some other unrelated changes

packages/server/src/api/rest/index.ts
lsmith77: Removed

packages/server/src/api/rest/index.ts
lsmith77: Leftover from an earlier iteration

packages/server/src/api/rest/index.ts
lsmith77: Done

packages/server/src/api/rest/index.ts
lsmith77: Done

packages/server/src/api/rest/index.ts
lsmith77: Done

packages/server/src/api/rest/index.ts
lsmith77: I have pushed a refactoring to use a single boolean flag to enable/disable nestedRoutes

lsmith77:
@ymc9 I have hopefully addressed all of your feedback. The PR now contains 3 commits:

1) the core functionality
2) a small refactoring to extract a new common helper for processSingleRead and processNestedSingleRead
3) a simplification of the configuration introduced in 1) to a single boolean flag

So let me know, if you don't want either 2) or 3), or do not even want any flag to enable/disable this feature and I will update the PR accordingly. Otherwise you can squash merge to get it all into a single commit, depending on your preference.

packages/server/src/api/rest/index.ts
```diff
@@ -462,18 +565,34 @@ export class RestApiHandler<Schema extends SchemaDef = SchemaDef> implements Api
                             requestBody,
                         );
                     }
-
+                    // /:type/:id/:relationship/:childId — nested update
```
ymc9: For nested update of a to-one relation, do you think we should support "PATCH /:type/:id/:relationship" too? Like:

`PATCH /post/1/author`

Note that you even if you pass author's id, the ORM doesn't support a where filter like:

```ts
db.post.update({
  where: { ... },
  data: { author: { // you can't put "where" here } }
});
```

So I believe with current implementation it'll fail even if you explicitly put author id there like:

`PATCH /post/1/author/1`

lsmith77: I have added support. Should it return 400 or 404 when nested routes are not enabled?

lsmith77:
rebased on `dev` to fix the merge conflicts.

packages/server/src/api/rest/index.ts
ymc9: Thanks! I believe currently 400 is returned and I think it's fine.

Btw, could you sync with the latest dev branch? The build error is due to some recent changes and it has been resolved.

PR #2487, opened 2026-03-15
fix: a few small fixes

elliots:
I'm importing schemas across packages, came across a few issues as I went. Happy to split it out if you prefer.

 - fix: load plugins with jiti: I wasnt able to load a plugin from a workspace package
 - createOpposite option on @ relation: im importing schema files across packages, I'd like a one way import.
 - fix: resolve plugin paths against the schema file where the plugin is: previously it was loading relative to the main schema, not the one where the plugin is actually defined.
 - fix: add `migrate diff` command: was missing, was it meant to be?

## Summary by CodeRabbit

* **New Features**
  * Plugins can now be resolved using bare package specifiers (e.g., workspace packages) in addition to file paths.
  * Plugin paths are now correctly resolved relative to the schema file where they are declared.

* **Tests**
  * Added test coverage for bare package specifier plugin loading and relative plugin path resolution.

### DIFF
--- a/packages/ide/vscode/scripts/post-build.ts
+++ b/packages/ide/vscode/scripts/post-build.ts
@@ -6,13 +6,8 @@ dotenv.config({ path: './.env' });
 
 const telemetryToken = process.env.VSCODE_TELEMETRY_TRACKING_TOKEN;
 if (!telemetryToken) {
-    if (process.env.CI) {
-        console.error('Error: VSCODE_TELEMETRY_TRACKING_TOKEN environment variable is not set');
-        process.exit(1);
-    } else {
-        console.warn('Warning: VSCODE_TELEMETRY_TRACKING_TOKEN environment variable is not set, skipping token injection');
-        process.exit(0);
-    }
+    console.warn('Warning: VSCODE_TELEMETRY_TRACKING_TOKEN environment variable is not set, skipping token injection');
+    process.exit(0);
 }
 const file = 'dist/extension.js';
 let content = fs.readFileSync(file, 'utf-8');


ymc9:
Hi @elliots , thanks for making this PR!

I have a few general thoughts:

1. About migrate diff

    Do you use it frequently? I can't remember exactly, but maybe I intentionally left it out when working on v3. This command feels too much coupled with Prisma Migrate - which may be replaced by a zenstack-home-made migration engine in the future. Also, to wrap around `prisma migrate diff`, when specifying a ZModel as output, I believe we also need to handle the prisma-generated schema and merge it into ZModel. Can be a bit tricky and fragile.

2. About one-direction relation

    This has been brought up quite a few times back in v2, and I think it's a very useful feature to have. Besides automatically filling the opposite side of relation on prisma schema level, I'm not sure if anything needs to be done on zenstack tool chain yet - the ORM API, zod schemas, tanstack hooks, etc.
    
    Shall we spin it out into a separate PR? It's more like a new feature.

ymc9:
> Hi @elliots , thanks for making this PR!
> 
> I have a few general thoughts:
> 
> 1. About migrate diff
>    Do you use it frequently? I can't remember exactly, but maybe I intentionally left it out when working on v3. This command feels too much coupled with Prisma Migrate - which may be replaced by a zenstack-home-made migration engine in the future. Also, to wrap around `prisma migrate diff`, when specifying a ZModel as output, I believe we also need to handle the prisma-generated schema and merge it into ZModel. Can be a bit tricky and fragile.
> 2. About one-direction relation
>    This has been brought up quite a few times back in v2, and I think it's a very useful feature to have. Besides automatically filling the opposite side of relation on prisma schema level, I'm not sure if anything needs to be done on zenstack tool chain yet - the ORM API, zod schemas, tanstack hooks, etc.
>    Shall we spin it out into a separate PR? It's more like a new feature.

Hi @elliots , since I haven't heard back from you about these comments, I've pushed a change to address them and will merge this PR once CI passes. Love to get the other part of the changes included in v3.5.

PR #2490, opened 2026-03-17
feat(clients): add ExtResult support to TanStack Query hooks

genu:
Add type inference helpers to client-helpers enabling computed fields from plugins to be reflected in result types across React, Vue, and Svelte adapters.

**Changes:**
- Add `InferSchema`, `InferOptions`, and `InferExtResult` type utilities to `client-helpers`
- Update `useClientQueries` to accept a `ClientContract` type parameter instead of requiring separate schema + options generics
- Simplify API by inferring schema, options, and ext-result types automatically
- Update React, Vue, and Svelte sliced-client tests to use the new API surface

This enables better type inference when using plugins that extend query results with computed fields.

## Summary by CodeRabbit

* **New Features**
  * Propagates ExtResult-style extensions across React, Vue, and Svelte query clients and re-exports schema/options inference helpers.
  * Adds public types for ORM write actions, query error/info shapes, and cached query metadata.

* **Tests**
  * Simplified sliced-client tests to derive client typings from instantiated clients rather than large inline generics.

### DIFF
--- a/packages/clients/tanstack-query/src/svelte/index.svelte.ts
+++ b/packages/clients/tanstack-query/src/svelte/index.svelte.ts
@@ -19,6 +19,7 @@ import {
     createInvalidator,
     createOptimisticUpdater,
     DEFAULT_QUERY_ENDPOINT,
+    type InferExtResult,
     type InferOptions,
     type InferSchema,
     type InvalidationPredicate,
@@ -38,6 +39,7 @@ import type {
     DeleteArgs,
     DeleteManyArgs,
     ExistsArgs,
+    ExtResultBase,
     FindFirstArgs,
     FindManyArgs,
     FindUniqueArgs,
@@ -72,7 +74,7 @@ import type {
     WithOptimistic,
 } from '../common/types.js';
 export type { FetchFn } from '@zenstackhq/client-helpers/fetch';
-export type { InferOptions, InferSchema } from '@zenstackhq/client-helpers';
+export type { InferExtResult, InferOptions, InferSchema } from '@zenstackhq/client-helpers';
 export type { SchemaDef } from '@zenstackhq/schema';
 
 type ProcedureHookFn<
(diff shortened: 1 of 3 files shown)


packages/clients/tanstack-query/src/svelte/index.svelte.ts
```diff
@@ -283,21 +296,36 @@ export type ModelQueryHooks<
...
+ *
+ * // With server client type for computed field support
+ * import type { DbType } from '~/server/db'
+ * const client = useClientQueries<DbType>(schema)
+ * ```
  */
-export function useClientQueries<Schema extends SchemaDef, Options extends QueryOptions<Schema> = QueryOptions<Schema>>(
-    schema: Schema,
+export function useClientQueries<
+    Client extends SchemaDef | ClientContract<any, any, any, any, any>,
```
ymc9: Shall we name this parameter `SchemaOrClient`

packages/clients/tanstack-query/src/svelte/index.svelte.ts
```diff
@@ -139,15 +146,20 @@ export type ModelMutationModelResult<
     TArgs,
     Array extends boolean = false,
     Options extends QueryOptions<Schema> = QueryOptions<Schema>,
-> = Omit<ModelMutationResult<SimplifiedResult<Schema, Model, TArgs, Options, false, Array>, TArgs>, 'mutateAsync'> & {
+    ExtResult extends ExtResultBase<Schema> = {},
```
ymc9, line 149: Hooks are not subject to result extension, so maybe we remove this parameter and pass `{}` down?

packages/clients/tanstack-query/src/svelte/index.svelte.ts
genu: Done — renamed to `SchemaOrClient` across all three implementations (React, Vue, Svelte).

packages/clients/tanstack-query/src/svelte/index.svelte.ts
genu: Done — removed the `ExtResult` type parameter from `ClientHooks`, `ModelQueryHooks`, `ModelMutationModelResult`, and `useModelQueries`. Now passing `{}` directly everywhere. Also cleaned up the unused `ExtResultBase` and `InferExtResult` imports.

ymc9: Hi @genu , my apologies that I didn't make it clear in the previous review comment here. I wanted to say that "custom procedure hooks" are not subject to result extension. Other regular hooks should carry such typing as you did previously ... sorry for the trouble

genu: No worries! Restored `ExtResult` threading for regular model hooks. Let me know if I missed anything.

Issue #2493, opened 2026-03-18
$setAuth() on hook client breaks no-hook-trigger guarantee, causing infinite recursion


docloulou:
Describe the bug
The documentation for entity mutation hooks states:

`mutation operations initiated with this client will not trigger the entity mutation hooks again
`
However, calling client.$setAuth(...) inside afterEntityMutation creates a new ZenStackClient instance (due to immutability) that loses the no-hook-trigger flag. Any mutation performed through this derived client re-triggers the entity mutation hooks, leading to infinite recursion.
To Reproduce
```
import { definePlugin } from '@zenstackhq/orm'

export const myPlugin = definePlugin({
  id: 'my-plugin',
  onEntityMutation: {
    runAfterMutationWithinTransaction: true,

    async beforeEntityMutation({ model, action, queryId, loadBeforeMutationEntities }) {
      if (model !== 'Order') return
      if (action !== 'update') return
      const entities = (await loadBeforeMutationEntities()) || []
      cache.set(queryId, entities)
    },

    async afterEntityMutation({ model, action, queryId, loadAfterMutationEntities, client }) {
      if (model !== 'Order') return
      if (action !== 'update') return

      const afterEntities = (await loadAfterMutationEntities()) || []
      const beforeEntities = cache.get(queryId) || []

      for (const after of afterEntities) {
        const before = beforeEntities.find(b => b.id === after.id)
        if (!before || after.amount === before.amount) continue

        // WORKS: using client directly (no-hook-trigger preserved)
        await client.order.update({
          where: { id: after.id },
          data: { status: after.amount === 0 ? 'completed' : 'active' },
        })

        // BUG: using $setAuth creates a NEW client that re-triggers hooks
        const adminClient = client.$setAuth({ role: 'admin' })
        await adminClient.order.update({
          where: { id: after.id },
          data: { status: after.amount === 0 ? 'completed' : 'active' },
        })
        // ^ This triggers beforeEntityMutation + afterEntityMutation again
        // causing infinite recursion
      }
    },
  },
})

const cache = new Map<any, any[]>()
```

Expected behavior
`client.$setAuth(...) ` should return a derived client that preserves the no-hook-trigger behavior of the original hook client. The auth context should change, but the hook suppression should be inherited.
Actual behavior
` client.$setAuth(...) `  returns a new ZenStackClient instance that re-triggers entity mutation hooks, causing infinite recursion when the hook performs mutations on the same model it listens to.

[...]

PR #2497, opened 2026-03-19
fix: preserve transaction state in $use, $unuse, and $unuseAll

### LINKED ISSUE #2494: $unuseAll() and $unuse() break transaction isolation
## Description

Calling `$unuseAll()` or `$unuse()` on a transaction client returns a new client that runs queries **outside** the active transaction.

## Reproduction

```ts
const client = new ZenStackClient(schema, {
  dialect: new PostgresDialect({ pool }),
  plugins: [new PolicyPlugin()],
});

await client.$transaction(async (tx) => {
  // Get txid inside the transaction
  const [{ txid: txTxid }] = await tx.$unuseAll().$queryRawUnsafe<{ txid: string }>(
    'SELECT txid_current() AS txid'
  );

  // Get txid via $unuseAll()
  const [{ txid: unuseTxid }] = await tx.$unuseAll().$queryRawUnsafe<{ txid: string }>(
    'SELECT txid_current() AS txid'
  );

  console.log('tx txid:', txTxid);
  console.log('$unuseAll() txid:', unuseTxid);
  // These are DIFFERENT — $unuseAll() escaped the transaction
});
```

Output:
```
tx txid: 17474959
$unuseAll() txid: 17474960
```

## Root cause

In `client-impl.ts`, `$unuseAll()` creates a new `ClientImpl` via `new ClientImpl(this.schema, newOptions, this)`. The constructor always runs `this.kysely = new Kysely(this.kyselyProps)` (line 131), which creates a fresh Kysely instance from `kyselyProps`.

However, when inside a transaction, the transaction state lives in `txClient.kysely` (set to the Kysely `Transaction` object on line 257 of `interactiveTransaction`). The `kyselyProps` don't carry the transaction context, so the new client gets a non-transactional connection.

The same issue applies to `$unuse()` and `$use()`.

## Suggested fix

After constructing the new client, propagate the transaction Kysely instance:

```ts
$unuseAll() {
    const newOptions = { ...this.options, plugins: [] };
    const newClient = new ClientImpl(this.schema, newOptions, this);
    newClient.inputValidator = new InputValidator(newClient, {
        enabled: newOptions.validateInput !== false,
    });
    // Preserve transaction state
    if (this.kysely.isTransaction) {
        newClient.kysely = this.kysely;
    }
    return newClient;
}
```

Same pattern for `$unuse()`, `$use()`, and `$setAuth()`.

## Version

`@zenstackhq/orm` 3.4.6

pkudinov:
Fixes #2494

## Problem

Calling `$use()`, `$unuse()`, or `$unuseAll()` on a transaction client returns a new client that runs queries **outside** the active transaction.

The constructor always creates a fresh Kysely instance from `kyselyProps` (`this.kysely = new Kysely(this.kyselyProps)`), but the transaction state lives in `txClient.kysely` which was set to the Kysely `Transaction` object after construction. The new client never receives this.

## Fix

After constructing the new client in `$use()`, `$unuse()`, and `$unuseAll()`, propagate the transaction Kysely instance:

```ts
if (this.kysely.isTransaction) {
    newClient.kysely = this.kysely;
}
```

## Tests

Three new tests in `tests/e2e/orm/client-api/transaction.test.ts` verify that `$unuseAll()`, `$unuse()`, and `$use()` on a transaction client stay in the same transaction by checking that writes roll back when the transaction fails.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures cloned/derived clients preserve the active transaction context so transactional operations behave correctly.

* **Tests**
  * Added end-to-end tests verifying transaction isolation and rollback when using client helper methods inside transactions.

### DIFF
--- a/packages/orm/src/client/client-impl.ts
+++ b/packages/orm/src/client/client-impl.ts
@@ -128,10 +128,14 @@ export class ClientImpl {
             });
         }
 
-        this.kysely = new Kysely(this.kyselyProps);
-        if (!executor && baseClient?.isTransaction) {
+        if (baseClient?.isTransaction && !executor) {
+            // if we're creating a derived client from a transaction client and not replacing
+            // the executor, reuse the current kysely instance to retain the transaction context
             this.kysely = baseClient.$qb;
+        } else {
+            this.kysely = new Kysely(this.kyselyProps);
         }
+
         this.inputValidator =
             baseClient?.inputValidator ??
             new InputValidator(this as any, { enabled: this.$options.validateInput !== false });
@@ -959,12 +963,7 @@ function collectExtResultFieldDefs(
  * - Injects `needs` fields into `select` when ext result fields are explicitly selected
  * - Recurses into `include` and `select` for nested relation fields
  */
-function prepareArgsForExtResult(
-    args: unknown,
-    model: string,
-    schema: SchemaDef,
-    plugins: AnyPlugin[],
-): unknown {
+function prepareArgsForExtResult(args: unknown, model: string, schema: SchemaDef, plugins: AnyPlugin[]): unknown {
     if (!args || typeof args !== 'object') {
         return args;
     }


packages/orm/src/client/client-impl.ts
```diff
@@ -379,6 +378,10 @@ export class ClientImpl {
         newClient.inputValidator = new InputValidator(newClient as any, {
             enabled: newOptions.validateInput !== false,
         });
+        // preserve transaction state so the new client stays in the same transaction
+        if (this.kysely.isTransaction) {
```
ymc9: Thanks @pkudinov , this is an important fix!

There're a few other places that need the same care: `$setAuth`, `$setOptions`. I think we can consolidate the change into the constructor. When `baseClient` is passed and `baseClient.kysely` is a transaction, we directly reuse it in the newly constructed client. What do you think?

pkudinov: Implemented in `abeac6fb`.

I moved the transaction reuse into the constructor for the non-executor clone path, so `$setAuth` and `$setOptions` are covered automatically alongside `$use`/`$unuse`/`$unuseAll`. The working version was to reuse the transaction through the public accessors (`baseClient.isTransaction` / `baseClient.$qb`) rather than reading private state directly.

I also extended the e2e coverage to include `$setAuth` and `$setOptions`, and updated the `$use` test to use a real `onQuery` hook.

Please let me know what you think @ymc9

ymc9: Looks good. I've made a local refactor and pushed a change. Will merge once CI passes.

docloulou:
This fix #2493 ? 
´´´
client.$setAuth(...) returns a new ZenStackClient instance that re-triggers entity mutation hooks, causing infinite recursion when the hook performs mutations on the same model it listens to.
´´´

PR #2500, opened 2026-03-19
fix(zod): properly infer scalar array types

### LINKED ISSUE #2499: `@zenstackhq/zod` scalar array types are incorrect
A clear and concise description of what the bug is and what's the expected behavior.

The TypeScript definition for inferring a Zod schema type from a Zen schema is incorrect for scalar arrays like `String[]`. The array-ness is lost, and the Zod types report `z.ZodString` instead of `z.ZodArray<z.ZodString>`.

This seems to be due to a missing `ZodArrayIf` in the scalar branch here:

https://github.com/zenstackhq/zenstack/blob/00768de0cc18bc2f5b0919b62902c841bdc8362c/packages/zod/src/types.ts#L22-L29

This means types do not match runtime, where the Zod schema is actually correct, because `applyCardinality` is used properly in `makeScalarFieldSchema` here:

https://github.com/zenstackhq/zenstack/blob/00768de0cc18bc2f5b0919b62902c841bdc8362c/packages/zod/src/factory.ts#L174

**Screenshots**

-   ZenStack version: 3.4.6 with `@zenstackhq/zod@3.4.6`
-   Database type: any
-   Node.js/Bun version: any
-   Package manager: any

**Additional context**

haltcase:
Closes #2499 

~I haven't added a test case yet~ because the current zmodel used for testing the zod package specifies `sqlite` which doesn't support array types.

Edit: I've now added a test case but it required switching the provider to `postgresql` and regenerating the schema. Let me know if that's not desired.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling so scalar fields declared as arrays are recognized as arrays before optional/nullable processing.

* **Tests**
  * Added/updated test coverage for array-typed scalar fields (e.g., Post.tags inferred as string[] for create/update).
  * Test schema switched to PostgreSQL and includes new array fields (Post.tags, Address.residents).

### DIFF
--- a/packages/zod/test/factory.test.ts
+++ b/packages/zod/test/factory.test.ts
@@ -376,7 +376,7 @@ describe('SchemaFactory - makeModelSchema', () => {
             const userSchema = factory.makeModelSchema('User');
             const result = userSchema.safeParse({
                 ...validUser,
-                address: { street: '123 Main St', city: 'Springfield', zip: null },
+                address: { residents: [], street: '123 Main St', city: 'Springfield', zip: null },
             });
             expect(result.success).toBe(true);
         });
@@ -385,7 +385,7 @@ describe('SchemaFactory - makeModelSchema', () => {
             const userSchema = factory.makeModelSchema('User');
             const result = userSchema.safeParse({
                 ...validUser,
-                address: { street: '123 Main St', city: 'Springfield', zip: '12345' },
+                address: { residents: [], street: '123 Main St', city: 'Springfield', zip: '12345' },
             });
             expect(result.success).toBe(true);
         });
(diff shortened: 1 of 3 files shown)


packages/zod/src/types.ts
ymc9, line 85: I believe this type also needs an update.

ymc9 (commented):
Thanks for the fix @haltcase ! LGTM. Left one comment there.

packages/zod/src/types.ts
haltcase: Started looking at it. It's a bit trickier so far and isn't immediately the same solution as the others; the type of `Field` in that type isn't compatible with `FieldIsArray` and I'll have to spend more time with it when I have a chance.

EDIT: I found that there's a separate utility for type defs `TypeDefFieldIsArray`.

haltcase:
> Thanks for the fix @haltcase ! LGTM. Left one comment there.

@ymc9 updated!

ymc9:
> > Thanks for the fix @haltcase ! LGTM. Left one comment there.
> 
> @ymc9 updated!

Thanks! Will merge when CI passes. It'll be included in v3.5.

PR #2502, opened 2026-03-20
fix(policy): allow dangerous raw SQL opt-in

### LINKED ISSUE #2495: PolicyPlugin rejects raw SQL methods ($queryRaw, $executeRaw)
## Description

`$queryRawUnsafe`, `$executeRawUnsafe`, `$queryRaw`, and `$executeRaw` fail with `"non-CRUD queries are not allowed"` when PolicyPlugin is active.

## Reproduction

```ts
const client = new ZenStackClient(schema, {
  dialect: new PostgresDialect({ pool }),
  plugins: [new PolicyPlugin()],
});

// This throws: "non-CRUD queries are not allowed"
await client.$queryRawUnsafe('SELECT 1');
```

## Root cause

In `client-impl.ts`, all four raw methods execute through `this.kysely`, which routes queries through `ZenStackQueryExecutor` and the plugin pipeline. PolicyPlugin's `handle()` (`policy-handler.ts:78-84`) rejects any node that isn't Select/Insert/Update/Delete:

```ts
if (!this.isCrudQueryNode(node)) {
    throw createRejectedByPolicyError(
        undefined,
        RejectedByPolicyReason.OTHER,
        'non-CRUD queries are not allowed',
    );
}
```

Raw SQL produces a `RawNode`, not a CRUD node, so it's always rejected.

## Why this should be fixed

Raw SQL has no model-level semantics — there's no table/row context for PolicyPlugin to enforce access control on. The caller is fully responsible for the SQL they write. Blocking raw SQL forces users to either:

- Maintain a separate non-policy client just for raw queries
- Call `$unuseAll()` before every raw query (which also has a transaction isolation bug, see #2494)
- Mutate `$options.plugins` at runtime to temporarily remove PolicyPlugin

All three workarounds are fragile and error-prone.

## Suggested fix

Route raw SQL methods through `kyselyRaw` instead of `this.kysely`:

```ts
$queryRawUnsafe<T = unknown>(query: string, ...values: any[]) {
    return createZenStackPromise(async () => {
        const compiledQuery = this.createRawCompiledQuery(query, values);
        const result = await this.kyselyRaw.executeQuery(compiledQuery);
        return result.rows as T;
    });
}
```

Note: `kyselyRaw` would also need to be updated inside `interactiveTransaction` to preserve transaction isolation (same pattern as `this.kysely = tx`).

Alternatively, PolicyPlugin could pass through non-CRUD nodes instead of rejecting them.

## Version

`@zenstackhq/orm` 3.4.6, `@zenstackhq/plugin-policy` 3.4.6

pkudinov:
Fixes #2495

## Summary
- add `PolicyPluginOptions` with a `dangerouslyAllowRawSql` escape hatch
- bypass `PolicyPlugin.onKyselyQuery` for `RawNode` queries when that option is enabled
- add a real policy-schema regression test showing that raw SQL still rejects by default, but succeeds inside the same transaction when the option is enabled

## Why
`$use()` / `$unuse()` / `$unuseAll()` transaction preservation (#2497) fixes the transaction-affinity side of the problem, but it does not help with raw SQL if the policy plugin still intercepts the query and rejects non-CRUD nodes.

This change keeps the default behavior unchanged and adds an explicit opt-in for applications that need raw SQL to remain available inside authenticated transactions.

## Testing
- `pnpm --filter @zenstackhq/plugin-policy build`
- `TEST_PG_USER=user TEST_PG_PASSWORD=password pnpm --filter @zenstackhq/plugin-policy test -- test/raw-sql.test.ts`

## Summary by CodeRabbit

* **New Features**
  * Policy plugin accepts configuration options, including a new dangerouslyAllowRawSql flag to change raw-SQL handling.

* **Tests**
  * Added end-to-end tests verifying policy behavior when raw SQL is used inside transactions.

### DIFF
--- a/packages/plugins/policy/src/plugin.ts
+++ b/packages/plugins/policy/src/plugin.ts
@@ -1,9 +1,10 @@
 import { type OnKyselyQueryArgs, type RuntimePlugin } from '@zenstackhq/orm';
 import type { SchemaDef } from '@zenstackhq/orm/schema';
+import type { PolicyPluginOptions } from './options';
 import { check } from './functions';
-import { PolicyHandler, type PolicyHandlerOptions } from './policy-handler';
+import { PolicyHandler } from './policy-handler';
 
-export type PolicyPluginOptions = PolicyHandlerOptions;
+export type { PolicyPluginOptions } from './options';
 
 export class PolicyPlugin implements RuntimePlugin<SchemaDef, {}, {}, {}> {
     constructor(private readonly options: PolicyPluginOptions = {}) {}
--- a/packages/plugins/policy/src/options.ts
+++ b/packages/plugins/policy/src/options.ts
@@ -1,0 +1,8 @@
+export type PolicyPluginOptions = {
+    /**
+     * Dangerously bypasses access-policy enforcement for raw SQL queries.
+     * Raw queries remain in the current transaction, but the policy plugin will
+     * not inspect or reject them.
+     */
+    dangerouslyAllowRawSql?: boolean;
+};
--- a/packages/plugins/policy/src/policy-handler.ts
+++ b/packages/plugins/policy/src/policy-handler.ts
@@ -43,6 +43,7 @@ import {
 import { match } from 'ts-pattern';
 import { ColumnCollector } from './column-collector';
 import { ExpressionTransformer } from './expression-transformer';
+import type { PolicyPluginOptions } from './options';
 import type { Policy, PolicyOperation } from './types';
 import {
     buildIsFalse,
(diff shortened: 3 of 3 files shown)


packages/plugins/policy/src/plugin.ts
```diff
@@ -23,6 +35,9 @@ export class PolicyPlugin implements RuntimePlugin<SchemaDef, {}, {}, {}> {
     }
 
     onKyselyQuery({ query, client, proceed }: OnKyselyQueryArgs<SchemaDef>) {
+        if (this.options.dangerouslyAllowRawSql && RawNode.is(query as never)) {
```
ymc9: Hi @pkudinov thanks for working this!

The approach looks good. I think it's cleaner to move this check into the `PolicyHandler` class since the enforcement logic is entirely there. We can pass the options down there. There'll probably be other uses for it in the future too.

pkudinov: Moved in `cc4504b2`.

`PolicyPlugin` now only passes `PolicyHandlerOptions` through, and `PolicyHandler.handle()` is responsible for deciding whether non-CRUD queries are rejected or explicitly allowed when `dangerouslyAllowRawSql` is enabled.

ymc9 (approved):
LGTM!

PR #2503, opened 2026-03-20
feat(zod): add ORM-style select/include/omit options to makeModelSchema

marcsigmund:
Add support for ORM-style field selection and relation inclusion in Zod schema generation. The `makeModelSchema` method now accepts optional `select`, `include`, and `omit` options to control which fields appear in the generated schema.

Key changes:
- Add overloaded signatures to `makeModelSchema` with type-safe options
- Add ModelSchemaOptions type enforcing mutual exclusion of select/include and select/omit at the type level
- Validate options at runtime with a recursive Zod schema – invalid combinations throw a clear error
- Implement `buildFieldsWithOptions` for field filtering and relation inclusion
- Add `GetModelSchemaShapeWithOptions` type to compute the resulting shape based on options
- Apply model-level `@@validate` rules selectively – each rule only runs when all fields it references are present in the resulting shape, keeping the zmodel the single source of truth for validation

Example:
```
// only id and email
factory.makeModelSchema('User', { select: { id: true, email: true } });

// all scalar fields minus username, plus the posts relation
factory.makeModelSchema('User', { omit: { username: true }, include: { posts: true } });

// nested relation options
factory.makeModelSchema('User', {
  select: {
    id: true,
    posts: { select: { title: true } },
  },
});
```

## Summary by CodeRabbit

* **New Features**
  * makeModelSchema accepts ORM-style options to select, include, or omit fields; generated schemas reflect only the requested fields and nested relation options.
  * Custom validations now skip rules that reference fields absent from a partial schema.

* **Types**
  * Added exported types enabling options-aware schema shape inference.

* **Tests**
  * Added runtime and type-level tests covering select/include/omit behaviors and invalid option combinations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,3 +111,23 @@ Thank you for your generous support!
 # Community
 
 Join our [discord server](https://discord.gg/Ykhr738dUe) for chat and updates!
+
+# Contributors
+
+Thanks to all the contributors who have helped make ZenStack better!
+
+### Source
+
+<a href="https://github.com/zenstackhq/zenstack/graphs/contributors">
+  <img src="https://contrib.rocks/image?repo=zenstackhq/zenstack" />
+</a>
+
+### Docs
+
+<a href="https://github.com/zenstackhq/zenstack-docs/graphs/contributors">
+  <img src="https://contrib.rocks/image?repo=zenstackhq/zenstack-docs" />
+</a>
+
+## License
+
+[MIT](LICENSE)
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "zenstack-v3",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "description": "ZenStack",
     "packageManager": "pnpm@10.23.0",
     "type": "module",
--- a/packages/auth-adapters/better-auth/package.json
+++ b/packages/auth-adapters/better-auth/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@zenstackhq/better-auth",
-    "version": "3.4.6",
+    "version": "3.5.0",
     "description": "ZenStack Better Auth Adapter. This adapter is modified from better-auth's Prisma adapter.",
     "type": "module",
     "scripts": {
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -1,0 +1,24 @@
+# @zenstackhq/cli
+
+The command-line interface for ZenStack. Provides commands for initializing projects, generating TypeScript code from ZModel schemas, managing database migrations, and etc.
+
+## Key Commands
+
+- `zenstack init` — Initialize ZenStack in an existing project
+- `zenstack generate` — Compile ZModel schema to TypeScript
+- `zenstack db push` — Sync schema to the database
+- `zenstack db pull` — Pull database schema changes into ZModel
+- `zenstack migrate dev` — Create and apply database migrations
+- `zenstack migrate deploy` — Deploy migrations to production
+- `zenstack format` — Format ZModel schema files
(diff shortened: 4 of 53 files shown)


packages/zod/src/factory.ts
```diff
@@ -39,29 +48,63 @@ class SchemaFactory<Schema extends SchemaDef> {
...
+
+            const shape = z.strictObject(fields);
+            return this.applyDescription(
+                addCustomValidation(shape, modelDef.attributes),
+                modelDef.attributes,
+            ) as unknown as z.ZodObject<GetModelFieldsShape<Schema, Model>, z.core.$strict>;
         }
 
+        // ── Options path ─────────────────────────────────────────────────────
+        const rawOptions = options as unknown as RawOptions;
```
ymc9: Shall we use a zod schema to validate options?

ymc9: If so, we should also follow ORM's rules that select/include cannot coexit, select/omit cannot coexit.

packages/zod/src/types.ts
```diff
@@ -135,3 +135,215 @@ type ZodOptionalAndNullableIf<T extends z.ZodType, Condition extends boolean> =
 type ZodOptionalIf<T extends z.ZodType, Condition extends boolean> = Condition extends true ? z.ZodOptional<T> : T;
 type ZodNullableIf<T extends z.ZodType, Condition extends boolean> = Condition extends true ? z.ZodNullable<T> : T;
 type ZodArrayIf<T extends z.ZodType, Condition extends boolean> = Condition extends true ? z.ZodArray<T> : T;
+
+// -------------------------------------------------------------------------
+// Query options types (Prisma-style include / select / omit)
```
ymc9: I think we can call it "ORM-style query options" now 😄 since zenstack v3 is pretty much decoupled from Prisma.

packages/zod/src/factory.ts
marcsigmund: Good idea
✅ done

packages/zod/src/types.ts
marcsigmund: ✅ done

packages/zod/src/factory.ts
```diff
@@ -30,6 +32,44 @@ export function createSchemaFactory<Schema extends SchemaDef>(schema: Schema) {
     return new SchemaFactory(schema);
 }
 
+/** Internal untyped representation of the options object used at runtime. */
+type RawOptions = {
```
ymc9, line 36: Currently `RawOptions` is not parameterized with schema and model type, so `select`, `include`, and `omit` are of generic record types. It's better to turn it into: `RawOptions<Schema extends SchemaDef, Model extends GetModels<Schema>>`, so that when writing the options, the user gets perfect auto completion for fields.

The same applies to the zod schemas that validate the options. With schema and model context, the zod schemas can be more precise and only allow valid fields.

@marcsigmund , we can either continue evolving this PR to address it, or we can merge it first and I can help make this revision if needed. Please let me know how you want to proceed.

marcsigmund: @ymc9 
Thanks for the feedback. I'm not fully confident I understand the revision well enough to implement it correctly. If you're open to it, I'd love for you to take that part over.

ymc9: Sure. I'll be happy to do that. Thanks!

ymc9: Hi @marcsigmund my apologies, apparently my brain wasn't working correctly when leaving the previous comment. The option object is already perfectly typed (select/include/omit). So it's all good. I've merged and it'll be part of the upcoming v3.5.

Thanks!

PR #2518, opened 2026-03-25
fix(orm): fixed postgres orderBy in nested queries (includes/selects)

### LINKED ISSUE #2464: orderBy on nested includes not working
```
     const route = await db.route.findFirst({
         where: { id: "..." },
         include: {
             stops: {
                 orderBy: { sequence: "asc" },
                 include: { customer: true, location: true }, // nested includes trigger lateral joins
             },
         },
     });

     console.log(route.stops.map(s => s.sequence));
     // Expected: [1, 2, 3]
     // Actual: [3, 2, 1] or any arbitrary order
```

On my end, the generated sql command look something like this:
```
     LEFT JOIN LATERAL (
       SELECT coalesce(
         jsonb_agg(jsonb_build_object('id', sub.id, 'sequence', sub.sequence, ...)),
         --        ^ no ORDER BY inside the aggregate
         '[]'::jsonb
       ) AS "$data"
       FROM (
         SELECT * FROM "RouteStop"
         WHERE "routeId" = $1
         ORDER BY "sequence" ASC   -- ordering is here, in the subquery
         LIMIT 9007199254740991
       ) AS sub
       LEFT JOIN LATERAL (...customer...) ON true   -- these joins can reorder rows
       LEFT JOIN LATERAL (...location...) ON true   -- before they reach jsonb_agg()
     ) AS "Route$stops" ON true
```

What I would expect:
```
     LEFT JOIN LATERAL (
       SELECT coalesce(
         jsonb_agg(
           jsonb_build_object('id', sub.id, 'sequence', sub.sequence, ...)
           ORDER BY sub."sequence" ASC   -- ORDER BY belongs inside the aggregate
         ),
         '[]'::jsonb
       ) AS "$data"
       FROM (
         SELECT * FROM "RouteStop"
         WHERE "routeId" = $1
         ORDER BY "sequence" ASC
         LIMIT 9007199254740991
       ) AS sub
       LEFT JOIN LATERAL (...customer...) ON true
       LEFT JOIN LATERAL (...location...) ON true
     ) AS "Route$stops" ON true
```

-   ZenStack version: 3.4.4
-   Database type: Postgresql
-   Node.js/Bun version: Bun 1.3.6
-   Package manager: Bun

**Additional context**

eredzik:
# Summary

Fixes #2464 

- Fix ordering of to-many relation arrays when the relation query includes nested include/select clauses. The buildRelationOrderByExpressions method in the lateral join dialect was not propagating ORDER BY into the jsonb_agg / JSON_ARRAYAGG call, causing databases (especially PostgreSQL) to return relation arrays in arbitrary order when nested joins were present.

# Test plan

-  New E2E test order-by-nested-includes.test.ts covers both SQLite and PostgreSQL:
keeps stable order for to-many include with nested includes — creates 10 (SQLite) / 2000 (PostgreSQL) posts with descending insert order, queries with orderBy: { sequence: 'asc' } and 'desc' while including nested author and comments.author relations. Asserts the returned sequence arrays match the expected sorted order.
keeps stable order for to-many select with nested selects — same scenario using select instead of include, verifying ordering is preserved through nested select clauses as well.
- Existing ORM client-api and relation tests continue to pass (no behavioral change to the query logic, only type-level cleanup).

## Summary by CodeRabbit

* **New Features**
  * To-many relation results now preserve requested ordering when included or selected, keeping sort direction across nested fetches.

* **Compatibility**
  * PostgreSQL now applies ordering inside aggregated relation arrays; MySQL retains provider-specific deterministic ordering behavior.

* **Tests**
  * Added end-to-end tests that verify stable ordering for nested includes and nested selects with ascending and descending sequences.

### DIFF
--- a/packages/orm/src/client/crud/dialects/lateral-join-dialect-base.ts
+++ b/packages/orm/src/client/crud/dialects/lateral-join-dialect-base.ts
@@ -213,10 +213,16 @@ export abstract class LateralJoinDialectBase<Schema extends SchemaDef> extends B
                 }
 
                 const expr = this.fieldRef(model, field, modelAlias);
+                let sort = typeof value === 'string' ? value : value.sort;
+                if (payload.take !== undefined && payload.take < 0) {
+                    // negative `take` requires negated sorting, and the result order
+                    // will be corrected during post-read processing
+                    sort = this.negateSort(sort, true);
+                }
                 if (typeof value === 'string') {
-                    items.push({ expr, sort: value });
+                    items.push({ expr, sort });
                 } else {
-                    items.push({ expr, sort: value.sort, nulls: value.nulls });
+                    items.push({ expr, sort, nulls: value.nulls });
                 }
             }
         }


packages/orm/src/client/crud/dialects/mysql.ts
```diff
@@ -192,8 +192,33 @@ export class MySqlCrudDialect<Schema extends SchemaDef> extends LateralJoinDiale
...
+                    return [sql`${expr} ${dir}`];
+                }
+
+                const nullsSql = nulls === 'first' ? sql`(${expr} IS NULL) DESC` : sql`(${expr} IS NULL) ASC`;
+                return [nullsSql, sql`${expr} ${dir}`];
+            }),
+            sql.raw(', '),
+        );
+
+        return this.eb.fn.coalesce(
```
ymc9: I believe mysql doesn't support order by inside `JSON_ARRAYAGG`. Maybe we need a subquery for sorting?

eredzik: You are right, current query was failing. I've added mysql to run the test and the original implementation works as expected in mysql (it sorts in subquery)

ymc9 (dismissed):
Thanks @eredzik .

I believe mysql needs some more fixes since the `JSON_ARRAYAGG` doesn't preserve input order either. I'll make a separate change for that. 

I've also simplified the tests and removed the multi-provider matrix since the CI automatically do multiple runs for each provider. Will merge once CI passes and publish a patch release.

eredzik:
Thanks for cleaning tests up, I was not aware that this is set up in this way for CI.

As for MySQL, when I ran MySQL tests on my PC with db in docker, tests returned correct ordering in both asc and desc cases with current version. I haven't ever used MySQL though so my knowledge there is limited unfortunately

ymc9:
> Thanks for cleaning tests up, I was not aware that this is set up in this way for CI.
> 
> As for MySQL, when I ran MySQL tests on my PC with db in docker, tests returned correct ordering in both asc and desc cases with current version. I haven't ever used MySQL though so my knowledge there is limited unfortunately

Same here, tests pass consistently, however mysql documentation seems to imply the order is not guaranteed, so may be one day someone will see a mismatch 🤷🏻‍♂️. I think it's safer to sort again post-read.

ymc9 (approved):

PR #2521, opened 2026-03-25
feat(openapi): add nestedRoutes support for openapi spec generation

lsmith77:
## Summary by CodeRabbit

* **New Features**
  * Configurable nested relationship routes in generated OpenAPI specs, adding conditional nested create/update/delete endpoints and single-resource child paths for collection relations.

* **Bug Fixes / Refactor**
  * Safer related-model handling and cleaner path-item generation to avoid unsafe assumptions.

* **Tests**
  * Added tests validating nestedRoutes behavior, OpenAPI schema validity, unique operationIds, and slicing-based exclusion of nested operations.

### DIFF
--- a/packages/server/src/api/rest/openapi.ts
+++ b/packages/server/src/api/rest/openapi.ts
@@ -61,6 +61,10 @@ export class RestApiSpecGenerator<Schema extends SchemaDef = SchemaDef> {
         return this.handlerOptions?.queryOptions;
     }
 
+    private get nestedRoutes(): boolean {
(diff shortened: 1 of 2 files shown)


packages/server/src/api/rest/openapi.ts
ymc9: Shall we give this method a new name? Since it's not only fetch anymore.

lsmith77: how about simply `buildRelatedPath` ?

PR #2525, opened 2026-03-26
feat(zod): enforce literal true for select/include/omit options + exclude relation fields from makeModelSchema by default + add "optionality" setting to control runtime optionality of fields

marcsigmund:
## Summary by CodeRabbit

* **Breaking Changes**
  * Schemas no longer include related model fields by default; relations must be explicitly opted into via select/include/omit (nested entries now require literal true).
  * Default schemas now reject relation fields at runtime unless explicitly included.

* **New Features**
  * Added an "optionality" setting to control field optionality at runtime ("all" vs "defaults").

* **Bug Fixes / Behavior**
  * Validation treats missing fields as absent to avoid spurious validations for partial inputs.

* **Documentation**
  * Added breaking-changes note describing the new default behavior.

### DIFF
--- a/packages/zod/src/factory.ts
+++ b/packages/zod/src/factory.ts
@@ -327,7 +327,7 @@ class SchemaFactory<Schema extends SchemaDef> {
 
         if (select) {
             // Only scalar fields explicitly selected (value is always `true` or nested options).
-            for (const [key] of Object.entries(select)) {
+            for (const key of Object.keys(select)) {
                 const fieldDef = modelDef.fields[key];
                 if (fieldDef && !fieldDef.relation) {
                     fields.add(key);
--- a/packages/zod/src/factory.ts
+++ b/packages/zod/src/factory.ts
@@ -34,9 +34,9 @@ export function createSchemaFactory<Schema extends SchemaDef>(schema: Schema) {
 
 /** Internal untyped representation of the options object used at runtime. */
 type RawOptions = {
-    select?: Record<string, unknown>;
-    include?: Record<string, unknown>;
-    omit?: Record<string, unknown>;
+    select?: Record<string, true | RawOptions>;
+    include?: Record<string, true | RawOptions>;
+    omit?: Record<string, true>;
 };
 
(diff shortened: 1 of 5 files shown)


packages/zod/src/factory.ts
```diff
@@ -296,9 +326,8 @@ class SchemaFactory<Schema extends SchemaDef> {
         const fields = new Set<string>();
 
         if (select) {
-            // Only scalar fields explicitly selected with a truthy value.
-            for (const [key, value] of Object.entries(select)) {
-                if (!value) continue;
+            // Only scalar fields explicitly selected (value is always `true` or nested options).
+            for (const [key] of Object.entries(select)) {
```
ymc9: Change to `Object.keys` here?

packages/zod/src/types.ts
```diff
@@ -326,34 +360,49 @@ type BuildIncludeOmitShape<
...
 /**
  * The top-level conditional that maps options → Zod shape.
  *
  * - No options / undefined  → existing `GetModelFieldsShape` (no change).
- * - `{ select: S }`         → `BuildSelectShape`.
- * - `{ include?, omit? }`   → `BuildIncludeOmitShape`.
+ * - `{ select: S }`         → `BuildSelectShape` (+ optionality wrapper).
+ * - `{ include?, omit? }`   → `BuildIncludeOmitShape` (+ optionality wrapper).
+ *
+ * Note: `optionality: "defaults"` is handled fully at runtime (it requires
```
ymc9: The `FieldDef` type carries a `default` and a `updatedAt` field that can be used to infer if it's auto-assigned. The "@zenstackhq/schema" package also exports a `FieldHasDefault` helper to do the inference.

ymc9 (commented):
Hi @marcsigmund , my apologies for the late review. Please see my comments attached. Thanks!

ymc9 (approved):
LGTM! I'm merging and will release it with v3.6.0. Thanks @marcsigmund !

PR #2528, opened 2026-03-30
fix(orm): unify zenstack v3 and prisma behavior on empty branches in OR filter

eredzik:
# Summary

I found this issue in our migration to v3 where behavior differed from previous prisma based v2 version.
For example queries such as:
```
await client.testItem.findMany({
    where: { OR: [{ name: undefined }] },
  });
```
Is treated by zenstack as: 
```
select "TestItem"."id" as "id", "TestItem"."name" as "name", "TestItem"."createdAt" as "createdAt" from "TestItem" where 1 
```
but prisma interpreted it as:
```
SELECT `main`.`TestItem`.`id`, `main`.`TestItem`.`name`, `main`.`TestItem`.`createdAt` FROM `main`.`TestItem` WHERE 1=0 LIMIT ? OFFSET ?
```

# Work done
- Added test to verify such case
- Added fix in code to unify the behavior to how prisma treated such cases
- Ran all tests locally and they passed

# Possible further discussions

The way zenstack interprets these queries seems more intuitive than prisma. Question is whether project aims for 1:1 behavior parity with how prisma queries data (I assume so)?

## Summary by CodeRabbit

* **Bug Fixes**
  * Query filters now ignore wholly-undefined branches and skip undefined operator values; AND/OR/NOT with no effective branches behave consistently (AND/NOT act as no-op/true, OR yields no matches).
  * Relation operations: empty relation payloads are treated as “match all” for disconnect/delete; single-delete vs multi-delete behavior and validation tightened so delete vs delete-many act predictably.

* **Tests**
  * Added/updated E2E tests covering these filter and relation behaviors.

### DIFF
--- a/packages/orm/src/client/crud/dialects/base-dialect.ts
+++ b/packages/orm/src/client/crud/dialects/base-dialect.ts
@@ -265,14 +265,10 @@ export abstract class BaseCrudDialect<Schema extends SchemaDef> {
                 this.and(...enumerate(payload).map((subPayload) => this.buildFilter(model, modelAlias, subPayload))),
             )
             .with('OR', () => {
-                const allBranches = enumerate(payload).map((subPayload) =>
-                    this.buildFilter(model, modelAlias, subPayload)
-                );
-                const meaningfulBranches = allBranches.filter((expr) => !this.isTrue(expr));
-                if (meaningfulBranches.length === 0) {
-                    return allBranches.length > 0 ? this.true() : this.false();
-                }
-                return this.or(...meaningfulBranches);
+                const branches = enumerate(payload)
+                    .map((subPayload) => this.buildFilter(model, modelAlias, subPayload))
+                    .filter((expr) => !this.isTrue(expr));
+                return this.or(...branches);
             })
             .with('NOT', () => this.eb.not(this.buildCompositeFilter(model, modelAlias, 'AND', payload)))
             .exhaustive();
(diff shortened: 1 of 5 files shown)


packages/orm/src/client/crud/operations/base.ts
```diff
@@ -2531,11 +2531,17 @@ export abstract class BaseOperationHandler<Schema extends SchemaDef> {
...
+                for (const element of args) {
+                    this.doNormalizeArgs(element);
+                }
+            } else {
+                for (const [key, value] of Object.entries(args)) {
+                    if (value === undefined) {
+                        delete args[key as keyof typeof args];
+                    } else if (value && isPlainObject(value)) {
+                        this.doNormalizeArgs(value);
+                    }
```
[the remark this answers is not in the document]
eredzik: Fixed issue

packages/orm/src/client/crud/dialects/base-dialect.ts
ymc9, line 276: I think we should probably do a refactor here and uniformly preprocess `payload` to strip:

1. object entries with `undefined` values like: `AND: { id: undefined, x: 1 }`, or `AND: [ {id: undefined}, { x, 1}]`
2. array elements that are objects that are empty (may be result of step above)

Then have the processed payload enter the following match statement? What do you think?

ymc9 (commented):
Hi @eredzik , thanks for working on the fix! Please see my comments attached.

packages/orm/src/client/crud/dialects/base-dialect.ts
ymc9: Hi @eredzik , I've made some refactor to uniformly handle AND/OR/NOT for conditions with undefined values. All seem to work fine. Merging and publishing a patch release. Thanks!

ymc9 (approved):

packages/orm/src/client/crud/dialects/base-dialect.ts
eredzik: Hi @ymc9, thanks for doing that. I unfortunately have not found spare time to tackle that refactor. I greatly appreciate you doing it and merging!

PR #2537, opened 2026-03-31
fix(policy): currentModel and currentOperation inside of collection predicates

### LINKED ISSUE #2536: currentModel() / currentOperation() does not work inside of collection predicates

This does not work when added to a model as policy:

```
@@allow('all', auth().groups?[name == currentModel()])
```

It fails with: 

```
Error: Failed to execute query: Error: Unsupported call expression function: currentModel
sql: 'select "User"."id" as "id", "User"."createdAt" as "createdAt", "User"."updatedAt" as "updatedAt", "User"."username" as "username", "User"."password" as "password", "User"."isAdmin" as "isAdmin" from "User" where ("User"."username" = $1) limit $2 offset $3 ',
  sqlParams: [ 'admin', 25, 0 ],
```

**Screenshots**

-   ZenStack version: 3.5.2
-   Database type: Postgresql
-   Node.js/Bun version: 24.13.1
-   Package manager: npm

**Additional context**

As discussed on discord, this seems like a bug.

wolflu05:
This fixes `currentModel()` and `currentOperation()` inside of collection predicates

fixes #2536 

## Summary by CodeRabbit

* **New Features**
  * Policy expressions can now access the current model type and CRUD operation via built-ins, and nested expressions evaluate with that context.

* **Tests**
  * Added a regression test validating context-aware policy evaluation using the new model/operation built-ins in nested rules.

### DIFF
--- a/tests/regression/test/issue-2536.test.ts
+++ b/tests/regression/test/issue-2536.test.ts
@@ -17,9 +17,13 @@ describe('Regression for issue #2536', () => {
                 users User[] @relation("UserGroups")
             }
 
-            model Foo {
-                id String @id @default(cuid())
+            // define a mixin to also check that currentModel correctly resolves to the model where the mixin is applied
+            type AuthPolicyMixin {
                 @@allow('all', auth().groups?[modelName == currentModel() && modelOperation == currentOperation()])
+            }
+
+            model Foo with AuthPolicyMixin {
+                id String @id @default(cuid())
             }
             `,
         );


ymc9 (approved):
Hi @wolflu05 , thanks for making the fix! It'll solve the immediate issue.

The expression evaluator's function call handling needs some overall refactoring. I'll make a follow-up PR a bit later.

wolflu05:
Thanks for you so quick response. I really start to enjoy using zenstack.

PR #2539, opened 2026-03-31
`@computed` fields from mixin types (`with`) cause `column does not exist` error when the model is explicitly included

lsmith77:
failing test to illustrate https://github.com/zenstackhq/zenstack/issues/2540 along with a one line fix

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end Vitest verifying computed fields inherited from a mixin are populated on child records both when queried directly and when returned via a parent query with nested includes.

* **Bug Fix**
  * Broadened computed-field detection so inherited/aggregated computed fields are surfaced correctly, preventing missing computed values for nested includes.


ymc9 (approved):
LGTM! I've changed the test to a regression since it targets a specific issue. Thanks @lsmith77 !

PR #2587, opened 2026-04-17
fix(cli): fix race condition during `push db` (#2491)

### LINKED ISSUE #2491: Race condition because `zen push db` creates a temporal file with a fixed name
`zen push db` creates a temporal file named `~schema.prisma`.  Running that command in parallel cause a race condition.

**Simulation**

```
$ echo `line 1` > zenstack/~schema.prisma # Process 1
$ zen db push # Process 2
$ echo `line 2` >> zenstack/~schema.prisma # Process 1
$ cat zenstack/~schema.prisma
```

The output is:
```
line 2
```

The expected output was:
```
line 1
line 2
```

-   ZenStack version: 3.4.6
-   Database type: PostgreSQL
-   Node.js/Bun version: v22.17.1
-   Package manager: yarn

**Additional context**
We run parallel tests with parallel containers for the databases.  But the test runner itself is running in the same single shell, so it causes the race condition.

jesus-gomez-aetonix:
Because a hardcoded prefix creating the temporal schema file `~schema.prisma`, multiple processes running `push db` will suffer a race condition.

Change the hardcoded prefix with a sort of UUID.

Fix: [2491](https://github.com/zenstackhq/zenstack/issues/2491).

## Summary by CodeRabbit

* **New Features**
  * Added a CLI flag to optionally generate randomized temporary Prisma schema filenames for migrate and db push flows.
* **Bug Fixes**
  * Reduces temporary schema filename collisions/overwrites during concurrent CLI operations.
* **Tests**
  * Added tests validating default and randomized temporary schema generation and cleanup.

### DIFF
--- a/packages/cli/src/actions/action-utils.ts
+++ b/packages/cli/src/actions/action-utils.ts
@@ -4,6 +4,7 @@ import { type Model, type Plugin, isDataSource, type LiteralExpr } from '@zensta
 import { type CliPlugin, PrismaSchemaGenerator } from '@zenstackhq/sdk';
 import colors from 'colors';
 import { createJiti } from 'jiti';
+import crypto from 'node:crypto';
 import fs from 'node:fs';
 import { createRequire } from 'node:module';
 import path from 'node:path';
@@ -95,7 +96,7 @@ export async function generateTempPrismaSchema(zmodelPath: string, folder?: stri
     if (!folder) {
         folder = path.dirname(zmodelPath);
     }
-    const prismaSchemaFile = path.resolve(folder, '~schema.prisma');
+    const prismaSchemaFile = path.resolve(folder, `~schema.${crypto.randomUUID()}.prisma`);
     fs.writeFileSync(prismaSchemaFile, prismaSchema);
     return prismaSchemaFile;
 }
--- a/packages/cli/src/actions/action-utils.ts
+++ b/packages/cli/src/actions/action-utils.ts
@@ -87,16 +87,19 @@ export function handleSubProcessError(err: unknown) {
     }
 }
 
-export async function generateTempPrismaSchema(zmodelPath: string, folder?: string) {
+export async function generateTempPrismaSchema(
+    zmodelPath: string,
+    opts: { folder?: string; randomName?: boolean } = {},
+) {
+    const { folder: folderOpt, randomName = false } = opts;
     const model = await loadSchemaDocument(zmodelPath);
     if (!model.declarations.some(isDataSource)) {
         throw new CliError('Schema must define a datasource');
     }
     const prismaSchema = await new PrismaSchemaGenerator(model).generate();
-    if (!folder) {
-        folder = path.dirname(zmodelPath);
-    }
-    const prismaSchemaFile = path.resolve(folder, `~schema.${crypto.randomUUID()}.prisma`);
+    const folder = folderOpt ?? path.dirname(zmodelPath);
+    const fileName = randomName ? `~schema.${crypto.randomUUID()}.prisma` : '~schema.prisma';
+    const prismaSchemaFile = path.resolve(folder, fileName);
     fs.writeFileSync(prismaSchemaFile, prismaSchema);
     return prismaSchemaFile;
 }
--- a/packages/cli/src/actions/db.ts
+++ b/packages/cli/src/actions/db.ts
@@ -22,6 +22,7 @@ type PushOptions = {
     schema?: string;
     acceptDataLoss?: boolean;
     forceReset?: boolean;
+    randomPrismaSchemaName?: boolean;
 };
 
 export type PullOptions = {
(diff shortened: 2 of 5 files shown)


packages/cli/src/actions/action-utils.ts
```diff
@@ -95,7 +96,7 @@ export async function generateTempPrismaSchema(zmodelPath: string, folder?: stri
     if (!folder) {
         folder = path.dirname(zmodelPath);
     }
-    const prismaSchemaFile = path.resolve(folder, '~schema.prisma');
+    const prismaSchemaFile = path.resolve(folder, `~schema.${crypto.randomUUID()}.prisma`);
```
ymc9: Hi @jesus-gomez-aetonix , a nitpick here, since we don't need very strong conflict avoidance here, maybe we can use a shorter id like `crypto.randomBytes(8).toString('hex')`?

There might be cases the temp file is not properly cleaned up (e.g., user ctrl+c forced an abort). Just hope the file name is not too ugly 😆

jesus-gomez-aetonix: Yes, we could easily do that.

Some considerations:

 - Maybe "ugly" is a feature and not an undesirable byproduct: It could highlight better than a beauty name a non-clean file (which should be cleaned up). [1]

 - `randomUUID()` is more intentional than `randomBytes(8).toString('hex')` in the sense that it conveys the idea of "uniqueness", which is tied to solving the problem of race condition.

 - Finally, I made a quick profiling, and the UUID version is 8x faster than the `randomBytes` and than alternatives Claude helped me to craft[2] (I was not expecting that) :

```
┌─────────┬────────────────────────────────────┬──────────────────┬──────────────────┬────────────────────────┬────────────────────────┬──────────┐
│ (index) │ Task name                          │ Latency avg (ns) │ Latency med (ns) │ Throughput avg (ops/s) │ Throughput med (ops/s) │ Samples  │
├─────────┼────────────────────────────────────┼──────────────────┼──────────────────┼────────────────────────┼────────────────────────┼──────────┤
│ 0       │ 'randomUUID()'                     │ '69.64 ± 0.56%'  │ '83.00 ± 41.00'  │ '17717945 ± 0.01%'     │ '12048193 ± 4048193'   │ 28718632 │
│ 1       │ 'randomBytes(8).toString("hex")'   │ '584.58 ± 0.08%' │ '583.00 ± 41.00' │ '1734695 ± 0.01%'      │ '1715266 ± 129753'     │ 3421242  │
│ 2       │ 'randomFillSync + toString("hex")' │ '508.52 ± 1.71%' │ '459.00 ± 41.00' │ '2061336 ± 0.01%'      │ '2178649 ± 178649'     │ 3932977  │
│ 3       │ 'randomBytes(8) only'              │ '549.12 ± 2.50%' │ '541.00 ± 41.00' │ '1866901 ± 0.01%'      │ '1848429 ± 151571'     │ 3642212  │
│ 4       │ 'randomFillSync only'              │ '499.22 ± 2.59%' │ '458.00 ± 41.00' │ '2190098 ± 0.01%'      │ '2183406 ± 183406'     │ 4006226  │
└─────────┴────────────────────────────────────┴──────────────────┴──────────────────┴────────────────────────┴────────────────────────┴──────────┘
```

What do you think?


---
Footnotes:

[1] Tangent: This gave me the idea for a small tweak in the future of appending some comment in the temporal schema that says: "If you can READ this message, that means I'm already dead." (kind of bad joke).

[2] The benchmark script:

```javascript
import { Bench } from 'tinybench';
import crypto from 'crypto';

const bench = new Bench({ time: 2000 });
const buf = Buffer.allocUnsafe(8);

bench
  // Full pipelines
  .add('randomUUID()', () => crypto.randomUUID())
  .add('randomBytes(8).toString("hex")', () => crypto.randomBytes(8).toString('hex'))
  .add('randomFillSync + toString("hex")', () => { crypto.randomFillSync(buf); return buf.toString('hex'); })

  // Isolated random generation (no toString)
  .add('randomBytes(8) only', () => crypto.randomBytes(8))
  .add('randomFillSync only', () => crypto.randomFillSync(buf));

await bench.run();
console.table(bench.table());
```

ymc9: Thanks for the update @jesus-gomez-aetonix . Just curious about your use case: do you run "db push" in parallel in large scale?

jesus-gomez-aetonix: I work in a place where they run test suites tests in parallel and each one launch its own docker containers (one of them for a fresh postgreSQL) and yes, each time they call `db push`.

I think they are caped to some number of parallel processes as a function of the available core/threads.

It is not much (something between 8 and 32), but enough to break the CI.

So during the migration from Prisma to Zenstack, that was one of the problems I faced, resolving it with the Prisma plugin, and changing `zen db push` with `prisma db push --schema=${tmpSchemaPath} --url=${connectionUri} --accept-data-loss` (thanks to your suggestion in Discord).

I wrote the ticket time ago before applying the workaround in our test runners.

ymc9: Got it. Since it's for testing, shall we introduce a CLI option to guard it so that it doesn't affect normal usage? Something like `--random-prisma-schema-name`? Using a regular UUID is completely fine then.

jesus-gomez-aetonix: I updated the PR with an implementation of that proposed option.  I wanted to make it the least disturbing change, but at the same time, I wanted it to be a flag that controls only the function `generateTempPrismaSchema`, independently of who is calling it.  That's why you see I updated not only `db push` but also `migrate`.

Let me know what you think.

ymc9 (approved):
Looks great to me! Thanks for the quick follow-up.

PR #2590, opened 2026-04-18
fix(orm): return @db.Time fields as Date instead of raw string (#2589)

### LINKED ISSUE #2589: @db.Time fields returned as raw strings (not Date) when fetched via nested include
## Summary

`ZenStackClient` returns raw strings for `@db.Time` fields when those fields live on a relation pulled in via `include`. Prisma v6 against the same database and schema returns `Date` objects.

The generated TypeScript types declare these fields as `Date`, so calling any `Date` method on the value (e.g. `getUTCFullYear()`) throws `… is not a function` at runtime. There's no compile-time hint that anything is wrong.

## Reproduction

Standalone repo: https://github.com/erwan-joly/zenstack-time-bug-repro

```bash
git clone https://github.com/erwan-joly/zenstack-time-bug-repro
cd zenstack-time-bug-repro
pnpm install
pnpm db:up
pnpm exec prisma db push --schema schema/schema.prisma --skip-generate
pnpm exec prisma generate --schema schema/schema.prisma
pnpm exec zen generate
pnpm exec tsx seed.ts
pnpm repro
```

Schema: a single `Exchange ↔ ExchangeTradingWindow` relation. The trading-window model has `open` / `close` typed as `DateTime @db.Time(6)`, plus a `@db.Date` and a `@db.Timestamp` for comparison.

The query (identical for both clients):

```ts
client.exchange.findMany({
  where: { id: '11111111-...' },
  include: { tradingWindows: true },
});
```

## Expected vs actual

```
=== Prisma v6 (PrismaClient) ===
tradingWindows[0].open          : 1970-01-01T09:30:00.000Z   typeof: object   isDate: true
tradingWindows[0].close         : 1970-01-01T16:00:00.000Z   typeof: object   isDate: true
tradingWindows[0].effectiveOn   : 2024-01-01T00:00:00.000Z   typeof: object   isDate: true
tradingWindows[0].createdAt     : 2026-04-18T07:42:59.769Z   typeof: object   isDate: true

=== ZenStack v3 (ZenStackClient) ===
tradingWindows[0].open          : 09:30:00                   typeof: string   isDate: false   <-- BUG
tradingWindows[0].close         : 16:00:00                   typeof: string   isDate: false   <-- BUG
tradingWindows[0].effectiveOn   : 2024-01-01T00:00:00.000Z   typeof: object   isDate: true
tradingWindows[0].createdAt     : 2026-04-18T07:42:59.769Z   typeof: object   isDate: true
```

`@db.Timestamp` and `@db.Date` round-trip correctly. Only `@db.Time` is broken.

## Root cause

The Postgres dialect builds nested includes as `LEFT JOIN LATERAL (… jsonb_build_object(…) …)`. The relation arrives as a JSONB column, so pg's per-OID type parsers don't fire on the leaves — `transformOutputDate` has to do the string→`Date` conversion itself.

Implementation in `@zenstackhq/orm@3.5.6` at `dist/index.cjs:2306`:

```js
transformOutputDate(value) {
  if (typeof value === "string") {
    const normalized = this.hasTimezoneOffset(value) ? value : `${value}Z`;
    const parsed = new Date(normalized);
    return Number.isNaN(parsed.getTime()) ? value : parsed;
    //                                       ^^^^^ silently returns the raw string
  }
  return value;
}
```

| Input format     | After `+ 'Z'`             | `new Date(...)`     |
| ---------------- | ------------------------- | ------------------- |
| `"2024-01-01 00:00:00"` (Timestamp) | `"… Z"`         | ✅ valid            |
| `"2024-01-01"` (Date)               | `"2024-01-01Z"` | ✅ valid (UTC midnight) |
| `"09:30:00"` (Time)                 | `"09:30:00Z"`   | ❌ **Invalid Date** |

When parsing fails, the function returns the original string instead of throwing or returning `null` — so the broken value silently propagates.

The dialect already maps `"@db.Time": "time"` in its TYPE_MAP at `dist/index.cjs:2155`, so it has the metadata it needs to handle this case — `transformOutputDate` just doesn't branch on the field's `@db.*` attribute.

## Suggested fix

Either:

1. Add a `transformOutputTime` branch and dispatch from `transformOutput` based on the field's `@db.Time`/`@db.Timetz` attribute (use `` new Date(`1970-01-01T${value.endsWith('Z') ? value : value + 'Z'}`) ``), or
2. Throw / return `null` from `transformOutputDate` instead of silently returning the raw string when `Date.parse` fails — at least surface the bug instead of masking it.

## Versions

- `@zenstackhq/orm` 3.5.6 (also reproduces in 3.4.6)
- `@zenstackhq/cli` 3.5.6
- `@prisma/client` 6.19.2 (used as the control)
- `pg` 8.20.0
- `kysely` 0.28.16
- Node 22.21.1
- PostgreSQL 17

erwan-joly:
## Summary

Fixes #2589.

Postgres builds nested includes as `LEFT JOIN LATERAL (... jsonb_build_object(...))`, so the per-OID type parsers pg-node registers don't fire on the leaves — `transformOutputDate` has to do the string→`Date` conversion itself. For `@db.Time` / `@db.Timetz` values (e.g. `09:30:00`, `09:30:00+00`), `new Date(value + 'Z')` returns Invalid Date and the silent fallback returned the raw string, contradicting the generated `Date` field type.

The fix threads `FieldDef` down to `transformOutputDate` so it can detect `@db.Time` / `@db.Timetz` via the field's `@db.*` attribute and anchor the value at the Unix epoch (`1970-01-01`) before parsing, matching Prisma's behavior. It also handles Postgres `timetz`'s `+HH` offset format (no minutes), which `new Date` otherwise rejects.

## Test plan

- [x] Added two cases to `tests/e2e/orm/client-api/timezone/pg-timezone.test.ts` covering `@db.Time` and `@db.Timetz`, both via nested `include` and on direct `findMany`.
- [x] Full `pg-timezone.test.ts` suite passes (28/28).

## Release Notes

* **Bug Fixes**
  * Improved PostgreSQL time-only and timetz field conversion to Date objects
  * Improved MySQL TIME field conversion to Date objects with consistent epoch handling

* **Tests**
  * Added timezone and date handling tests for PostgreSQL databases
  * Added timezone and date handling tests for MySQL databases

### DIFF
--- a/packages/orm/src/client/crud/dialects/mysql.ts
+++ b/packages/orm/src/client/crud/dialects/mysql.ts
@@ -118,7 +118,7 @@ export class MySqlCrudDialect<Schema extends SchemaDef> extends LateralJoinDiale
         }
     }
 
-    override transformOutput(value: unknown, type: BuiltinType, array: boolean, fieldDef?: FieldDef) {
+    override transformOutput(value: unknown, type: BuiltinType, array: boolean) {
         if (value === null || value === undefined) {
             return value;
         }
@@ -127,7 +127,7 @@ export class MySqlCrudDialect<Schema extends SchemaDef> extends LateralJoinDiale
             case 'Boolean':
                 return this.transformOutputBoolean(value);
             case 'DateTime':
-                return this.transformOutputDate(value, fieldDef);
+                return this.transformOutputDate(value);
             case 'Bytes':
                 return this.transformOutputBytes(value);
             case 'BigInt':
@@ -165,18 +165,17 @@ export class MySqlCrudDialect<Schema extends SchemaDef> extends LateralJoinDiale
         return new Decimal(value);
     }
 
-    private transformOutputDate(value: unknown, fieldDef?: FieldDef) {
-        if (value instanceof Date) {
-            return value;
-        }
+    private transformOutputDate(value: unknown) {
         if (typeof value !== 'string') {
             return value;
         }
 
-        // MySQL `TIME` columns are returned as bare time strings (e.g. "09:30:00"),
-        // which `new Date` can't parse — anchor at the Unix epoch to match Prisma.
-        const isTimeField = fieldDef?.attributes?.some((a) => a.name === '@db.Time');
-        const anchored = isTimeField ? `1970-01-01T${value}` : value;
+        // MySQL `TIME` columns return bare time strings ("09:30:00") that `new Date`
+        // can't parse on their own — anchor at the Unix epoch. Detect by shape rather
+        // than the schema attribute so the runtime stays decoupled from `@db.*`
+        // (which is migration/db-push only): TIME starts with `HH:`, DATE/DATETIME
+        // values always start with `YYYY-`.
+        const anchored = /^\d{2}:/.test(value) ? `1970-01-01T${value}` : value;
 
         // MySQL DateTime columns are returned as strings (non-ISO but parsable as JS Date),
         // convert to ISO Date by appending 'Z' if not present
--- a/packages/orm/src/client/crud/dialects/mysql.ts
+++ b/packages/orm/src/client/crud/dialects/mysql.ts
@@ -118,7 +118,7 @@ export class MySqlCrudDialect<Schema extends SchemaDef> extends LateralJoinDiale
         }
     }
 
-    override transformOutput(value: unknown, type: BuiltinType, array: boolean) {
+    override transformOutput(value: unknown, type: BuiltinType, array: boolean, fieldDef?: FieldDef) {
         if (value === null || value === undefined) {
             return value;
         }
@@ -127,7 +127,7 @@ export class MySqlCrudDialect<Schema extends SchemaDef> extends LateralJoinDiale
             case 'Boolean':
                 return this.transformOutputBoolean(value);
             case 'DateTime':
-                return this.transformOutputDate(value);
+                return this.transformOutputDate(value, fieldDef);
             case 'Bytes':
                 return this.transformOutputBytes(value);
             case 'BigInt':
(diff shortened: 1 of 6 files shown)


packages/orm/src/client/crud/dialects/mysql.ts
```diff
@@ -165,16 +165,22 @@ export class MySqlCrudDialect<Schema extends SchemaDef> extends LateralJoinDiale
...
     }
 
-    private transformOutputDate(value: unknown) {
-        if (typeof value === 'string') {
-            // MySQL DateTime columns are returned as strings (non-ISO but parsable as JS Date),
-            // convert to ISO Date by appending 'Z' if not present
-            return new Date(!value.endsWith('Z') ? value + 'Z' : value);
-        } else if (value instanceof Date) {
+    private transformOutputDate(value: unknown, fieldDef?: FieldDef) {
+        if (value instanceof Date) {
```
ymc9: These two branches can be merged I guess? My initial implementation was dumb 😄.

packages/orm/src/client/crud/dialects/mysql.ts
```diff
@@ -165,16 +165,22 @@ export class MySqlCrudDialect<Schema extends SchemaDef> extends LateralJoinDiale
...
             return value;
-        } else {
+        }
+        if (typeof value !== 'string') {
             return value;
         }
+
+        // MySQL `TIME` columns are returned as bare time strings (e.g. "09:30:00"),
+        // which `new Date` can't parse — anchor at the Unix epoch to match Prisma.
+        const isTimeField = fieldDef?.attributes?.some((a) => a.name === '@db.Time');
```
ymc9: Currently `@db.XXX` series of attributes are not consumed by the ORM runtime and only affects migration/db-push. I think it's beneficial to stay so to decouple these two subsystems.

Do you think we can, instead of relying on attribute check, simply parse and determine if the value is a time not a datetime? Maybe "including : and excluding -" can be sufficient?

ymc9 (commented):
Hey @erwan-joly , many thanks for creating the issue and putting effort into the fix! Overall looks very good to me. Please check my comments.

packages/orm/src/client/crud/dialects/mysql.ts
erwan-joly: ~~I would tend to expect the string parsing to be less reliable as it would potentially break if what's in the string looks like a datetime ?~~

Nevermind — PG always emits date/datetime values year-first (`YYYY-...`) and time-only values hour-first (`HH:...`), so an anchored `^\d{2}:` check is unambiguous. Will switch to the format-detection approach.

ymc9 (approved):
Thanks for the quick update! Merging and it'll be part of the upcoming 3.6.0 release.

PR #2591, opened 2026-04-18
fix(orm): resolve delegate-inherited fields in cursor pagination

### LINKED ISSUE #2588: buildCursorFilter doesn't resolve @@delegate fields, causes "column does not exist" on cursor + orderBy(parent-field)
## Summary

`findMany({ cursor, orderBy: { <delegate-parent-field>: ... } })` fails with a Postgres error because `buildCursorFilter` references the field on the child model's alias, ignoring `@@delegate` inheritance.

## Repro

Schema:
```zmodel
model Asset {
  id        String @id @default(uuid())
  createdAt DateTime @default(now())
  assetType String
  @@delegate(assetType)
}
model Notification extends Asset {
  title String
}
```

Query:
```ts
db.notification.findMany({
  cursor: { id: someId },
  skip: 1,
  orderBy: { createdAt: "desc" },   // createdAt lives on Asset
  take: 25,
})
```

Result: `error: column Notification.createdAt does not exist`

## Isolation

| Query | Result |
|---|---|
| `orderBy: { createdAt }` (no cursor) | ✓ |
| `cursor` + `orderBy: { id }` (Notification's own column) | ✓ |
| `cursor` alone | ✓ |
| **`cursor` + `orderBy: { createdAt }`** | ✗ |

## Root cause

`@zenstackhq/orm@3.5.6`, `dist/index.js`.

`buildOrderBy` (line ~1405) resolves delegate fields via `fieldDef.originModel`:

```js
const buildFieldRef = (model, field, modelAlias) => {
  const fieldDef = requireField(this.schema, model, field);
  return fieldDef.originModel
    ? this.fieldRef(fieldDef.originModel, field, fieldDef.originModel)
    : this.fieldRef(model, field, modelAlias);
};
```

`buildCursorFilter` (line ~871) does not — it hard-codes the child alias:

```js
andFilters.push(this.eb(
  this.eb.ref(`${modelAlias}.${field}`),   // ignores originModel
  op,
  this.buildSelectModel(model, subQueryAlias).select(`${subQueryAlias}.${field}`).where(cursorFilter)
));
```

Same issue in the `subQueryAlias.${field}` reference on the inner select.

## Suggested fix

Use the same `originModel`-aware resolution as `buildOrderBy`. When a cursor or orderBy field's `fieldDef.originModel` is set, reference the parent table/alias for both the outer comparison and the inner sub-select.

## Environment

- `@zenstackhq/orm` 3.5.6
- Postgres via pg driver

genu:
## Summary

Closes #2588.

`buildCursorFilter` hardcoded the child model alias for both the outer reference and the sub-select projection, so `findMany({ cursor, orderBy: { <delegate-parent-field>: ... } })` errored with `column "Child.field" does not exist`. The fix mirrors the `originModel`-aware field resolution already used by `buildFilter` and `buildOrderBy` — when a field has `originModel`, reference the parent table alias (which `buildSelectModel` already joins using the parent's model name).

## Test plan

- [x] Added `tests/regression/test/issue-2588.test.ts` covering delegate-parent cursor, cursor+skip, mixed child/delegate orderBy, and child-only baseline.
- [x] Existing cursor tests in `tests/e2e/orm/client-api/find.test.ts` pass (18/18).
- [x] Existing delegate tests pass (23/23).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cursor-based pagination to correctly reference fields inherited from parent/delegated models, ensuring accurate ordering and stable paginated query results.

* **Tests**
  * Added a regression test suite that verifies cursor pagination and sorting when ordering by inherited/delegated fields across related models.

### DIFF
--- a/packages/orm/src/client/crud/dialects/base-dialect.ts
+++ b/packages/orm/src/client/crud/dialects/base-dialect.ts
@@ -252,12 +252,17 @@ export abstract class BaseCrudDialect<Schema extends SchemaDef> {
                 const [field, order] = orderByItems[j]!;
                 const _order = negateOrderBy ? (order === 'asc' ? 'desc' : 'asc') : order;
                 const op = j === i ? (_order === 'asc' ? '>=' : '<=') : '=';
+                // Fields inherited from a delegate base live on the base table, which is
+                // joined with its model name as alias. See buildSelectModel/buildDelegateJoin.
+                const fieldDef = requireField(this.schema, model, field);
+                const outerAlias = fieldDef.originModel ?? modelAlias;
+                const subSelectAlias = fieldDef.originModel ?? subQueryAlias;
                 andFilters.push(
                     this.eb(
-                        this.eb.ref(`${modelAlias}.${field}`),
+                        this.eb.ref(`${outerAlias}.${field}`),
                         op,
                         this.buildSelectModel(model, subQueryAlias)
-                            .select(`${subQueryAlias}.${field}`)
+                            .select(`${subSelectAlias}.${field}`)
                             .where(cursorFilter),
                     ),
                 );
--- a/tests/regression/test/issue-2588.test.ts
+++ b/tests/regression/test/issue-2588.test.ts
@@ -1,0 +1,74 @@
+import { createTestClient } from '@zenstackhq/testtools';
+import { describe, expect, it } from 'vitest';
+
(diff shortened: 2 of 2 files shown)


ymc9 (approved):
LGTM! Thanks @genu it's an important fix.

PR #2602, opened 2026-04-22
fix(server): declare h3 as optional peer dependency (#2601)

### LINKED ISSUE #2601: @zenstackhq/server 3.6.x bundles h3 types into nuxt.d.mts, breaking consumers on newer h3
## Symptom

In 3.6.x, consuming `@zenstackhq/server/nuxt` from a project on a newer `h3` (e.g. Nuxt 4 uses `h3@1.15.9`) produces a TS2345:

```
Argument of type 'H3Event<EventHandlerRequest>' is not assignable to parameter of type 'H3Event<EventHandlerRequest>'.
Type is missing the following properties: fetch, $fetch, waitUntil, captureError
```

Two `H3Event` types with the same name, structurally incompatible. Worked in 3.5.6.

## Root cause

The published `packages/server/dist/nuxt.d.mts` **inlines** `H3Event` (and its transitive graph) from whatever `h3` version sits in zenstack's own `node_modules` at build time — currently `1.15.5`. So downstream projects on newer h3 end up with a stale structural copy.

The source is fine — `packages/server/src/adapter/nuxt/handler.ts` does `import { type H3Event } from 'h3'`. The issue is the build pipeline.

### What changed in 3.6.0

Build tool was swapped from `tsup` → `tsdown`. `tsup` emitted declarations via `tsc`, preserving `import ... from 'h3'`. `tsdown` (via `rolldown-plugin-dts`) bundles declarations by default, and with no `external` list / no `dts.resolve: false`, it inlines types from any dep that isn't marked external.

### Why h3 isn't external

`packages/server/package.json` lists `h3` only as a `devDependency`, and the shared `packages/config/tsdown-config/src/index.ts` doesn't add it to `external`. Same likely applies to other adapter deps (`nuxt`, `next`, `express`, `fastify`, `hono`, `elysia`, `@sveltejs/kit`) — worth checking.

## Suggested fix

One of:
- Add `h3` (and the other adapter deps) to `peerDependencies` in `packages/server/package.json`, so tsdown treats them as external.
- Or mark them `external` explicitly in the server's `tsdown.config.ts`.
- Or disable dts bundling (`dts: { resolve: false }`) for this package.

## Repro

Nuxt 4 project with `@zenstackhq/server@3.6.1` + h3@1.15.9; pass the handler's event into any function typed with the real `H3Event` from `h3`. Downgrading to 3.5.6 or narrowing the consumer's param type to `{ headers: Headers }` avoids the error.

genu:
## Summary

Fixes #2601. In 3.6.x, `dist/nuxt.d.mts` inlined `H3Event` and its transitive type graph from the pinned devDependency `h3@1.15.5`, causing TS2345 for consumers on a newer `h3` (e.g. Nuxt 4 on `h3@1.15.9`).

Root cause: the `tsup` → `tsdown` migration changed dts emission. `rolldown-plugin-dts` bundles declarations and externalizes only what's declared in `dependencies` / `peerDependencies`. `h3` was only in `devDependencies`, so its types got inlined.

This PR adds `h3` as an optional `peerDependency` (matching how `nuxt`, `next`, `express`, `fastify`, `hono`, `elysia`, `@sveltejs/kit`, `fastify-plugin` are already declared), keeping the fix consistent with the rest of the adapter deps and requiring no build-config changes.

## Test plan

- [x] Rebuild `@zenstackhq/server`; confirm `dist/nuxt.d.mts` now emits `import { H3Event, EventHandlerRequest } from "h3"` instead of inlined interfaces (size dropped to ~0.88 kB).
- [ ] Verify in a Nuxt 4 consumer (`h3@1.15.9`) that passing the handler's event into functions typed with the real `H3Event` no longer produces TS2345.

## Summary by CodeRabbit

* **Chores**
  * Added support for an optional dependency to enhance compatibility with different runtime environments and project configurations.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -170,6 +170,7 @@
         "express": "^5.0.0",
         "fastify": "^5.0.0",
         "fastify-plugin": "^5.0.0",
+        "h3": "^1.15.0",
         "hono": "^4.6.0",
         "next": "^15.0.0 || ^16.0.0",
         "nuxt": "^4.0.0",
@@ -189,6 +190,9 @@
             "optional": true
         },
         "elysia": {
+            "optional": true
+        },
+        "h3": {
             "optional": true
         },
         "hono": {


ymc9 (approved):
Yes, it should indeed be an optional peer. Thanks for the fix!

PR #2617, opened 2026-04-25
fix(better-auth): array support

### LINKED ISSUE #2615: @zenstackhq/better-auth adapter serializes Better Auth string[] fields before ZenStack validation
### What happened?

When using `@zenstackhq/better-auth` with Better Auth's OAuth provider, dynamic client registration fails for models that contain native array fields.

In our setup, Better Auth's OAuth provider declares several fields as `type: "string[]"`, and the ZenStack Better Auth schema generator emits matching ZenStack native array fields such as:

```zmodel
model OauthClient {
  redirectUris           String[]
  scopes                 String[]
  grantTypes             String[]
  responseTypes          String[]
  postLogoutRedirectUris String[]
}
```

However, at runtime, `@zenstackhq/better-auth` creates the Better Auth adapter without setting or forwarding `supportsArrays`:

```ts
const adapterOptions = {
  config: {
    adapterId: "zenstack",
    adapterName: "ZenStack Adapter",
    usePlural: config.usePlural ?? false,
    debugLogs: config.debugLogs ?? false,
    transaction: ...
  },
  adapter: createCustomAdapter(db)
};
```

Better Auth core defaults `supportsArrays` to `false`. For schema fields with `type: "string[]"` or `type: "number[]"`, the adapter factory then `JSON.stringify`s array values before calling the underlying adapter.

That means ZenStack receives strings for fields whose generated schema expects arrays, and validation fails before the DB write.

Example failure from dynamic OAuth client registration:

```text
Invalid input: expected array, received string
path: scopes

Invalid input: expected array, received string
path: redirectUris

Invalid input: expected array, received string
path: grantTypes

Invalid input: expected array, received string
path: responseTypes
```

### Versions

```text
@zenstackhq/better-auth: 3.5.6
better-auth: 1.5.4
@better-auth/core: 1.5.4
@better-auth/oauth-provider: 1.5.4
PostgreSQL provider
```

I also looked at the recent schema-generator changes in #2614. They look useful for schema generation/import behavior, but this runtime issue appears separate because the adapter config passed to `createAdapterFactory` still doesn't declare array support.

### Minimal reproduction shape

1. Configure Better Auth with ZenStack adapter and the OAuth provider.
2. Generate/port the Better Auth OAuth provider schema into ZenStack models, where OAuth provider `string[]` fields become `String[]`.
3. Call OAuth dynamic client registration with array fields:

```json
{
  "redirect_uris": ["http://localhost:3000/callback"],
  "scope": "openid profile email",
  "grant_types": ["authorization_code", "refresh_token"],
  "response_types": ["code"]
}
```

4. The create call fails because Better Auth transformed those arrays into JSON strings before `db.oauthClient.create(...)`.

### Expected behavior

For providers/schemas that support native arrays, the ZenStack Better Auth adapter should preserve array values when Better Auth schema fields are `string[]` / `number[]`.

A possible fix would be to set or forward Better Auth's `supportsArrays` adapter option, likely defaulting to `true` for PostgreSQL/native-array ZenStack schemas, or exposing it explicitly through `AdapterConfig`.

Alternatively, if the adapter intentionally does not support native arrays, the schema generator should not emit `String[]` for Better Auth array fields, because the generated schema and runtime adapter behavior currently disagree.

pkudinov:
## Summary

Fixes #2615.

This updates the ZenStack Better Auth adapter so Better Auth array fields are preserved for PostgreSQL-backed ZenStack schemas instead of being JSON-stringified before ZenStack validation. The adapter config now exposes `supportsArrays`, defaulting to native array support for PostgreSQL and disabled array support for SQLite.

Schema generation now uses the same capability decision as runtime: `string[]` and `number[]` become native `String[]` and `Int[]` when arrays are supported, and fall back to `Json` fields when they are not.

This problem prevents normal use of [Better Auth OAuth Provider plugin](https://better-auth.com/docs/plugins/oauth-provider).

## Validation

- `pnpm --filter @zenstackhq/better-auth lint`
- `pnpm --filter @zenstackhq/better-auth build`
- `pnpm --filter @zenstackhq/better-auth test`

## Release Notes

* **Refactor**
  * Enhanced Better Auth adapter configuration to intelligently handle array-type fields based on database provider—arrays are natively supported on PostgreSQL, while other databases use JSON serialization for compatibility.

* **Tests**
  * Added test suite validating array field handling and schema generation behavior across different database configurations.

### DIFF
--- a/packages/auth-adapters/better-auth/src/adapter.ts
+++ b/packages/auth-adapters/better-auth/src/adapter.ts
@@ -1,5 +1,5 @@
 import type { BetterAuthOptions } from '@better-auth/core';
-import type { DBAdapter, DBAdapterDebugLogOption, Where } from '@better-auth/core/db/adapter';
+import type { DBAdapter, Where } from '@better-auth/core/db/adapter';
 import { BetterAuthError } from '@better-auth/core/error';
 import type { ClientContract, ModelOperations, UpdateInput } from '@zenstackhq/orm';
 import type { GetModels, SchemaDef } from '@zenstackhq/orm/schema';
@@ -8,30 +8,9 @@ import {
     type AdapterFactoryCustomizeAdapterCreator,
     type AdapterFactoryOptions,
 } from 'better-auth/adapters';
+import { getSupportsArrays, type AdapterConfig } from './config';
 
-/**
- * Options for the ZenStack adapter factory.
- */
-export interface AdapterConfig {
-    /**
-     * Database provider
-     */
-    provider: 'sqlite' | 'postgresql';
-
-    /**
-     * Enable debug logs for the adapter
-     *
-     * @default false
-     */
-    debugLogs?: DBAdapterDebugLogOption | undefined;
-
-    /**
-     * Use plural table names
-     *
-     * @default false
-     */
-    usePlural?: boolean | undefined;
-}
+export type { AdapterConfig } from './config';
 
 /**
  * Create a Better-Auth adapter for ZenStack ORM.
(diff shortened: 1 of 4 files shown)


ymc9 (approved):
Looks great to me! Thanks for making this fix @pkudinov .

PR #2632, opened 2026-04-30
fix(orm): coerce ISO strings on DateTime input, with strictDateInput opt-in (#2631)

### LINKED ISSUE #2631: ZenStack 3.6: strict Date input validator rejects ISO strings (regression vs 3.4 / Prisma)
## Background

After upgrading from `@zenstackhq/orm@3.4.6` → `@zenstackhq/orm@3.6.4`, code paths that previously passed an ISO string to a `DateTime` field now fail with:

```
Validation error:
  Invalid input: expected object, received array at "data.<relation>.createMany.data"
  Invalid ISO datetime at "data.<relation>.createMany.data[0].<field>"
  Invalid ISO date at "data.<relation>.createMany.data[0].<field>"
  Invalid input: expected date, received string at "data.<relation>.createMany.data[0].<field>"

  caused by: ZodError
    at InputValidator.validate (.../validator.ts:294)
    at InputValidator.validateCreateArgs (.../validator.ts:80)
    at CreateOperationHandler.handle (.../create.ts:13)
```

The affected fields are backed by `@db.Time(6)` / `@db.Date`. The inputs were valid ISO 8601 strings (e.g. `"09:00:00"`, `"2024-04-01"`, or `"2024-04-01T00:00:00.000Z"`).

## Minimal repro

```prisma
model Parent {
  id       String  @id @default(uuid())
  children Child[]
}

model Child {
  id       String   @id @default(uuid())
  parentId String
  open     DateTime @db.Time(6)
  parent   Parent   @relation(fields: [parentId], references: [id])
}
```

```ts
// 3.4.6 — works (Prisma coerces the string → Date)
// 3.6.4 — throws (zod union rejects all three branches)
await db.parent.create({
  data: {
    children: {
      createMany: {
        data: [{ open: "09:00:00" }],
      },
    },
  },
});
```

## What I think is happening

In 3.6.x the input layer builds a Zod schema where `DateTime` resolves to a union roughly equivalent to `z.union([z.iso.datetime(), z.iso.date(), z.date()])`. None of the three branches match a bare `@db.Time` value like `"09:00:00"`, so the union rejects.

3.4.x went through a Prisma-compatible coercion layer that accepted ISO strings and converted them to `Date` for the engine. The new behavior is technically more correct (a `DateTime` field semantically expects a `Date`), but it's a breaking change I couldn't find in the 3.5 / 3.6 release notes.

Related to #2590, which fixed the *read* side of `@db.Time` (raw strings being silently returned instead of `Date`s and crashing downstream `.getUTCFullYear()` calls). The input side now needs every caller updated.

## Suggested resolution

Pick one:

1. **Restore Prisma-compatible coercion** — accept ISO strings on `DateTime` inputs and convert to `Date`. Matches Prisma's longstanding behavior.
2. **Add a config option**, e.g. `ClientOptions.strictDateInputs: boolean` (default `false` to match 3.4.x). Lets users opt into the strict mode while keeping backward compat by default.
3. **Keep the strict behavior** but call it out as a breaking change in the release notes with a migration guide entry (recommended fix: `new Date(isoString)` or equivalent).

Filing this neutrally — could go either way. Thanks for ZenStack!

## Versions

- `@zenstackhq/orm`: 3.6.4 (was 3.4.6 working)
- `@zenstackhq/cli`: 3.6.4
- `prisma`: 7.8.0
- `node`: 22.x

erwan-joly:
Fixes #2631.

## Background

The strict zod input validator introduced in 3.5+ broke every caller passing ISO strings to `DateTime` fields, including bare time-only strings like `"09:00:00"` for `@db.Time` columns. Earlier versions silently coerced these via Prisma's input layer; the new validator rejects them outright. Existing user code that worked unchanged across years of Prisma → ZenStack v2 → ZenStack v3.4 suddenly fails with `Invalid input: expected date, received string`.

## Approach

Per option 2 from the issue, this restores Prisma-compatible coercion as the **default** while leaving strict validation available behind a new opt-in flag. That keeps existing callers working while preserving the stricter semantics for users who want them:

```ts
const db = new ZenStackClient(schema, {
  dialect: ...,
  strictDateInput: true, // opt in to date-objects-only validation
});
```

Default (`strictDateInput: false` or unset) accepts:
- `Date` objects (already worked)
- ISO datetime strings, e.g. `"2024-01-15T10:30:00.000Z"` (already worked)
- ISO date strings, e.g. `"2024-01-15"` (already worked in orm factory; new in standalone zod factory)
- ISO time-only strings, e.g. `"09:30:00"`, `"09:30:00.123"`, `"09:30:00Z"`, `"09:30:00+12:00"` (new)

All string forms are coerced to a `Date` for the engine. Time-only strings are anchored to the Unix epoch (`1970-01-01T<time>`), matching the existing OID-1083 read-side behavior introduced in #2590.

## Files changed

- `packages/orm/src/client/options.ts` — new `strictDateInput?: boolean` option with JSDoc explaining the default and tradeoff
- `packages/orm/src/client/zod/factory.ts` — new exported helper `coercedDateTimeSchema()`; `makeDateTimeValueSchema` switches on `strictDateInput`
- `packages/zod/src/factory.ts` — same coercion applied in the standalone factory (no setting — these schemas are typically used for form validation where coercion is more important)
- `packages/zod/test/factory.test.ts` — regression tests for the four accepted forms plus a non-parseable rejection case, all referencing #2631

## Tests

```
✓ accepts DateTime as a Date object
✓ accepts DateTime as an ISO datetime string
✓ accepts DateTime as an ISO date string (#2631)
✓ accepts DateTime as a bare time-only string for @db.Time fields (#2631)
✓ accepts DateTime as a time-only string with timezone (#2631)
✓ rejects DateTime as a non-parseable string
```

The strict path (when `strictDateInput: true`) keeps the existing `z.union([z.iso.datetime(), z.iso.date(), z.date()])` behavior unchanged, so users who've adopted the strict semantics on 3.5/3.6 are unaffected.

Happy to adjust naming (`strictDateInput` vs `strictDateInputs` vs anything else) or re-shape per review.

## Summary by CodeRabbit

* **New Features**
  * Improved DateTime field input handling with support for multiple formats: native Date objects, ISO datetime strings, ISO date strings, and time-only strings.

* **Tests**
  * Added regression test coverage for DateTime input coercion scenarios.

### DIFF
--- a/packages/orm/src/client/options.ts
+++ b/packages/orm/src/client/options.ts
@@ -203,6 +203,16 @@ export type ClientOptions<Schema extends SchemaDef> = QueryOptions<Schema> & {
     validateInput?: boolean;
 
     /**
+     * Whether to require `Date` objects (rather than ISO strings) for `DateTime` field inputs. Defaults
+     * to `false`, matching Prisma's longstanding behavior of coercing ISO strings — including bare
+     * time-only strings like `"09:00:00"` for `@db.Time` fields — to `Date`.
+     *
+     * Set to `true` to opt into strict input validation that rejects all string forms.
+     * @see https://github.com/zenstackhq/zenstack/issues/2631
+     */
+    strictDateInput?: boolean;
+
+    /**
      * Whether to use compact alias names (e.g., "$$t1", "$$t2") when transforming ORM queries to SQL.
      * Defaults to `true`.
      *
--- a/packages/orm/src/client/options.ts
+++ b/packages/orm/src/client/options.ts
@@ -203,16 +203,6 @@ export type ClientOptions<Schema extends SchemaDef> = QueryOptions<Schema> & {
     validateInput?: boolean;
 
     /**
-     * Whether to require `Date` objects (rather than ISO strings) for `DateTime` field inputs. Defaults
-     * to `false`, matching Prisma's longstanding behavior of coercing ISO strings — including bare
-     * time-only strings like `"09:00:00"` for `@db.Time` fields — to `Date`.
-     *
-     * Set to `true` to opt into strict input validation that rejects all string forms.
-     * @see https://github.com/zenstackhq/zenstack/issues/2631
-     */
-    strictDateInput?: boolean;
-
-    /**
      * Whether to use compact alias names (e.g., "$$t1", "$$t2") when transforming ORM queries to SQL.
      * Defaults to `true`.
      *
--- a/packages/orm/src/client/zod/factory.ts
+++ b/packages/orm/src/client/zod/factory.ts
@@ -85,11 +85,14 @@ export function createQuerySchemaFactory(clientOrSchema: any, options?: any) {
 }
 
 /**
- * Builds a `DateTime` value schema that accepts a `Date` object or an ISO
- * datetime / date / time-only string and coerces it to a `Date`. Time-only
- * strings (e.g. `"09:00:00"` for `@db.Time` fields) are anchored to the Unix
- * epoch. Strings that don't parse fall through and are rejected by `z.date()`
- * with the standard error.
+ * Builds a `DateTime` value schema that accepts a `Date` object or any string
+ * the JS `Date` constructor parses, and coerces it to a `Date`. ISO datetime,
+ * ISO date, and time-only strings (e.g. `"09:00:00"` for `@db.Time` fields,
+ * anchored to the Unix epoch) are the documented happy paths; other formats
+ * accepted by `new Date(...)` also pass through, mirroring Prisma's pre-3.5
+ * behaviour. Strings the engine can't parse fall through and are rejected by
+ * `z.date()` with the standard error. Callers wanting strict ISO-or-Date
+ * validation should set `ClientOptions.strictDateInput: true`.
  *
  * Used when `ClientOptions.strictDateInput` is left at its default (`false`).
  * @see https://github.com/zenstackhq/zenstack/issues/2631
--- a/packages/orm/src/client/zod/factory.ts
+++ b/packages/orm/src/client/zod/factory.ts
@@ -85,6 +85,29 @@ export function createQuerySchemaFactory(clientOrSchema: any, options?: any) {
 }
 
 /**
+ * Builds a `DateTime` value schema that accepts a `Date` object or an ISO
+ * datetime / date / time-only string and coerces it to a `Date`. Time-only
+ * strings (e.g. `"09:00:00"` for `@db.Time` fields) are anchored to the Unix
+ * epoch. Strings that don't parse fall through and are rejected by `z.date()`
+ * with the standard error.
+ *
+ * Used when `ClientOptions.strictDateInput` is left at its default (`false`).
+ * @see https://github.com/zenstackhq/zenstack/issues/2631
+ */
+export function coercedDateTimeSchema(): ZodType {
+    return z.preprocess((val) => {
+        if (typeof val !== 'string') return val;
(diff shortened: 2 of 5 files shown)


erwan-joly:
Thanks for the catch — pushed [e8cc785e](https://github.com/erwan-joly/zenstack/commit/e8cc785e) addressing both points:

1. **Strict mode now actually strict** — switched the strict branch from `z.union([z.iso.datetime(), z.iso.date(), z.date()])` to plain `z.date()` so it matches the JSDoc.

2. **Standalone factory reverted** — that was breaking the rpc-openapi baseline test because the `z.preprocess` shape differs from the prior union. Reverted to `z.union([z.date(), z.iso.datetime()])` for that path. The ORM-side coercion in `packages/orm/src/client/zod/factory.ts` (the runtime validator users hit) keeps the fix; standalone consumers who want coercion can layer it themselves.

3. **Strict-mode tests added** — new `tests/regression/test/issue-2631.test.ts` covers the default behavior end-to-end (Date, ISO datetime, ISO date, time-only with and without TZ, non-parseable rejection) and the strict contract (Date-only, all string forms rejected).

packages/orm/src/client/zod/factory.ts
```diff
@@ -84,6 +84,29 @@ export function createQuerySchemaFactory(clientOrSchema: any, options?: any) {
...
+export function coercedDateTimeSchema(): ZodType {
+    return z.preprocess((val) => {
+        if (typeof val !== 'string') return val;
+        if (/^\d{2}:\d{2}(?::\d{2}(?:\.\d+)?)?(?:Z|[+-]\d\d(?::\d\d)?)?$/.test(val)) {
+            const hasTz = val.endsWith('Z') || /[+-]\d\d(?::\d\d)?$/.test(val);
+            const d = new Date(`1970-01-01T${val}${hasTz ? '' : 'Z'}`);
+            return isNaN(d.getTime()) ? val : d;
+        }
+        const d = new Date(val);
+        return isNaN(d.getTime()) ? val : d;
```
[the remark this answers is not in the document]
erwan-joly: The permissive fallback is intentional — Prisma accepts anything `new Date()` parses, and #2631 is specifically about restoring that compatibility. Tightening the fallback to ISO-only would re-introduce the original bug for callers passing other Date-parseable formats. Updated the JSDoc on `coercedDateTimeSchema` in 6956c7e to describe the actual contract ("any string the JS `Date` constructor parses") and point readers at `strictDateInput: true` for callers who want strict ISO-or-Date validation.

packages/orm/src/client/options.ts
```diff
@@ -202,6 +202,16 @@ export type ClientOptions<Schema extends SchemaDef> = QueryOptions<Schema> & {
...
 
+    /**
+     * Whether to require `Date` objects (rather than ISO strings) for `DateTime` field inputs. Defaults
+     * to `false`, matching Prisma's longstanding behavior of coercing ISO strings — including bare
+     * time-only strings like `"09:00:00"` for `@db.Time` fields — to `Date`.
+     *
+     * Set to `true` to opt into strict input validation that rejects all string forms.
+     * @see https://github.com/zenstackhq/zenstack/issues/2631
+     */
+    strictDateInput?: boolean;
```
ymc9: Hey @erwan-joly , thanks for making continuous improvements. Date time handling is nasty ...

I'm thinking the new behavior (more accommodating) is probably preferred for most. Maybe we can drop this config altogether?

erwan-joly: Sure thing, didn’t want to “break” it if it was on purpose I see now it wasn’t so will drop the configuration altogether

ymc9 (approved):

PR #2634, opened 2026-04-30
fix(orm): format Date as HH:MM:SS for @db.Time / @db.Timetz columns (#2633)

### LINKED ISSUE #2633: ZenStack 3.6: writes to @db.Time columns fail with PG 22007 (Date serialized as ISO datetime)
## Summary

The write side of `@db.Time` is broken in 3.6.x. When a JS `Date` is sent to a `@db.Time` column, the underlying kysely pg driver serializes the parameter as the ISO datetime returned by `Date.prototype.toISOString()` (e.g. `\"2026-04-29T21:00:00.000Z\"`). Postgres TIME columns reject the string with:

```
error: invalid input syntax for type time: \"2026-04-29T21:00:00.000Z\"
  code: 22007
```

This is the write-side counterpart to #2590 (which fixed `@db.Time` reads silently returning raw strings).

## Repro

```prisma
model Parent {
  id       String  @id @default(uuid())
  children Child[]
}

model Child {
  id       String   @id @default(uuid())
  parentId String
  open     DateTime @db.Time(6)
  close    DateTime @db.Time(6)
  parent   Parent   @relation(fields: [parentId], references: [id])
}
```

```ts
await db.parent.create({
  data: {
    children: {
      createMany: {
        data: [{
          open:  new Date('1970-01-01T09:00:00Z'),
          close: new Date('1970-01-01T16:00:00Z'),
        }],
      },
    },
  },
});
```

The validator (after #2632) accepts the `Date`, but the resulting INSERT throws `22007`.

## What I think is happening

`@zenstackhq/orm` builds the query through kysely's pg dialect. Kysely's parameter binding is column-type-agnostic by design — when given a `Date`, it serializes via the standard pg-types path which produces an ISO timestamp string. Postgres' TIME column type rejects ISO timestamp input; it expects `HH:MM:SS` (with optional fractional seconds and timezone offset).

For `@db.Date` and `@db.Timestamp` this isn't a problem — Postgres accepts the ISO string for both. Only TIME is strict.

## Suggested fix

Either:

1. **Pre-serialize at the ZenStack input layer** for `@db.Time` columns: detect the column's native type and convert the `Date` (or string) into `HH:MM:SS` before kysely builds the query.
2. **Pre-serialize at the dialect layer**: extend `PostgresCrudDialect` to wrap `Date` parameters bound to TIME columns with a `sql.lit(...)` value formatted as TIME.
3. **Document & expose a helper** for users to opt into manual TIME formatting (similar to how `@db.Bytes` requires explicit `Uint8Array`).

Option 1 keeps the user surface unchanged — they pass a `Date` (or, after #2632, an ISO time-only string) and the query just works. Happy to send a PR if there's interest.

## Workaround we're using

Migrating our TIME columns to TIMESTAMP. The application only reads time-of-day from the field anyway, so the date prefix is harmless. Not viable for users who genuinely need TIME semantics (range queries, etc.).

## Versions

- `@zenstackhq/orm`: 3.6.4
- `prisma`: 7.8.0
- `kysely`: 0.28.16
- `pg`: 8.20.0
- `node`: 22.x

erwan-joly:
Fixes #2633.

## Background

`PostgresCrudDialect.transformInput` converted every `DateTime` input to `Date.prototype.toISOString()`, which Postgres TIME / TIMETZ columns reject with:

```
error: invalid input syntax for type time: "2026-04-29T21:00:00.000Z"
code: 22007
```

This is the write-side counterpart to #2590 (which fixed `@db.Time` reads silently returning raw strings).

## Approach

The dialect now inspects the field's `@db.*` attribute and routes time-of-day values to a `formatTimeOfDay` helper that emits PG-acceptable `HH:MM:SS.fff` (or `HH:MM:SS.fff+ZZ:ZZ` for TIMETZ). All other `DateTime` columns (TIMESTAMP / TIMESTAMPTZ / DATE) keep the existing ISO behaviour — Postgres accepts ISO input for those types natively.

To reach the field's attributes from the dialect, `BaseCrudDialect.transformInput` gains an optional 4th parameter, `fieldDef?: FieldDef`. The default behaviour is unchanged for callers that don't pass it.

## Files changed

- `packages/orm/src/client/crud/dialects/base-dialect.ts` — adds optional `fieldDef` param to `transformInput`.
- `packages/orm/src/client/crud/dialects/postgresql.ts` — adds `formatTimeOfDay` helper; `transformInput` reads `@db.Time` / `@db.Timetz` and routes through it.
- `packages/orm/src/client/crud/operations/base.ts` — threads `fieldDef` through every `transformInput` write call site (create, createMany, update, upsert, automatic `updatedAt`).
- `tests/regression/test/issue-2633.test.ts` — regression covering single create and nested `createMany` against both `@db.Time` and `@db.Timetz`, plus a round-trip read.

## Repro

```prisma
model TradingHour {
  id    Int      @id @default(autoincrement())
  open  DateTime @db.Time(6)
  close DateTime @db.Time(6)
}
```

```ts
// 3.6.4 — throws PG 22007
// this PR — works
await db.tradingHour.create({
  data: {
    open:  new Date('1970-01-01T09:00:00.000Z'),
    close: new Date('1970-01-01T16:30:00.000Z'),
  },
});
```

## Notes

- Filter call sites that compare against TIME columns benefit from the same coercion via the array-filter path. The other base-dialect filter call sites (top-level scalar comparisons, JSON null checks, etc.) keep the simpler 3-arg form because they don't have field-level context handy. If a user files a follow-up about `WHERE open = ?` against TIME, that's the natural extension point.
- Pairs naturally with #2632 (input validator coercion) — together they restore parity with Prisma's longstanding `@db.Time` ergonomics.

## Summary by CodeRabbit

* **New Features**
  * Dialect-aware input handling so native column metadata is respected during input transformation.

* **Bug Fixes**
  * Improved DateTime serialization for PostgreSQL time-type columns to preserve time-only values and format them correctly instead of forcing full ISO datetime.

* **Tests**
  * Added regression test verifying Date inputs round-trip correctly for native PostgreSQL time/timetz columns.

### DIFF
--- a/packages/orm/src/client/crud/dialects/base-dialect.ts
+++ b/packages/orm/src/client/crud/dialects/base-dialect.ts
@@ -87,8 +87,13 @@ export abstract class BaseCrudDialect<Schema extends SchemaDef> {
 
     /**
      * Transforms input value before sending to database.
+     *
+     * `fieldDef` is optional so existing callers that don't have it stay
+     * source-compatible. Dialects can use it to inspect `@db.*` native-type
+     * attributes (e.g. to format `@db.Time` values as `HH:MM:SS` rather than
+     * full ISO timestamps).
      */
-    transformInput(value: unknown, _type: BuiltinType, _forArrayField: boolean) {
+    transformInput(value: unknown, _type: BuiltinType, _forArrayField: boolean, _fieldDef?: FieldDef) {
         return value;
     }
 
@@ -523,7 +528,7 @@ export abstract class BaseCrudDialect<Schema extends SchemaDef> {
             }
 
             invariant(fieldDef.array, 'Field must be an array type to build array filter');
-            const value = this.transformInput(_value, fieldType, true);
+            const value = this.transformInput(_value, fieldType, true, fieldDef);
 
             let receiver = fieldRef;
             if (isEnum(this.schema, fieldType)) {
--- a/packages/orm/src/client/crud/dialects/postgresql.ts
+++ b/packages/orm/src/client/crud/dialects/postgresql.ts
@@ -17,6 +17,18 @@ import type { ClientOptions } from '../../options';
 import { isEnum, isTypeDef } from '../../query-utils';
 import { LateralJoinDialectBase } from './lateral-join-dialect-base';
 
+/**
+ * Formats a JS `Date` as a Postgres TIME / TIMETZ literal (`HH:MM:SS.fff`,
+ * optionally with `+ZZ:ZZ` for TIMETZ). Reads UTC components so the value
+ * round-trips with ISO-input parsing — callers anchor time-only inputs to
+ * the Unix epoch.
+ */
+function formatTimeOfDay(date: Date, withTimezone: boolean): string {
+    const pad = (n: number, w = 2) => String(n).padStart(w, '0');
+    const time = `${pad(date.getUTCHours())}:${pad(date.getUTCMinutes())}:${pad(date.getUTCSeconds())}.${pad(date.getUTCMilliseconds(), 3)}`;
+    return withTimezone ? `${time}+00:00` : time;
+}
+
 export class PostgresCrudDialect<Schema extends SchemaDef> extends LateralJoinDialectBase<Schema> {
     private static typeParserOverrideApplied = false;
 
@@ -154,7 +166,7 @@ export class PostgresCrudDialect<Schema extends SchemaDef> extends LateralJoinDi
 
     // #region value transformation
 
-    override transformInput(value: unknown, type: BuiltinType, forArrayField: boolean): unknown {
+    override transformInput(value: unknown, type: BuiltinType, forArrayField: boolean, fieldDef?: FieldDef): unknown {
         if (value === undefined) {
             return value;
         }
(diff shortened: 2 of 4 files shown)


packages/orm/src/client/crud/dialects/base-dialect.ts
```diff
@@ -87,8 +87,13 @@ export abstract class BaseCrudDialect<Schema extends SchemaDef> {
...
     /**
      * Transforms input value before sending to database.
+     *
+     * `fieldDef` is optional so existing callers that don't have it stay
+     * source-compatible. Dialects can use it to inspect `@db.*` native-type
+     * attributes (e.g. to format `@db.Time` values as `HH:MM:SS` rather than
+     * full ISO timestamps).
      */
-    transformInput(value: unknown, _type: BuiltinType, _forArrayField: boolean) {
+    transformInput(value: unknown, _type: BuiltinType, _forArrayField: boolean, _fieldDef?: FieldDef) {
```
erwan-joly, line 96: Very close to what we initially did and reverted on a previous PR — but I don't think there's a better way to solve this here. The shape-detect trick from #2590 worked for the read side because PG emits time-only values (`09:30:00`) and datetimes (`2026-04-29 ...`) with distinguishable prefixes. On the write side the input is always a `Date` object, which carries no shape clue about whether the destination column is `time` or `timestamp`, so we genuinely need the field metadata to format correctly. Open to alternatives if you see one.

ymc9 (approved):
Looks good to me!

I don't have a better solution than inspecting the `@db.Time` attribute 😂. `Time` may deserve a built-in type for cleaner handling, but that's gonna be a bigger change and probably not worth it yet.

PR #2652, opened 2026-05-06
fix(cli): add missing opposite relation fields during db pull when multiple FKs target the same model

svetch:
Back-reference relation fields (the opposite side of a relation, with @relation but no `fields` arg) were silently skipped during the merge phase of `db pull` when no matching field existed in the original schema. This caused models like `Users` that are referenced by many tables (e.g., via `user_created`/`user_updated` FKs) to be missing their back-reference fields after pulling.

The fix adds relation-name-based matching as a new step in the field matching algorithm, and removes the blanket early-skip that discarded all unmatched back-references. Named back-references that don't match any existing field are now correctly added as new fields.

## Release Notes

* **Bug Fixes**
  * Improved database pull reconciliation to correctly handle back-reference relation fields, preventing them from being incorrectly marked as deleted.
  * Back-reference fields can now be reconciled by relation name, enabling proper schema restoration when pulling from database.

* **Tests**
  * Added test coverage for pulling schemas with multiple relations from the same model.

### DIFF
--- a/packages/cli/src/actions/db.ts
+++ b/packages/cli/src/actions/db.ts
@@ -1,5 +1,5 @@
 import { formatDocument, ZModelCodeGenerator } from '@zenstackhq/language';
-import { DataModel, Enum, type DataField, type Model } from '@zenstackhq/language/ast';
+import { DataModel, Enum, isDataField, type DataField, type Model } from '@zenstackhq/language/ast';
 import colors from 'colors';
 import fs from 'node:fs';
 import path from 'node:path';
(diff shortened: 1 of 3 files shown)


ymc9:
Hi @svetch , the build failure was likely caused by an old issue on dev branch. Could you merge from upstream and push? Thanks!

packages/cli/src/actions/pull/utils.ts
```diff
@@ -122,6 +122,20 @@ export function getRelationFkName(decl: DataField): string | undefined {
...
+/**
+ * Gets the relation name from the @relation attribute's first positional argument.
+ * e.g., @relation('myRelation', fields: [...], references: [...]) -> "myRelation"
+ * e.g., @relation(fields: [...], references: [...]) -> undefined
+ * e.g., @relation('backRef') -> "backRef"
+ */
+export function getRelationName(decl: DataField): string | undefined {
+    const relationAttr = decl?.attributes?.find((a) => a.decl?.ref?.name === '@relation');
+    if (!relationAttr) return undefined;
+    const firstPositionalArg = relationAttr.args.find((a) => !a.name);
```
ymc9: Positional parameters can still be passed with name like `@relation(name: 'foo', ...)`. The most reliable way to get an argument is to use its resolved parameter's name. The language package has a `getAttributeArgLiteral` helper for it.

packages/cli/src/actions/db.ts
```diff
@@ -314,11 +326,24 @@ async function runPull(options: PullOptions) {
...
                     }
 
+                    if (originalFields.length === 0) {
+                        // Try matching by relation name (the first positional arg in @relation)
+                        // This is essential for back-reference fields that only have a relation name
+                        const newRelName = getRelationName(f as any);
+                        if (newRelName) {
+                            originalFields = originalDataModel.fields.filter(
+                                (d) =>
+                                    d.$type === 'DataField' &&
```
ymc9: Using `isDataField(d)` is more idiomatic.

packages/cli/src/actions/db.ts
```diff
@@ -520,6 +545,18 @@ async function runPull(options: PullOptions) {
                         );
                         if (matchByFkName) return false;
 
+                        // Try matching by relation name (for named back-reference fields)
+                        const originalRelName = getRelationName(f as any);
```
ymc9: Is this cast necessary?

ymc9 (commented):
Hi @svetch , thanks for making this fix. Please take a look at my comments. Thanks!

ymc9 (approved):
LGTM. Thanks!

Issue #2644, opened 2026-05-04
zenstack v2: dependency updates needed and version mismatches


sw34:
**Description and expected behavior**
I need to continue use of zenstack v2 as I currently require SQL Server support.  npm audit is reporting that some dependencies of zenstack are using vulnerable versions.

he other issue is that both the "zenstack" and "@zenstackhq/runtime" packages report that 2.22.1 is the "latest" version, but should be 2.22.2.

**Screenshots**

<img width="878" height="213" alt="Image" src="https://github.com/user-attachments/assets/01d47641-1d48-4a93-8152-5510ebe38a47" />

showing one of a couple like this

<img width="396" height="25" alt="Image" src="https://github.com/user-attachments/assets/b26ccf08-cba7-4481-9e6a-35700a4d4ec2" />

<img width="396" height="21" alt="Image" src="https://github.com/user-attachments/assets/553ed87d-8780-4304-8a3e-d52ca56fd9f5" />

**Environment (please complete the following information):**

-   ZenStack version: 2.22.2
-   Database type: SQL Server]
-   Node.js/Bun version: Node v22.15.1
-   Package manager: npm

**Additional context**

ymc9:
Hi @sw34 , I went through audits and updated several high and moderate dependencies - uuid, yaml, tmp. You may still see a report about "underscore", which is an indirect dependency brought by the upstream "logic-solver" package that we don't control. You may have to configure a version override explicitly in your project for it.

Please upgrade to 2.22.3 and let me know if you still see other audit issues.

sw34:
Thank you @ymc9 - this resolved most issues.

I don't see the "logic-solver" package, or at least I dont think I do, my only remaining audit is:

```
lodash  <=4.17.23
Severity: high
Lodash has Prototype Pollution Vulnerability in `_.unset` and `_.omit` functions - https://github.com/advisories/GHSA-xxjr-mmjv-4gpg
lodash vulnerable to Code Injection via `_.template` imports key names - https://github.com/advisories/GHSA-r5fr-rjxr-66jc
lodash vulnerable to Prototype Pollution via array path bypass in `_.unset` and `_.omit` - https://github.com/advisories/GHSA-f23m-r3pf-42rh
fix available via `npm audit fix --force`
Will install zenstack@0.5.0, which is a breaking change
node_modules/lodash
  @chevrotain/cst-dts-gen  10.0.0 - 10.5.0
  Depends on vulnerable versions of @chevrotain/gast
[...]

PR #2661, opened 2026-05-08
allow large pagesize limits on rest api

### LINKED ISSUE #2660: [BUG] zenstack v2 rest api limits pagesize to 100
the rest api limits page size to a max of 100 if no option is set at endpoint creation

Proposed pull request only limits to pageSize option if it was explicitly set

-   ZenStack version: [e.g., 2.2.1]
-   Database type: mssql
-   Node.js/Bun version: node 22
-   Package manager: npm

sw34:
resolves #2660

### DIFF
--- a/packages/server/src/api/rest/index.ts
+++ b/packages/server/src/api/rest/index.ts
@@ -1672,7 +1672,9 @@ class RequestHandler extends APIHandlerBase {
             if (isNaN(limit) || limit <= 0) {
                 limit = pageSizeOption;
             }
-            limit = Math.min(pageSizeOption, limit);
+            if (this.options.pageSize) {
+                limit = Math.min(this.options.pageSize, limit);
+            }
         }
 
         return { offset, limit };
--- a/packages/server/src/api/rest/index.ts
+++ b/packages/server/src/api/rest/index.ts
@@ -1672,8 +1672,8 @@ class RequestHandler extends APIHandlerBase {
             if (isNaN(limit) || limit <= 0) {
                 limit = pageSizeOption;
             }
-            if (this.options.pageSize) {
-                limit = Math.min(this.options.pageSize, limit);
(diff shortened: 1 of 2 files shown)


ymc9 (commented):
Hi @sw34 ,

The original design intention is to prevent the client from "abusing" the API by limiting the max number of rows that can be returned. If a higher limit is needed, one can configure it when creating the API handler. Will setting the option solve your problem?

sw34:
Yes, that would work, but that would also mean that all queries would use that pageSize if a limit was not set at the query level, whereas my use case is I only have a few that require this higher limit.

Perhaps new/changed handler options, where one (likely existing) sets the default pageSize and one (a new one) sets an upper limit?

sw34:
Another comment, this only overrides the hardcoded default pageSize of 100.  If a pageSize is specified at handler creation that pageSize is honored as the upper limit.  

That said we can also overcome this with pulling back multiple pages before acting on the data, so if your not comfortable with the changes I am happy to close the pull request and issue.

ymc9:
> Yes, that would work, but that would also mean that all queries would use that pageSize if a limit was not set at the query level, whereas my use case is I only have a few that require this higher limit.
> 
> Perhaps new/changed handler options, where one (likely existing) sets the default pageSize and one (a new one) sets an upper limit?

Hi @sw34 , just wondering, is what you want equivalent to setting the `pageSize` option to `Infinity`? So that the query-level limit is always used.

sw34:
@ymc9 I will close this pull request and cancel the issue, I now see why you coded the way you did, I will set the option to a larger page size and set my query defaults to a lower amount, unless the larger one as needed.

 I also created issue #2644, can you update dependencies for the 2x branch, or give me some pointers how best to do that and I will submit pull requests.  I need to stay on v2 as we us mssqlserver for a datasource.

PR #2663, opened 2026-05-09
fix: detect policy plugin by stable id

### LINKED ISSUE #2662: [BUG] PolicyPlugin detection fails under Next.js/Turbopack bundling
Suggested github repo issue text for the zenstack repo

**Title**

PolicyPlugin detection via `constructor.name` breaks under Next.js/Turbopack production bundling

**Description**

In `@zenstackhq/orm@3.6.3`, policy mode detection relies on:

```ts
plugin.constructor.name === "PolicyPlugin"
```

This is brittle under production bundling/minification. Bundlers are allowed to rewrite class declarations/exports and do not guarantee that `constructor.name` is preserved.

In a Next.js 16/Turbopack production bundle, `PolicyPlugin` is emitted as an anonymous class assigned to an export object, similar to:

```js
exports.PolicyPlugin = class {
  // ...
}
```

For this pattern, JavaScript does not infer the class name from the property assignment, so:

```ts
new PolicyPlugin().constructor.name
```

can become an empty string, even though the plugin is still installed and active.

This creates an inconsistent state:

- ORM layer thinks `PolicyPlugin` is not enabled.
- `PolicyPlugin` is still present and intercepts Kysely queries.
- For `delete({ where })`, ORM generates `DELETE ... RETURNING *` instead of policy-safe `RETURNING id`.
- The active `PolicyPlugin` treats the full-returning mutation as needing read-back.
- Since the row has already been deleted, post-delete read-back returns `0 rows`.
- The mutation fails with:

```txt
result is not allowed to be read back
```

and the transaction is rolled back.

**Expected Behavior**

PolicyPlugin detection should not rely on `constructor.name`, because class names are not stable under bundling/minification.

When `PolicyPlugin` is active, ORM should consistently use policy-safe mutation behavior. In particular, `delete({ where })` should not end up as `DELETE ... RETURNING *` followed by post-delete read-back.

**Actual Behavior**

In Next.js/Turbopack production runtime:

```sql
DELETE ... RETURNING *
```

is generated, followed by a failed read-back and transaction rollback.

In a direct Node runtime:

```sql
DELETE ... RETURNING id
```

is generated, and the delete succeeds.

**Environment**

- `@zenstackhq/orm`: `3.6.3`
- `@zenstackhq/plugin-policy`: `3.6.3`
- Next.js: `16.2.4`
- Node.js: 24.11.0
- Package manager: pnpm
- Runtime: Next.js production server with Turbopack bundling
- Database: PostgreSQL

Albatrosso:
### Summary
Fixes #2662 

This updates ZenStack ORM policy plugin detection so it no longer relies on `constructor.name`, which can be changed or removed by production bundling/minification.

Policy mode is now detected through the stable runtime plugin id (`policy`). This keeps ORM mutation behavior consistent when `PolicyPlugin` is active, including policy-safe delete behavior that avoids `DELETE ... RETURNING *` followed by failed post-delete read-back.

This problem can break deletes in Next.js/Turbopack production bundles with:

`result is not allowed to be read back`

### Validation
pnpm --filter @zenstackhq/orm build
TEST_DB_PROVIDER=sqlite pnpm --dir tests/regression exec vitest run test/policy-plugin-detection.test.ts

PostgreSQL validation was attempted but blocked by local default credential auth:
`password authentication failed for user "postgres"`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved policy enforcement detection to work reliably across different code optimization levels.

* **Tests**
  * Added regression test suite for policy enforcement functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/zenstackhq/zenstack/pull/2663)

### DIFF
--- a/packages/orm/src/client/crud/operations/base.ts
+++ b/packages/orm/src/client/crud/operations/base.ts
@@ -221,7 +221,7 @@ export abstract class BaseOperationHandler<Schema extends SchemaDef> {
 
     // TODO: this is not clean, needs a better solution
     protected get hasPolicyEnabled() {
-        return this.options.plugins?.some((plugin) => plugin.constructor.name === 'PolicyPlugin');
+        return this.options.plugins?.some((plugin) => plugin.id === 'policy') ?? false;
     }
 
     protected requireModel(model: string) {
--- a/tests/regression/test/policy-plugin-detection.test.ts
+++ b/tests/regression/test/policy-plugin-detection.test.ts
@@ -1,0 +1,29 @@
+import { PolicyPlugin } from '@zenstackhq/plugin-policy';
+import { createTestClient } from '@zenstackhq/testtools';
+import { describe, expect, it } from 'vitest';
+
+describe('PolicyPlugin detection', () => {
+    it('uses plugin id when constructor names are bundled or minified', async () => {
+        const MinifiedPolicyPlugin = class a extends PolicyPlugin {};
+        const plugin = new MinifiedPolicyPlugin();
+
+        expect(plugin.id).toBe('policy');
+        expect(plugin.constructor.name).toBe('a');
+
+        const db = await createTestClient(
+            `
+model User {
+    id String @id
+    name String
+
+    @@allow('all', true)
+}
+            `,
+            { plugins: [plugin] },
+        );
+
+        await db.user.create({ data: { id: 'u1', name: 'User 1' } });
+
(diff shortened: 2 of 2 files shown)


ymc9 (approved):
LGTM. Thanks for the fix @Albatrosso .

I hate that the ORM core needs such a detection. Will think about how to get rid of it down the road.

PR #2675, opened 2026-05-19
fix(orm): use raw alias for join keys when PK/FK has field access policy

### LINKED ISSUE #2674: `@deny` on PK/FK fields breaks `include` when `external_id_mapping` is configured — related records silently return empty

When `external_id_mapping` is configured and a model's integer `id` (or a relation FK) has `@deny`, all `include`-based queries for relations using that field as a join key silently return empty arrays instead of the expected related records.

ZenStack implements field-level `@deny` by wrapping the source table in a SQL subquery that substitutes `CASE WHEN <deny_condition> THEN NULL ELSE field END AS field`. This NULL propagates into Prisma's include mechanism: Prisma reads the now-NULL PK from the parent row and constructs `WHERE fk_field IN (NULL)`, which matches zero rows.

```zmodel
model Post {
    // Integer PK hidden from API consumers; compound natural key used as JSON:API id via externalIdMapping
    id           Int    @id @default(autoincrement()) @deny('all', auth().role != 'SUPERADMIN')
    external_ref String
    source       String
    comments     Comment[]

    @@unique([source, external_ref])
}

model Comment {
    id      Int  @id
    post_id Int? @deny('all', auth().role != 'SUPERADMIN')
    post    Post? @relation(fields: [post_id], references: [id])
}
```

With `externalIdMapping: { Post: 'source_external_ref' }` and a `USER`-role auth context:

```ts
// Expected: post with populated comments array
// Actual:   { id: null, comments: [] }
const post = await db.post.findUnique({
    where: { source_external_ref: { source: 'blog', external_ref: 'my-post' } },
    include: { comments: true },
});
```

The same failure cascades to all HasMany relations whose FK has `@deny`, and to nested includes where the intermediate model's `id` or FK is denied.

The expected behavior is that `@deny` only redacts the field value from the serialized response and does not affect the SQL join keys used internally to resolve `include`. This is especially important when `external_id_mapping` is in use: the integer `id` is already absent from the public identifier (replaced by a compound natural key), so `@deny` on it should only suppress it from `attributes` — not NULL it in SQL.

**Screenshots**

N/A

**Environment:**

- ZenStack version: 3.6.1
- Database type: PostgreSQL
- Node.js version: 24.15.0
- Package manager: npm

**Additional context**

The failure cascades through the entire include tree:

- `Post.id @deny` → Prisma gets `post.id = null` → all HasMany `WHERE post_id IN (null)` return empty
- `Comment.post_id @deny` → even if `post.id` is fixed, the Comment subquery returns `post_id = null` → `WHERE post_id = {post.id}` still matches nothing
- Nested includes fail in the same way whenever an intermediate model's `id` or FK is denied

Only BelongsTo relations whose FK sits on the *parent* model (and is not itself denied) continue to work correctly, since Prisma fetches those using the parent's own attributes rather than the child's FK column.

Workaround: remove `@deny` from all PK/FK fields used as join keys and strip the values at the application layer if they must not be exposed.

lsmith77:
fix #2674

When a PK or FK field carries a `@deny` or `@allow` field-level policy, the
policy handler wraps it in `CASE WHEN … THEN NULL`. This caused `include` to
silently return empty arrays because join conditions evaluated as `fk = NULL`.

**Fix:** Two-part:

1. The policy handler emits a `__zs_raw_<field>` alias (the raw column value)
   alongside the potentially-nulled expression for any PK or FK field that has
   a field access policy.

2. Join condition builders (`buildJoinPairs`, SQLite's `buildRelationJoinFilter`,
   M2M path in the lateral-join dialect) use the raw alias selectively:
   - **HasMany** (`ownedByModel = false`): raw alias on *both* sides — the child's
     FK may be denied to hide which parent it belongs to, but the parent must
     still be able to fetch its children.
   - **BelongsTo** (`ownedByModel = true`): raw alias only on the relation's PK
     side; the FK side stays as a plain ref. Denying the FK is designed to hide
     the relation entirely (preserves pre-existing behavior verified by
     `field-level-policy.test.ts`).

Regression tests added in `tests/regression/test/issue-2674.test.ts`.

Note: `externalIdMapping` in the issue title is a REST API option unrelated to
the fix; the root cause is purely at the ORM layer.

Design notes:

fieldHasConditionalReadPolicy is static schema analysis — it can't know the live auth context. In practice this is fine: non-literal conditions like auth().role != 'ADMIN' compile to parameterised SQL regardless of the caller's role, so the raw alias is always projected. The one unsupported edge case is a policy whose condition is non-literal but collapses to trueNode at runtime for a specific auth value (e.g. @allow('read', auth() == null) with a null-auth caller); documented in the JSDoc.

BelongsTo FK intentionally keeps a plain ref — denying the FK nulls the join and hides the relation. This is pre-existing behaviour, verified by test 2.

The M2M test uses an explicit join model because $pushSchema does not create implicit _AToB tables. The implicit-M2M joinKeyRef branch is therefore not covered by a regression test; a follow-up would require usePrismaPush or extending $pushSchema.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cases where field-level read policies on primary/foreign keys could break joins or nested includes; related records now load while denied key fields resolve to null.

* **New Features**
  * Improved policy-aware join handling so key columns are exposed alongside guarded values, keeping joins and lateral subqueries reliable under conditional read policies.

* **Tests**
  * Expanded regression suite with additional scenarios covering denied and allowed policies, ownership branches, and include behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/zenstackhq/zenstack/pull/2675?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/orm/src/client/crud/dialects/lateral-join-dialect-base.ts
+++ b/packages/orm/src/client/crud/dialects/lateral-join-dialect-base.ts
@@ -9,6 +9,7 @@ import {
     getDelegateDescendantModels,
     getManyToManyRelation,
     isRelationField,
+    joinKeyRef,
     requireField,
     requireIdFields,
     requireModel,
@@ -139,14 +140,17 @@ export abstract class LateralJoinDialectBase<Schema extends SchemaDef> extends B
             const relationIds = requireIdFields(this.schema, relationModel);
             invariant(parentIds.length === 1, 'many-to-many relation must have exactly one id field');
             invariant(relationIds.length === 1, 'many-to-many relation must have exactly one id field');
+            // Use raw-alias refs so field access policies wrapping PK/FK in CASE WHEN NULL do not break the join.
+            const relationIdRef = joinKeyRef(this.schema, relationModel, relationModelAlias, relationIds[0]!);
+            const parentIdRef = joinKeyRef(this.schema, model, parentAlias, parentIds[0]!);
             query = query.where((eb) =>
                 eb(
-                    eb.ref(`${relationModelAlias}.${relationIds[0]}`),
+                    eb.ref(relationIdRef),
                     'in',
                     eb
                         .selectFrom(m2m.joinTable)
                         .select(`${m2m.joinTable}.${m2m.otherFkName}`)
-                        .whereRef(`${parentAlias}.${parentIds[0]}`, '=', `${m2m.joinTable}.${m2m.parentFkName}`),
+                        .whereRef(parentIdRef, '=', `${m2m.joinTable}.${m2m.parentFkName}`),
                 ),
             );
         } else {
--- a/packages/orm/src/client/crud/dialects/sqlite.ts
+++ b/packages/orm/src/client/crud/dialects/sqlite.ts
@@ -21,6 +21,7 @@ import {
     getDelegateDescendantModels,
     getManyToManyRelation,
     getRelationForeignKeyFieldPairs,
+    joinKeyRef,
     requireField,
     requireIdFields,
     requireModel,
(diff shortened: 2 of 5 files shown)


lsmith77:
@coderabbitai full review

lsmith77:
@coderabbitai review

ymc9:
Hi @lsmith77 , thanks for working on this. I've been thinking about this and tend to feel the current behavior is by design. The mental model of "field-level" policy is: if a field is not readable, it appears absent. This is consistent with row-level: if a row is not readable, it appears non-existent. In this sense, a join "should" fail because the keys aren't there ...

Also, from architecture point view, policy is implemented as a plugin, so a tighter coupling between it and the ORM is something we tried to avoid as much as possible.

I think it's quite rare to mark id fields as non-readable. Any specific reason for that?

lsmith77:
The use case relates to the `external_id_mapping` feature. I want to hide surrogate keys from the API users and instead want to just expose natural keys.

ymc9:
> The use case relates to the `external_id_mapping` feature. I want to hide surrogate keys from the API users and instead want to just expose natural keys.

Got it. Have you tried if the `@omit` or the `omit` option solves this problem?

lsmith77:
`@omit` seems to work. I wanted to expose the surrogate keys to internal API calls. But I guess I can work around this.

ymc9:
Closing per previous discussions.

PR #2703, opened 2026-06-09
fix: `isEmpty` function when using `postgresql` dialect

sanny-io:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostgreSQL array length operations to explicitly specify dimension parameters for correct SQL generation.

* **Tests**
  * Added comprehensive end-to-end test for PostgreSQL policy behavior with isEmpty function in deny rules.


sanny-io:
Context: `isEmpty` straight up did not work on postgres. `array_length` **requires** two parameters, the second is the dimension of the array. Previously it would result in an error saying `function array_length(text[]) does not exist`

For `in` expressions, see below.

https://discord.com/channels/1035538056146595961/1513844077383258122/1513844077383258122

sanny-io:
I believe the issue may only happen when using access policies, so I have relocated the test.

ymc9 (approved):

PR #2714, opened 2026-06-16
fix: pageParam in seInternalInfiniteQuery for Vue

### LINKED ISSUE #2713: [Vue] useInfiniteFindMany ignores pageParam — infinite pagination never advances (diverges from React adapter)
**Describe the bug**

In the Vue TanStack Query runtime, `useInfiniteFindMany` never paginates: every page after the first re-fetches the initial args. The pageParam returned by `getNextPageParam` is ignored because the default queryFn doesn't read it. The React runtime handles it correctly `(pageParam ?? args)`, so this is Vue-specific.

**Affected versions**
Confirmed in 3.7.1, 3.7.2 (latest) and 3.8.0-beta.3 (canary) — the compiled queryFn is identical across them.

**Root cause**
`useInternalInfiniteQuery` in the Vue runtime builds the default queryFn ignoring `pageParam`:

**Vue (dist/vue.js):**
```js
queryFn: ({ signal }) => {
  const reqUrl = makeUrl(endpoint, model, operation, argsValue); // argsValue, not pageParam
  return fetcher(reqUrl, { signal }, fetch);
},
initialPageParam: toValue(argsValue),
```

**React (dist/react.js) — correct:**
```js
queryFn: ({ pageParam, signal }) =>
  fetcher(makeUrl(endpoint, model, operation, pageParam ?? args), { signal }, fetch),
initialPageParam: args,
```

The Vue version destructures only `{ signal }` and uses the closed-over `argsValue`, so the `pageParam` produced by `getNextPageParam` never reaches `makeUrl`.

Reproduction
```js
const { fetchNextPage } = client.post.useInfiniteFindMany(
  { take: 10, orderBy: { createdAt: 'desc' } },
  {
    getNextPageParam: (lastPage, pages) =>
      lastPage.length < 10
        ? undefined
        : { take: 10, skip: pages.length * 10, orderBy: { createdAt: 'desc' } },
  },
);
// call fetchNextPage()
```

`getNextPageParam` correctly returns `{ skip: 10, ... }`, but the outgoing request still sends `skip: 0`, so page 2 returns the same rows as page 1 (duplicates) and pagination never advances.

**Expected behavior**
The Vue runtime should use `pageParam`, like React:
```js
queryFn: ({ pageParam, signal }) =>
  fetcher(makeUrl(endpoint, model, operation, pageParam ?? toValue(args)), { signal }, fetch),
```

**Workaround**
Override `queryFn` in the options passed to `useInfiniteFindMany` (options are spread after the defaults, so it wins), rebuilding the request with makeUrl/fetcher and `pageParam ?? args`.

**Environment**
@zenstackhq/tanstack-query 3.7.2 (also 3.7.1, 3.8.0-beta.3)
Vue runtime (/vue), @tanstack/vue-query v5, Nuxt 4

caiotarifa:
Closes #2713

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed infinite-query pagination to correctly use the current page parameter when constructing requests, instead of reusing the initial request parameters for subsequent pages.

### DIFF
--- a/packages/clients/tanstack-query/src/vue.ts
+++ b/packages/clients/tanstack-query/src/vue.ts
@@ -651,8 +651,8 @@ export function useInternalInfiniteQuery<TQueryFnData, TData, TPageParam = unkno
         const argsValue = toValue(args);
         return {
             queryKey: queryKey.value,
-            queryFn: ({ signal }: any) => {
-                const reqUrl = makeUrl(endpoint, model, operation, argsValue);
+            queryFn: ({ pageParam, signal }: any) => {
+                const reqUrl = makeUrl(endpoint, model, operation, pageParam ?? argsValue);
                 return fetcher<TQueryFnData>(reqUrl, { signal }, fetch);
             },
             initialPageParam: toValue(argsValue) as TPageParam,


ymc9 (approved):
Thanks @caiotarifa . LGTM!

Issue #2531, opened 2026-03-30
Many-to-many update check failing with connect


f8k8:
**Description and expected behavior**
When trying to create and connect an object, in the same operation, which has the following policy:
``@@allow('read,update', auth() != null && parents?[id == auth().id])``

The create fails with a model not updatable error:
`ORMError: many-to-many relation participant model "Child" not updatable`

The read back should succeed though, as the parent would be connected as part of the operation:
```
const child = await authedClient.child.create({
    data: {
      name: "Child",
      parents: {
        connect: { id: userId },
      },
    },
  });
```

**Environment:**

-   ZenStack version: 3.5.2
-   Database type: Postgres / SQLite
-   Node.js version: 24.11.1
-   Package manager: npm

**Additional context**
There's a reproduction repo here: https://github.com/f8k8/ZenstackConnectPermissionIssue
Running `npm install` will install the dependencies and seed the database, then run `npm run start` to see the error.

ludovicmotte:
Hi, I think I also reproduced this issue in ZenStack 3.9.2.
I see a commit that seems to fix it but is in part of a release ?

ludovicmotte:
I found a workaround: https://github.com/zenstackhq/zenstack/issues/2382#issuecomment-5730788305

Issue #2382, opened 2026-02-13
[Feature Request] Field-level access overrides for implicit relations


Jamie-Fairweather:
**Is your feature request related to a problem? Please describe.**
In ZenStack v3, implicit relations (e.g. many-to-many) cannot be updated unless the related model allows update. This makes it impossible to safely support common scenarios where a model is intentionally read-only, but users should still be able to create or remove links to it.

Example:
`Club`: full access for owner
`Activity`: publicly readable, no create/update/delete
`Club` ↔ `Activity`: many-to-many

A club owner should be able to link/unlink existing Activities to their Club, but because Activity does not allow update, relation updates are blocked. The only workaround is to relax model-level permissions (unsafe) or introduce an explicit join table.

**Describe the solution you'd like**
Support fine-grained authorization for implicit relation updates, without requiring update access on the related model.

This could be achieved by one or more of the following:
- Explicit policies for implicit relation updates (connect / disconnect)
- Field-level overrides of model access policies (e.g. allow updates on a relation field while the model itself remains read-only)
- First-class separation of relation mutation permissions from general update

This would enable safe linking of read-only reference data (tags, activities, categories, etc.) to user-owned models.

**Describe alternatives you've considered**
- Defining an explicit join table and configuring access policies on it: Works, but adds unnecessary schema complexity for a very common use case
- Granting update access on the related model: Not acceptable, as it allows unintended mutations of shared data

**Additional context**
This appears to be a missing feature in v3. Field-level overrides or relation-level access control existed in some form in v2 but are not yet available in v3. This limitation was confirmed by @ymc9, who indicated that explicit join tables are currently the only supported solution.

ludovicmotte:
Hi @Jamie-Fairweather 
I had the same issue.
I just found a workaround:
 - relax model-level permissions on Activity    e.g. @@allow('update', true)
 - add on each field an update deny                e.g. activityName  String  @deny('update', auth().role != ADMIN)

Hope it helps.

But, yes, I hope Zenstack will provide a cleaner solution.

ludovicmotte:
Following up on my workaround — I analyzed its limits and it's narrower than it looks:

The workaround breaks when two relations share the same endpoint pair with the same update policies. A link's effective permission is the intersection of the UPDATE policies of the two endpoint rows, evaluated per row without relation context. So:

✅ Club.activities (owner-only) vs Event.activities (admin-only) → works, different writable endpoints
[...]

Issue #856, opened 2023-11-26
[Feature Request] Consider using field-level access policies to control permissions for implicit many-to-many relation


ymc9:
Discord link: https://discord.com/channels/1035538056146595961/1178398784980525207

PR #2730, opened 2026-06-22
perf(orm): memoize implicit m2m join-table and model lookups (#2715)

### LINKED ISSUE #2715: PolicyPlugin: resolveManyToManyJoinTable does an uncached full-schema scan on every query (~29% CPU self-time)
### Summary

In `@zenstackhq/plugin-policy`, `PolicyHandler.resolveManyToManyJoinTable(tableName)` performs an **O(models × fields) linear scan of the entire schema on every invocation, with no memoization**. It is invoked **once per table per (sub)query** while building policy filters (and again on the mutation/validation path). The result is a pure function of `(schema, tableName)`, and the schema is immutable for the client's lifetime, so it is trivially cacheable.

In CPU profiling of a read-heavy endpoint, `getModelPolicyFilterForManyToManyJoinTable` (which calls `resolveManyToManyJoinTable`) is **28.9% of total CPU self-time**, and the m2m-resolution cluster is **~40%**. The database itself is **0.28%** of CPU — the cost is entirely in this JS resolution. Because it's single-threaded CPU work it serializes under concurrency, which on a 1-vCPU container turns into large p50 latency increases.

### Version

`@zenstackhq/plugin-policy` and `@zenstackhq/orm` **3.6.4**; `kysely` 0.28.17; Postgres 16. Latest stable is 3.7.2; from the changelog there's no perf change on this path, so I believe it's still present on `main` — happy to confirm against a specific ref.

### Root cause

`resolveManyToManyJoinTable` scans every field of every model to find the join table matching `tableName`, recomputed on each call (`src/policy-handler.ts`, paraphrased from 3.6.4):

```ts
resolveManyToManyJoinTable(tableName) {
  for (const model of Object.values(this.client.$schema.models))
    for (const field of Object.values(model.fields)) {
      const m2m = QueryUtils.getManyToManyRelation(this.client.$schema, model.name, field.name);
      if (m2m?.joinTable === tableName) {
        // ...build + return the descriptor
      }
    }
  // returns undefined for every non-m2m table — *after* scanning the whole schema
}
```

The inner `getManyToManyRelation` (in `@zenstackhq/orm`) is itself non-trivial per call (`requireField` ×2, `requireIdFields` ×2, attribute inspection).

It's on the hot path because `buildPolicyFilter` calls it first, for **every table referenced in a query**:

```ts
buildPolicyFilter(model, alias, operation) {
  const m2mFilter = this.getModelPolicyFilterForManyToManyJoinTable(model, alias, operation); // -> resolveManyToManyJoinTable
  if (m2mFilter) return m2mFilter;
  // ...normal policy compilation
}
```

So even a regular (non-m2m) model pays a full schema scan that returns nothing. It's also reached via `isManyToManyJoinTable` and `tryRejectNonexistentModel` → `tryRejectNonexistingTables`. Net cost per query ≈ `(#tables in query) × (#models × #fields)` `getManyToManyRelation` calls. For a deep read with several includes that traverse implicit m2m relations, this dominates the request.

### Evidence

CPU self-time, deep read query, 56,750 samples (Node 20, `node:inspector`):

| % self-time | function | package |
|---:|---|---|
| **28.9%** | `getModelPolicyFilterForManyToManyJoinTable` | plugin-policy |
| 16.1% | `getModel` | orm |
| 9.0% | `resolveManyToManyJoinTable` | plugin-policy |
| 1.7% | `getManyToManyRelation` | orm |
| **0.28%** | (actual SQL / `pg`) | pg |

Buckets: plugin-policy 39.3%, orm 26.5%, kysely 4.4%, pg 0.28%, idle/GC ~29%. The policy transform adds ~+1ms to a PK lookup but **~+20ms to a deep multi-include read** (warm/reused client). Under concurrency it scales ~linearly (single-thread CPU serialization).

Full CPU profile (sanitized; opens in Chrome DevTools → Performance, or speedscope): https://gist.github.com/abetss/58446f08bf2210876603d9838731f892

### Proposed fix

Memoize `resolveManyToManyJoinTable` by `tableName`. The schema is fixed for the client's lifetime, so an instance-level cache that also caches the negative result removes the repeated scans:

```ts
#m2mCache = new Map(); // tableName -> descriptor | undefined

resolveManyToManyJoinTable(tableName) {
  if (this.#m2mCache.has(tableName)) return this.#m2mCache.get(tableName);
  const result = /* ...existing scan... */;
  this.#m2mCache.set(tableName, result);
  return result;
}
```

Or stronger: precompute a `Map<joinTableName, descriptor>` once when the schema is bound (single pass over the schema), making `resolveManyToManyJoinTable` an O(1) lookup. `getModel` (16%) looks similarly memoizable.

### Reproduction

The profile above is from a deep read with nested implicit-m2m includes. I can put together a minimal standalone repro (a small schema with one implicit m2m relation + a `@@allow` read policy, profiling a nested read) if that would help — happy to do so.

### Environment

Node 20 (profiled) and Bun 1.3 (our production runtime); cause ranking is identical on both engines, magnitude ~4× larger on Bun/JavaScriptCore.

abetss:
Fixes #2715.

## Summary

`PolicyHandler.resolveManyToManyJoinTable(tableName)` performed an `O(models × fields)` scan of the **entire schema on every call** — and it's called once per table per (sub)query while building policy filters (plus on the mutation/validation path). For a regular non-m2m table the scan runs to completion and returns `undefined`, so every query pays a full schema scan per referenced table. On a CPU profile of a deep, policy-filtered read this (`getModelPolicyFilterForManyToManyJoinTable`) was ~29% self-time, with `getModel` (16%) close behind; the database itself was 0.3%.

The resolution is a pure function of `(schema, tableName)`, and the schema is immutable for a client's lifetime. This PR memoizes it.

## What changed

Rather than caching inside the policy plugin, the structural lookups are memoized once in `@zenstackhq/orm`'s `QueryUtils`, keyed by the immutable schema, so **every consumer** (ORM query building + all plugins) shares them:

- `query-utils.ts`: a `WeakMap<SchemaDef, …>` index memoizing `getModel` and `getManyToManyRelation`, plus a new O(1) `getManyToManyJoinTable(schema, tableName)` that builds a `Map<joinTable, endpoints>` in a single pass on first use.
- `policy-handler.ts`: `resolveManyToManyJoinTable` is now a thin O(1) lookup into that index (the per-call scan is gone). The descriptor it returns is unchanged.

Typed throughout (no `any`/`as`).

## Why it's access-control-safe

`resolveManyToManyJoinTable` reads only `schema.models` + the table name and returns a **structural** descriptor (`{firstModel, firstField, firstIdField, …}`). It takes no `$auth`/per-request state and mutates nothing. The **per-user policy filter is still built fresh, per call, downstream** in `getModelPolicyFilterForManyToManyJoinTable` from that descriptor + `auth()`. So the cache cannot change any access-control outcome — it only skips recomputing a deterministic lookup.

What this deliberately does **not** cache (to stay forward-compatible):
- the compiled policy filter node — left per-call, so the newly-created-side skip from #2531 and future field-level relation policies (#856, #2382) keep working;
- anything `auth()`/value-dependent.

The `WeakMap` is keyed by the schema object, so the cache dies with the schema (no leak, no cross-schema collision), and the negative (`undefined`) result is cached too — which matters because non-m2m tables are the dominant call pattern.

## Benchmarks

Deep, policy-filtered read on a ~114-model schema (Bun 1.3.2 / JavaScriptCore; medians). Bun is the worst case for this JS-bound path; Node/V8 shows the same ranking at ~¼ the magnitude.

| scenario | before | after | Δ |
|---|--:|--:|--:|
| warm deep read (reused client) | 92.7 ms | 10.5 ms | −89% |
| cold deep read (chain rebuilt per call) | 136.8 ms | 31.6 ms | −77% |
| warm PK lookup | 6.9 ms | 0.65 ms | −91% |
| cold PK lookup | 42.7 ms | 14.9 ms | −65% |
| concurrency c=8 (wall) | 698 ms | 46.5 ms | −93% |
| concurrency c=16 (wall) | 1380 ms | 87 ms | −94% |

The PK-lookup wins (which touch no m2m code at all) come from memoizing `getModel` — that's why the index lives in `QueryUtils` rather than the policy plugin.

A sanitized CPU profile (`.cpuprofile`) backing the original 29%/16% attribution is available — happy to attach or gist it.

## Tests

- New `tests/regression/test/issue-2715.test.ts` guards the memoization (resolution is correct for two-model, self-relation, explicit `@relation` name, and multiple relations; repeat resolution returns the same cached descriptor; negative results are cached). Verified it fails if the memoization is removed.
- Existing m2m + policy e2e suites (`relation/many-to-many`, `policy/migrated/relation-many-to-many-filter`, `connect-disconnect`, `self-relation`) pass unchanged — the behavior-preservation gate.

## Questions for maintainers

1. **Where should the index live?** I put it as a module-level `WeakMap<SchemaDef, …>` inside `QueryUtils`. If you'd prefer no module-level state, it could hang off the long-lived client/schema object instead — happy to move it.
2. **Scope:** I included `getModel`/`getManyToManyRelation` memoization alongside the join-table fix because the profile and the PK-lookup numbers show `getModel` is the larger lever. I can split it out if you'd rather keep this PR to `resolveManyToManyJoinTable` only.

## Summary by CodeRabbit

* **Performance Improvements**
  * Model resolution queries now execute faster
  * Many-to-many relationship handling optimized for improved performance

* **New Features**
  * Enhanced support for implicit many-to-many join table discovery and resolution

* **Tests**
  * Added regression test for many-to-many join table functionality

### DIFF
--- a/packages/orm/src/client/query-utils.ts
+++ b/packages/orm/src/client/query-utils.ts
@@ -20,8 +20,38 @@ export function hasModel(schema: SchemaDef, model: string) {
         .includes(model.toLowerCase());
 }
 
+/**
+ * Structural lookups derived purely from the (immutable) schema. The schema does not
+ * change for a client's lifetime, so these results are memoized per-schema and shared
+ * across every consumer (orm query building + plugins). Keyed by the schema object so
+ * the cache dies with the schema and never leaks across schemas. See issue #2715.
+ */
+interface SchemaLookupCache {
+    model: Map<string, ModelDef | undefined>;
+    m2mRelation: Map<string, ReturnType<typeof computeManyToManyRelation>>;
+    m2mJoinTable?: Map<string, ManyToManyJoinTableEndpoints | undefined>;
+}
+
+const schemaLookupCache = new WeakMap<SchemaDef, SchemaLookupCache>();
+
+function getSchemaLookupCache(schema: SchemaDef): SchemaLookupCache {
+    let cache = schemaLookupCache.get(schema);
+    if (!cache) {
+        cache = { model: new Map(), m2mRelation: new Map() };
+        schemaLookupCache.set(schema, cache);
+    }
+    return cache;
+}
+
 export function getModel(schema: SchemaDef, model: string) {
-    return Object.values(schema.models).find((m) => m.name.toLowerCase() === model.toLowerCase());
+    const cache = getSchemaLookupCache(schema);
+    const key = model.toLowerCase();
+    if (cache.model.has(key)) {
+        return cache.model.get(key);
+    }
+    const result = Object.values(schema.models).find((m) => m.name.toLowerCase() === key);
+    cache.model.set(key, result);
+    return result;
 }
 
 export function getTypeDef(schema: SchemaDef, type: string) {
@@ -260,6 +290,58 @@ export function makeDefaultOrderBy(schema: SchemaDef, model: string) {
 }
 
 export function getManyToManyRelation(schema: SchemaDef, model: string, field: string) {
+    const cache = getSchemaLookupCache(schema);
+    const key = `${model} ${field}`;
+    if (cache.m2mRelation.has(key)) {
+        return cache.m2mRelation.get(key);
+    }
+    const result = computeManyToManyRelation(schema, model, field);
+    cache.m2mRelation.set(key, result);
+    return result;
+}
+
+/**
+ * Endpoints of an implicit many-to-many relation, identified by its join table name.
+ */
+export interface ManyToManyJoinTableEndpoints {
+    model: string;
+    field: string;
+    otherModel: string;
+    otherField: string;
+}
+
+/**
+ * Resolve the relation endpoints for an implicit many-to-many join table by its table
+ * name, or `undefined` if the table is not an implicit m2m join table. The join-table
+ * index is built once per schema (single pass) and reused, making this an O(1) lookup.
+ * See issue #2715.
+ */
+export function getManyToManyJoinTable(
+    schema: SchemaDef,
+    joinTableName: string,
+): ManyToManyJoinTableEndpoints | undefined {
+    const cache = getSchemaLookupCache(schema);
+    if (!cache.m2mJoinTable) {
+        const map = new Map<string, ManyToManyJoinTableEndpoints>();
+        for (const model of Object.values(schema.models)) {
+            for (const field of Object.values(model.fields)) {
+                const m2m = getManyToManyRelation(schema, model.name, field.name);
+                if (m2m?.joinTable && !map.has(m2m.joinTable)) {
+                    map.set(m2m.joinTable, {
+                        model: model.name,
+                        field: field.name,
+                        otherModel: m2m.otherModel,
+                        otherField: m2m.otherField,
+                    });
+                }
+            }
+        }
+        cache.m2mJoinTable = map;
+    }
+    return cache.m2mJoinTable.get(joinTableName);
+}
+
+function computeManyToManyRelation(schema: SchemaDef, model: string, field: string) {
     const fieldDef = requireField(schema, model, field);
     if (!fieldDef.array || !fieldDef.relation?.opposite) {
         return undefined;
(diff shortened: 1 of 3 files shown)


ymc9 (approved):
Hi @abetss , many thanks for working on this! The change looks great to me and the numbers are exciting. I appreciate that you attached detailed profiling results.

For the proper place to hold the cache, the `@zenstackhq/schema` package has a `SchemaAccessor` wrapper that's supposed to both provide quick access utilities and cache as necessary. However, it hasn't been completed yet and a refactor to use it has not been done. I think it's fine to keep your cache in query-utils for now and it can be consolidated later.

PR #2734, opened 2026-06-26
fix(policy): resolve this.relation.field against @@allow model, unnest array columns in subqueries

### LINKED ISSUE #2733: Bug: this.relationField resolves against collection model inside auth().collection?[...] predicates
## Description

Inside an access policy collection predicate on `auth()`, the `this` keyword resolves **scalar fields** against the `@@allow` model correctly, but resolves **relation fields** against the collection's model instead — contradicting the docs and breaking scope-tree RBAC patterns.

## Reproduction

Minimal ZModel:

```zmodel
model ActorContext {
  id           String @id
  capabilityGrants DerivedCapabilityGrant[]
  @@auth
}

model DerivedCapabilityGrant {
  id             String @id
  actorContextId String
  capabilityKey  String
  grantScopeId   String
  grantScope     AuthScope @relation(fields: [grantScopeId], references: [id])
}

model AuthScope {
  id        String    @id
  path      String[]
  documents DocumentBase[]
}

model DocumentBase {
  documentId  String    @id
  tenantId    String
  authScopeId String
  authScope   AuthScope @relation(fields: [authScopeId], references: [id])
  
  @@allow('read',
    auth() != null &&
    auth().tenantId == tenantId &&
    auth().capabilityGrants?[
      capabilityKey == 'read' &&
      (grantScopeId == this.authScopeId || grantScopeId in this.authScope.path)
    ]
  )
}
```

The ZModel parser accepts `grantScopeId in this.authScope.path` at generation time. But at runtime, the policy engine crashes:

```
Error: Field "authScope" not found in model "DerivedCapabilityGrant"
```

Stack trace:

```
ExpressionTransformer.transformRelationAccess (expression-transformer.ts:837)
ExpressionTransformer._member (expression-transformer.ts:705)
```

## Expected Behavior

According to the [docs](https://zenstack.dev/docs/reference/zmodel/expression), "the keyword `this` can be used inside the CONDITION to refer to fields belonging to the model where the rule is defined (the outer model)."

So `this.authScope.path` inside `auth().capabilityGrants?[...]` should resolve:
1. `this` → `DocumentBase` (the model with `@@allow`)
2. `this.authScope` → `DocumentBase.authScope` relation
3. `this.authScope.path` → `AuthScope.path`

## Actual Behavior

The engine resolves `authScope` against `DerivedCapabilityGrant` (the collection's model), which doesn't have that field.

## Workaround

1. `check()` delegation to AuthScope's `@@allow` (which doesn't suffer from this issue since `this` resolves correctly in top-level or `check`-based predicates)
2. Denormalizing the relation field into a scalar array (e.g., `scopingPath String[]` on DocumentBase) and checking `grantScopeId in this.scopingPath`

## Environment

- ZenStack version: 3.8.0
- Provider: PostgreSQL (PGLite but I doubt that caused this bug)
- ZenStack CLI: zen generate
- ORM client

## Notes

- **Scalar fields** on `this` work correctly (e.g., `grantScopeId == this.authScopeId` passes)
- **Relation fields** on `this` behave identically with and without binding variables (`[g, ...]`)
- The failure is at runtime (expression-transformer.ts:837), not at parse time — the ZModel parser accepts the syntax

hechang27-sprt:
Fixes #2733.

## Problem

Two bugs in the policy plugin expression transformer:

1. **`this.relation.field` resolves against the wrong model** inside `auth().collection?[...]` predicates. The expression transformer uses `context.thisType` to resolve the field definition but passes `restContext` (with `modelOrType` still pointing to the collection model) to `transformRelationAccess`. This causes `Field "X" not found in model "<collection>"` crashes.

2. **`x in this.relation.arrayField` generates broken SQL on PostgreSQL**. The expression compiles to `= ANY((SELECT arrayCol FROM ...))` where the subquery returns array-typed rows (`integer[]`). PostgreSQL `ANY(subquery)` requires scalar rows, producing `operator does not exist: integer = integer[]`.

### Fix 1: Resolve `this.relation.field` against the `@@allow` model

Pass `modelOrType: context.thisType` and `alias: context.thisAlias` to `transformRelationAccess` when the receiver is `this`, so relation fields and table aliases resolve against the model bearing the policy. Also fixed `getFieldDefFromFieldRef` to resolve multi-segment `this.relation.field` chains (needed for the array-field detection in Fix 2).

### Fix 2: Cross-database `scalar IN relation.arrayField`

Single entry point `buildArrayInExists` in `_binary`, with provider-specific handling:

| Provider | Approach | Generated SQL |
|----------|----------|--------------|
| PostgreSQL | `unnest()` in subquery column + `= ANY(subquery)` | `scalar = ANY(SELECT unnest("col") FROM "t" WHERE <join>)` |
| SQLite | Nested EXISTS with `json_each` | `EXISTS (SELECT 1 FROM "t" WHERE <join> AND EXISTS (SELECT 1 FROM json_each("t"."col") WHERE "value" = scalar))` |
| MySQL | Nested EXISTS with `JSON_TABLE` | `EXISTS (SELECT 1 FROM "t" WHERE <join> AND EXISTS (SELECT 1 FROM JSON_TABLE(...) AS jt WHERE jt."value" = scalar))` |

The SQLite/MySQL `EXISTS` rebuild preserves the original subquery `FROM` clause, so many-to-many joins are kept intact.

## Verification

Two new PG-specific tests added to `auth-access.test.ts`:

| Test | Without fix | With fix |
|------|------------|----------|
| `this.author.level` in `auth().permissions?[p, ...]` | Field "author" not found in model "Permission" | ✓ passes |
| `this.id in this.group.visibleDocIds` | operator does not exist: integer = integer[] | ✓ passes |

All existing policy tests continue to pass on PostgreSQL, SQLite, and MySQL (no regressions).

## Summary by CodeRabbit

* **New Features**
  * Expanded policy `in` handling for correlated subqueries that select array-typed fields, compiling to array membership checks across supported SQL dialects.
  * Enhanced evaluation of `this`-rooted multi-segment relation/collection expressions in authorization conditions.

* **Bug Fixes**
  * Fixed related-field comparisons inside policy rules, including nested relation hops.
  * Corrected collection predicate resolution to ensure proper SQL-based scoping for `this`-rooted expressions during auth evaluation.

* **Tests**
  * Added end-to-end authorization tests covering array membership, related-field gating, and correct auth scoping for `this`-rooted predicates.

### DIFF
--- a/packages/plugins/policy/src/expression-transformer.ts
+++ b/packages/plugins/policy/src/expression-transformer.ts
@@ -702,7 +702,10 @@ export class ExpressionTransformer<Schema extends SchemaDef> {
             } else {
                 // transform the first segment into a relation access, then continue with the rest of the members
                 const firstMemberFieldDef = QueryUtils.requireField(this.schema, context.thisType, expr.members[0]!);
-                receiver = this.transformRelationAccess(expr.members[0]!, firstMemberFieldDef.type, restContext);
+                receiver = this.transformRelationAccess(expr.members[0]!, firstMemberFieldDef.type, {
+                    ...restContext,
+                    modelOrType: context.thisType,
+                });
                 members = expr.members.slice(1);
                 // startType should be the type of the relation access
                 startType = firstMemberFieldDef.type;
@@ -756,7 +759,7 @@ export class ExpressionTransformer<Schema extends SchemaDef> {
             currType = fieldDef.type;
         }
 
-        let currNode: SelectQueryNode | ColumnNode | ReferenceNode | undefined = undefined;
+        let currNode: SelectQueryNode | ColumnNode | ReferenceNode | FunctionNode | undefined = undefined;
 
         for (let i = members.length - 1; i >= 0; i--) {
             const member = members[i]!;
@@ -788,7 +791,9 @@ export class ExpressionTransformer<Schema extends SchemaDef> {
                 invariant(i === members.length - 1, 'plain field access must be the last segment');
                 invariant(!currNode, 'plain field access must be the last segment');
 
-                currNode = ColumnNode.create(member);
+                currNode = fieldDef.array && this.schema.provider.type === 'postgresql'
+                    ? FunctionNode.create('unnest', [ColumnNode.create(member)])
+                    : ColumnNode.create(member);
             }
         }
 
--- a/packages/plugins/policy/src/expression-transformer.ts
+++ b/packages/plugins/policy/src/expression-transformer.ts
@@ -1066,8 +1066,7 @@ export class ExpressionTransformer<Schema extends SchemaDef> {
                 // available for casting in buildComparison.
                 return walkRelationChain(model, [expr.receiver.field, ...expr.members]);
             }
-        } else {
-            return undefined;
         }
+        return undefined;
     }
 }
--- a/packages/plugins/policy/src/expression-transformer.ts
+++ b/packages/plugins/policy/src/expression-transformer.ts
@@ -1138,18 +1138,17 @@ export class ExpressionTransformer<Schema extends SchemaDef> {
                 .toOperationNode();
         }
 
-        // combine array check with original WHERE
         const combinedWhere = subquery.where
             ? conjunction(this.dialect, [subquery.where.where, arrayCheck])
             : arrayCheck;
 
-        // preserve original FROM to handle joins (m2m etc.)
         return UnaryOperationNode.create(
             {
-                kind: 'SelectQueryNode',
-                from: subquery.from,
+                ...(subquery as SelectQueryNode),
                 where: WhereNode.create(combinedWhere),
-                selections: [SelectionNode.create(AliasNode.create(ValueNode.createImmediate(1), IdentifierNode.create('_')))],
+                selections: [
+                    SelectionNode.create(AliasNode.create(ValueNode.createImmediate(1), IdentifierNode.create('_'))),
+                ],
             } as SelectQueryNode,
             OperatorNode.create('exists'),
         );
--- a/packages/plugins/policy/src/expression-transformer.ts
+++ b/packages/plugins/policy/src/expression-transformer.ts
@@ -40,9 +40,7 @@ import {
     ReferenceNode,
     SelectionNode,
     SelectQueryNode,
-    sql,
     TableNode,
-    UnaryOperationNode,
     ValueListNode,
     ValueNode,
     WhereNode,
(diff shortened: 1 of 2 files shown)


hechang27-sprt:
Pushed commit 29a60bb to branch fix/this-authscope-relation-resolution.

What changed:
- preserve this-rooted collection predicates on the @@allow model when they appear inside auth-binding predicates
- normalize relation comparisons in those value-backed paths
- add a focused regression in tests/e2e/orm/policy/auth-access.test.ts (Fix #3)

Verification:
- A/B tested the new ZenStack regression directly in tests/e2e/orm/policy/auth-access.test.ts
- with patch: PASS
- stashing only packages/plugins/policy/src/expression-transformer.ts, rebuilding @zenstackhq/plugin-policy, same test: FAIL with "Field \"authScope\" not found in model \"RoleAssignment\""
- restoring the patch and rebuilding: PASS

So the new fix is necessary, and those two broader-suite failures are not reproduced by simply removing this patch alone.

hechang27-sprt:
Addressed the latest CodeRabbit finding in commit 9992d72.

Change:
- `buildArrayInExists` no longer reconstructs a fresh query from `froms[0]`
- it now reuses the incoming `SelectQueryNode` shape and only replaces:
  - `where` with the original predicate conjoined with the array containment check
  - `selections` with a constant projection for `EXISTS`

This preserves joins/correlation from relation access subqueries, including the many-to-many path CodeRabbit called out, and also avoids assuming `subquery.where` already exists.

Verification:
- `pnpm -C devenv/zenstack-fork --filter @zenstackhq/plugin-policy build`
- `TEST_PG_HOST=localhost TEST_PG_PORT=5432 TEST_PG_USER=postgres TEST_PG_PASSWORD=123456 pnpm -C devenv/zenstack-fork/tests/e2e exec vitest run orm/policy/auth-access.test.ts -t "Fix #2" --config vitest.config.ts`
- `TEST_PG_HOST=localhost TEST_PG_PORT=5432 TEST_PG_USER=postgres TEST_PG_PASSWORD=123456 pnpm -C devenv/zenstack-fork/tests/e2e exec vitest run orm/policy/auth-access.test.ts -t "Fix #3" --config vitest.config.ts`

I tried to add a dedicated m2m regression for `this.m2mRelation.arrayField`, but the ZModel resolver rejects that direct member chain today (`Could not resolve reference to MemberAccessTarget`). So the code fix is in, but there is not yet a focused schema-level regression for that exact syntax shape.

ymc9 (commented):
Hi @hechang27-sprt , many thanks for the PR!

One thing I'd like to address is that with the change, the policy plugin now contains logic that checks for database type - which the original design tried to avoid (so the plugin is db-type agnostic). The dialect abstraction already contains array operation abstractions that can be used here.

Do you mind if I make changes directly off your branch for the comments above, together with some simplification on the `this` and binding resolution issues?

hechang27-sprt:
@ymc9 of course not as it's mostly ai generated code, feel free to do so.

ymc9:
@coderabbitai review

PR #2740, opened 2026-06-29
feat(orm)!: Add typed typed mutation hooks #2365

### LINKED ISSUE #2365: Add type inference to `loadBeforeMutationEntities` and `loadAfterMutationEntities`
Is there a reliable way to infer the proper return type of `loadAfterMutationEntities` and `loadBeforeMutationEntities`?

https://github.com/zenstackhq/zenstack-v3/blob/35a9e0d6575d7baf7d0aa3e29484d7aedb56b4d0/packages/runtime/src/client/plugin.ts#L170-L174

https://github.com/zenstackhq/zenstack-v3/blob/35a9e0d6575d7baf7d0aa3e29484d7aedb56b4d0/packages/runtime/src/client/plugin.ts#L153-L158

If the return data is typed, it becomes easier working with any follow up queries:

```ts
export const onAfterPostDelete = definePlugin<SchemaType>({
  id: "onAfterPostDelete",
  onEntityMutation: {
    async beforeEntityMutation({ model, action, loadBeforeMutationEntities, client, queryId }) {
      const deletedPost = await loadBeforeMutationEntities() // <--- typed as Record<string, unknown>[] | undefined but its really a Post[]

      if (action === "delete" && model === "Post") {
        // Do some logic
        const tags = await client.tag.findMany({
          where: {
            postId: deletedPost[0].id, // <-- This doesn't work unless we cast deletedPost to proper type like Post[]
          },
        })
      }
    },
  },
})

```

genu:
fixes: #2365

## Summary by CodeRabbit 
- **New Features**
  - Added support for model-specific entity mutation hooks, with strongly typed loaded entities.
  - Model-specific hooks can run alongside catch-all hooks and receive the correct entity snapshots.
  - Per-model configuration for whether after-mutation hooks run inside or outside transactions.

- **Bug Fixes**
  - Fixed mutation hook handling so model-specific hooks execute even when only per-model hooks are provided.
  - Improved scheduling so after-mutation hooks correctly follow in-transaction vs out-of-transaction behavior.

- **Breaking Changes**
  - Catch-all entity mutation hooks must now be registered under `onEntityMutation.$all`.

- **Tests**
  - Expanded E2E coverage for per-model mutation hooks and updated related regression tests.

### DIFF
--- a/BREAKINGCHANGES.md
+++ b/BREAKINGCHANGES.md
@@ -4,3 +4,4 @@
 1. `@omit` and `@password` attributes have been removed
 1. SWR plugin is removed
 1. `makeModelSchema()` no longer includes relation fields by default — use `include` or `select` options to opt in, mirroring ORM behaviour
+1. ORM entity mutation catch-all hooks now need to be registered under `onEntityMutation.$all`
--- a/packages/orm/src/client/executor/zenstack-query-executor.ts
+++ b/packages/orm/src/client/executor/zenstack-query-executor.ts
@@ -123,7 +123,12 @@ export class ZenStackQueryExecutor extends DefaultQueryExecutor {
     }
 
     private get hasEntityMutationPluginsWithAfterMutationHooks() {
-        return (this.client.$options.plugins ?? []).some((plugin) => plugin.onEntityMutation?.afterEntityMutation);
+        return (this.client.$options.plugins ?? []).some((plugin) => {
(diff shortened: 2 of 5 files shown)


ymc9:
Hi @genu , my apologies for missing this PR. Looks good and I'll merge before the release of 3.10 since it's breaking.

ymc9 (approved):

PR #2760, opened 2026-07-18
fix: rename native type mappings to match datasource during Prisma schema generation

sanny-io:
Prisma's native type mappings rely on the datasource name. This is a problem because ZenStack uses a hardcoded prefix of `@db.*` for native type mappings.

For example, the following schema is valid in ZenStack, but **invalid in Prisma**. Migration commands will fail.

```prisma
datasource ds {
    provider = 'postgresql'
    url      = env('POSTGRES_URL')
}

model User {
    id String @id @db.Text // it must be `@ds.Text` for Prisma
}
```

This change renames native type mappings to match the datasource during Prisma schema generation. An alternative to this is renaming the datasource itself, but since that's user-defined, I thought this was less invasive.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prisma schema generation now rewrites native type mapping attributes to use the active datasource alias (for example, `@ds.Text`, `@ds.Boolean`, `@ds.JsonB`, `@ds.Timestamptz(3)`, `@ds.ByteA`) instead of emitting `@db.*`.
* **Tests**
  * Added a PostgreSQL-focused test to verify `@db.*` mappings are not produced and the corresponding `@ds.*` mappings are present.


ymc9:
Thanks @sanny-io ! I was naive to assume the `@db` was a fixed name for Prisma and got to know that late 😄. I think it's a good and pragmatic fix.

PR #2766, opened 2026-07-23
[CI baseline] test run off main — do not merge

ymc9:
Draft PR to get a Build-and-Test run on the current `main` code (plus one empty commit, since the workflow only triggers on `pull_request`).

Purpose: establish a baseline for the intermittent CLI-test failures seen on PR #2744 (three runs, three different `packages/cli` tests, all with the signature "zen CLI subprocess exits 0 but its output file is missing"). If this baseline run shows the same class of failure on `main`, the issue predates all recent changes and is environmental; if it's green repeatedly, the trigger is somewhere in the recent `dev` state.

Will be closed without merging.

### DIFF
--- a/packages/cli/src/actions/action-utils.ts
+++ b/packages/cli/src/actions/action-utils.ts
@@ -63,7 +63,9 @@ export async function loadSchemaDocument(
     const returnServices = opts.returnServices ?? false;
     const mergeImports = opts.mergeImports ?? true;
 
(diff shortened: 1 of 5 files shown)


ymc9:
Diagnostic mission accomplished — this PR's instrumentation identified the root cause of the intermittent CLI-test failures (version-check fetch stranding → event-loop drain → silent exit 0) and verified the fix over 3 consecutive green runs. The clean fix is in #2767. Closing and deleting the branch; logs remain available on this PR's check runs.

PR #2775, opened 2026-07-29
feat(cli): add --introspect option for studio command for bootstrapping without schema 

jiashengguo:
## Summary by CodeRabbit

* **New Features**
  * Added an `--introspect` option to generate a schema from your database when no schema file is available.

* **Enhancements**
  * Improved schema resolution flows for `--introspect` vs missing schema, including smarter `DATABASE_URL` handling for SQLite (relative `file:` rewriting).
  * Enhanced database pulls with explicit `--provider` and `--database-url` support (including improved SQLite path normalization).
  * Improved automatic driver/provider loading and clearer guidance when required drivers aren’t installed.

* **Bug Fixes**
  * Corrected SQLite nullability handling for primary key columns during introspection.

* **Tests**
  * Expanded coverage for schema resolution and driver/module loading behaviors.

### DIFF
--- a/packages/cli/src/actions/action-utils.ts
+++ b/packages/cli/src/actions/action-utils.ts
@@ -408,7 +408,7 @@ export async function loadPackage(pkgName: string) {
                 const projectRequire = createRequire(path.resolve(process.cwd(), 'package.json'));
                 return projectRequire(pkgName);
             } catch (requireErr) {
-                if (process.env.DEBUG) {
+                if (process.env['DEBUG']) {
                     console.error(importErr, resolvedImportErr, requireErr);
                 }
                 return null;
@@ -416,4 +416,3 @@ export async function loadPackage(pkgName: string) {
         }
     }
 }
-
--- a/packages/cli/src/actions/action-utils.ts
+++ b/packages/cli/src/actions/action-utils.ts
@@ -399,14 +399,14 @@ export async function loadPackage(pkgName: string) {
     } catch (importErr) {
         // If zenstack CLI is running directly using npx/pnpm in a temp folder,
         // dynamic package import will fail. Also check the current project folder.
+        let projectRequire: NodeJS.Require | undefined;
         try {
-            const projectRequire = createRequire(path.resolve(process.cwd(), 'package.json'));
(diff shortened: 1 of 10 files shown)


packages/cli/src/actions/proxy.ts
```diff
@@ -479,3 +488,188 @@ function startServer(
...
+
+        // Install required packages
+        await installDbDriverPackage(providerType);
+
+        // Introspect and generate schema
+        const urlFromEnv = !options.databaseUrl;
+        const schemaFile = await introspectAndGenerateSchema(databaseUrl, providerType, urlFromEnv);
+
+        // Run code generation from the generated schema
+        const { run: runGenerate } = await import('./generate');
```
ymc9: Do we need the dynamic "./generate" import here?

packages/cli/src/actions/action-utils.ts
```diff
@@ -382,3 +382,37 @@ export function startUsageTipsFetch() {
...
+}
+
+export async function loadPackage(pkgName: string) {
+    try {
+        return await import(pkgName);
+    } catch (importErr) {
+        // If zenstack CLI is running directly using npx/pnpm in a temp folder,
+        // dynamic package import will fail. Also check the current project folder.
+        try {
+            const projectRequire = createRequire(path.resolve(process.cwd(), 'package.json'));
```
ymc9: `projectRequire` is duplicated.

packages/cli/src/actions/proxy.ts
jiashengguo: Nice catch!  The funny thing is that I did notice this because it cause a build warnning.   But I also see the thinking process of claude for this:

```
Build succeeds. The INEFFECTIVE_DYNAMIC_IMPORT warning is because generate.ts is already statically imported by index.ts. Since we can't avoid this (the barrel export already includes generate), let me change the dynamic import to a static import instead — it'll be bundled the same way anyway:
```

So I just simply assume there is a reason for this without deep thinking. Go back to check the implement history, it seems it just use dynamic import at very beginning without a reason.

PR #2780, opened 2026-08-01
fix: add per-provider native type mapping diagnostics

### LINKED ISSUE #2759: ZModel accepts `@db.*` native type attributes on the sqlite provider (Prisma rejects them)
## Problem

ZModel's `@db.*` native type attributes are validated against the field's type (`@@@targetField`) but not against the datasource provider. Under a `sqlite` datasource, this compiles without any diagnostic:

```zmodel
datasource db {
    provider = "sqlite"
    url      = "file:./dev.db"
}

model M {
    id Int    @id
    v  String @db.VarChar(200) // silently ignored
}
```

Prisma's sqlite connector registers no native types at all and fails validation instead (verified with Prisma 6.19):

```
P1012: Native type VarChar is not supported for sqlite connector.
```

Since SQLite is affinity-typed, the migration engine (and DDL generation generally) ignores the attribute — so the user's declared intent is silently dropped. This surfaces when switching a project's provider to sqlite or when authoring against multiple providers: no error, no warning, no effect.

## Expected

Validation should scope native type attributes by connector, like Prisma:

- reject any `@db.*` attribute when the datasource provider is `sqlite`;
- ideally, also enforce per-connector attribute lists for `postgresql`/`mysql` (e.g. `@db.Uuid` is pg-only, `@db.LongText` is mysql-only), which would additionally resolve the related `@db.TinyInt`-on-Boolean mismatch tracked separately.

## Context

Found while building the Prisma physical-schema parity test suite for the native migration engine. Physical-schema parity itself is unaffected (Prisma-validated schemas can't contain the construct), but the validation behavior diverges.

sanny-io:
Fixes #2759 

## Summary by CodeRabbit

* **New Features**
  * Added validation for native database type mappings based on the selected datasource provider.
  * Supports provider-specific mappings for SQLite, PostgreSQL, and MySQL.

* **Bug Fixes**
  * Reports clear errors when a native type mapping is incompatible with the configured database provider.

* **Tests**
  * Added coverage for supported and unsupported native mappings across all supported providers.

### DIFF
--- a/packages/language/src/utils.ts
+++ b/packages/language/src/utils.ts
@@ -7,6 +7,7 @@ import { PLUGIN_MODULE_NAME, STD_LIB_MODULE_NAME, type ExpressionContext } from 
 import {
     InternalAttribute,
     isArrayExpr,
+    isAttribute,
     isBinaryExpr,
     isConfigArrayExpr,
     isDataField,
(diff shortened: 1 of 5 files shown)


ymc9 (approved):
Thank you @sanny-io . LGTM! Merging.

PR #2792, opened 2026-08-07
feat(orm): strict types

sanny-io:
Adds support for specifying custom type definitions should reject unknown fields via a new `@@strict` attribute. This is reflected in both intellisense and runtime input validation.

## Sample Usage

```prisma
model User {
    id      String   @id @default(uuid())
    profile Profile? @json
}

type Profile {
    bio String

    @@strict
}
```

Why not `@json(strict: true)`? A type-level `@@strict` attribute is simpler to determine strictness, and **also works with custom procedures**.

## Summary by CodeRabbit

- **New Features**
  - Added the `@@strict` attribute for type definitions.
  - Strict types reject unknown fields in JSON data and API operations.
  - Generated schemas and client typings preserve strictness across supported clients.
  - Added profile support and an `updateProfile` operation to the example API.

- **Bug Fixes**
  - Non-strict types continue allowing additional fields.
  - Invalid use of `@@strict` on models or enums now reports a validation error.


ymc9 (approved):
Thanks @sanny-io , the changes look great. It's an excellent addition!

Issue #2764, opened 2026-07-20
`zen generate --lite` does not strip attributes from `typeDefs`, shipping access policies to the browser


ErikDakoda:
### Version

`@zenstackhq/sdk` / `@zenstackhq/cli` **3.8.3** (current latest).

### Summary

`zen generate --lite` strips attributes from **models** and from **fields**, but not from **`typeDefs`**. Any access-control policy declared on a `type` rather than a `model` therefore still ships to the browser in `schema-lite.ts`.

This matters because the docs present `schema-lite.ts` as the client-side artifact and justify it partly on the grounds that the full schema embeds access policies. For schemas that put shared policies on `type` declarations, that guarantee does not hold.

To be clear about severity: this is **information disclosure, not privilege escalation** — the server still enforces policies. But the client bundle ends up describing the authorization model.

### Reproduction

In our generated `schema-lite.ts`, `--lite` removed every field-level attribute (487 `@default` and 55 `@updatedAt` occurrences → 0) and every model-level attribute (0 remaining), but left **18 type-level attributes** intact:

| typeDef | remaining attributes |
| --- | --- |
| `UserBase` | 5 × `@@allow` |
| `Base` | 2 × `@@allow`, 2 × `@@deny`, 3 × `@@index` |
| `TaxonomyBase` | 3 × `@@allow`, 3 × `@@index` |

The retained expressions are fully readable in the emitted AST — for example an `@@allow('all', auth().role == 'ADMIN')` rule, and a space-membership read rule, both reconstructable from the shipped `ExpressionUtils` calls.

### Cause

`packages/sdk/src/ts-schema-generator.ts`. The model path honours `lite`; the typeDef path receives the flag and uses it only for fields, not for the type's own attributes.

```ts
// createDataModelObject (line ~374)
const allAttributes = lite
    ? [] // in lite mode, skip all model-level attributes
    : getAllAttributes(dm).filter(...);

// createTypeDefObject (line ~503)
const allAttributes = getAllAttributes(td);   // `lite` is in scope but unused here
```

`createTypeDefObject` does pass `lite` down to `createDataFieldObject(field, undefined, lite)`, so field attributes inside typeDefs are stripped correctly — it is only the type-level attributes that are missed.

### Suggested fix

Mirror the model path:

```ts
const allAttributes = lite ? [] : getAllAttributes(td);
```

### Related, and possibly a separate decision

[...]

PR #2794, opened 2026-08-10
fix(zod): lite schema validation

### LINKED ISSUE #2793: Lite schema strips @@@validation attributes, breaking createSchemaFactory client validation
## Problem

`zen generate --lite` (and `liteOnly`) drops **all** attributes from the emitted TypeScript schema — both model-level and field-level — via `TsSchemaGenerator`:

- model: `const allAttributes = lite ? [] : getAllAttributes(dm)...`
- field: `if (!lite && field.attributes.length > 0) ...`

Additionally, `createTypeDefObject` **always** emits every typedef attribute (including policy `@@allow` / `@@deny`), even in lite mode — the opposite inconsistency.

This silently breaks the documented client-side `createSchemaFactory` / `@zenstackhq/zod` path: `addStringValidation` / `addNumberValidation` read `fieldDef.attributes`, so with a lite schema they emit bare `z.string()` / `z.number()` with no `.min()` / `.max()` / `.regex()` / `.date()` / `.time()` / etc.

Verified still present in **3.8.3 and 3.9.0**. There is no CLI flag to keep validation attributes in lite mode.

## Symptom (before / after)

`User.dateOfBirth String? @date` (formerly `@regex` ISO date):

- **Full `schema.ts`**: field has `attributes: [{ name: "@date" }]` → factory emits `z.string().date()` → editor renders a DatePicker and validates inline.
- **Lite `schema-lite.ts`**: no `attributes` → factory emits bare `z.string()` → editor renders a plain text field; invalid values surface as server `ORMError INVALID_INPUT` toasts instead of inline zod errors. Char counters from `@length` also disappear (`stringLengthLimitsOfZodSchema` finds no checks).

## Proposed fix

In lite mode, keep attributes whose **resolved declaration** carries the stdlib `@@@validation` marker (`@length`, `@regex`, `@email`, `@url`, `@datetime`, `@date`, `@time`, `@phone`, `@startsWith`, `@endsWith`, `@contains`, `@trim`, `@lower`, `@upper`, `@gt`, `@gte`, `@lt`, `@lte`, and model-level `@@validate`). Whitelist by the marker, not by name — so policy attributes can never leak into the client bundle.

Apply the same filter in `createTypeDefObject` (today it emits all typedef attributes unconditionally, which can leak policy into lite).

Optional: gate behind `--lite-validation` if a hard break of current lite output is a concern; defaulting it on would match the documented factory usage.

## Workaround

We currently re-emit `schema-lite.ts` after `zen generate` with a local script that patches `TsSchemaGenerator` prototype methods and filters by `@@@validation`. Happy to turn that into a PR against `ts-schema-generator.ts` if maintainers are open to it.

## Environment

- `@zenstackhq/sdk` / `cli` / `zod` 3.9.0
- Client schemas via `createSchemaFactory(schemaLite)`

sanny-io:
Fixes #2793 and #2764

Adds a new `@@@lite` attribute, which specifies that an attribute should not be stripped when generating lite schemas. All validation attributes, `@meta`, and `@@meta`, are decorated with it.

`@default` and `@updatedAt` are also decorated with because the frontend hooks rely on checking them.

The Zod factory test suite now runs against both a full schema and a lite schema. 

## Summary by CodeRabbit

* **New Features**
  * Added lite schema generation that retains supported validation, metadata, and default attributes.
  * Added lite-compatible schema coverage for models, relationships, computed fields, and authorization settings.

* **Tests**
  * Expanded schema generation tests to verify supported attributes are retained and unsupported attributes are removed.
  * Added separate full-schema and lite-schema test coverage.


ymc9 (approved):
LGTM @sanny-io . I've simplified the zod test part of changes to use `describe.each` instead of multiple vitest projects.

PR #2798, opened 2026-08-11
fix(orm): allow setting `null` on optional `Decimal` fields when using `sqlite` provider

### LINKED ISSUE #2797: SQLite dialect: writing null to a nullable Decimal field crashes with "Cannot read properties of null (reading 'toString')"
## Description

With the SQLite dialect (v3), any `create`/`update` that writes `null` to a nullable `Decimal` field crashes:

```
TypeError: Cannot read properties of null (reading 'toString')
 ❯ SqliteCrudDialect.transformInput src/client/crud/dialects/sqlite.ts:102:46
 ❯ CreateOperationHandler.create src/client/crud/operations/base.ts:454:55
```

## Repro

```zmodel
model Project {
    id         String   @id @default(cuid())
    hourlyRate Decimal?
}
```

```ts
await db.project.create({ data: { hourlyRate: null } }); // throws
```

## Cause

`SqliteCrudDialect.transformInput` only early-returns on `undefined`, and the `Decimal` branch calls `.toString()` unguarded ([sqlite.ts:102](https://github.com/zenstackhq/zenstack/blob/main/packages/orm/src/client/crud/dialects/sqlite.ts#L102)):

```ts
case 'Decimal':
    return (value as Decimal).toString();
```

The Postgres dialect guards this same case:

```ts
value !== null ? value.toString() : value
```

## Suggested fix

Mirror the Postgres guard in the SQLite dialect:

```ts
case 'Decimal':
    return value !== null ? (value as Decimal).toString() : value;
```

Affects at least 3.6.4 through 3.9.0; current `main` still has the unguarded call. Happy to send a PR if useful.

sanny-io:
Fixes #2797 

## Summary by CodeRabbit

* **Bug Fixes**
  * Optional Decimal fields in SQLite now correctly preserve `null` values when creating and retrieving records.
  * Non-null Decimal values continue to be handled as expected.

* **Tests**
  * Added end-to-end coverage for nullable Decimal fields in SQLite.

### DIFF
--- a/packages/orm/src/client/crud/dialects/sqlite.ts
+++ b/packages/orm/src/client/crud/dialects/sqlite.ts
@@ -99,7 +99,7 @@ export class SqliteCrudDialect<Schema extends SchemaDef> extends BaseCrudDialect
                           ? new Date(value).toISOString()
                           : value;
                 case 'Decimal':
-                    return (value as Decimal).toString();
+                    return value !== null ? value.toString() : value;
                 case 'Bytes':
                     return Buffer.from(value as Uint8Array);
                 default:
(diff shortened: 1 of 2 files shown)


ymc9 (approved):

ymc9:
LGTM @sanny-io . I'm merging and will release a new patch update soon.

PR #2804, opened 2026-08-15
fix(language): `null` suggestions in `DefaultValue` context

sanny-io:
`@default(null)` is never valid in ZModel and should not be suggested.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed the invalid `null` suggestion from `@default` attribute completions.
  * Boolean values (`true` and `false`) remain available as completion options.


sanny-io:
Closing this as it can interfere with expressions like `@default(null == a)`

PR #2807, opened 2026-08-17
feat(language): implicitly convert `enum` references to arrays

### LINKED ISSUE #1211: [Feature Request] Allow for Prisma enums in validation rules
**Is your feature request related to a problem? Please describe.**
While Zenstack is already offering great validation capabilities, it would be nice if it were possible to use Prisma enums
inside a validate block, for example with an `in` operation. This would allow for an always-in-sync validation instead of relying on manual updates in both codebase and validation rules.

One might ask why not set the field as the enum? The answer to that being that working with primitives is easier to migrate and also better supported across third party vendors, like Hasura GraphQL.

**Describe the solution you'd like**
Allow for Prisma enums to be referenced in validations rules, especially with the `in` operator.

**Describe alternatives you've considered**
-

**Additional context**
-

sanny-io:
Closes #1211

## Summary by CodeRabbit

- **New Features**
  - Added enum-based validation for string fields, including checks against all permitted enum values.
  - Added support for enum-constrained model fields, with valid and invalid status handling.
  - Added address type validation with `RESIDENTIAL` and `COMMERCIAL` values.
  - Preserved default metadata and strict settings in generated lightweight schemas.

- **Bug Fixes**
  - Improved enum reference resolution and generated validation expressions for consistent schema behavior.


ymc9 (commented):
Hi @sanny-io ,

Thanks for working on this. I think there are two issues with this PR:

1. Enums are are now reference target, so they can potentially appear in places where references are allowed (e.g., `@default(Role)`?, haven't tried though ...).
2. I think `x in Enum` should also work in access policies, and I believe it already works today, but a caveat is that enum fields can be name mapped with `@map`, so the current approach that directly translates into an array literal probably broke it.

PR #2809, opened 2026-08-20
fix(policy): do not treat Json kind keys as Kysely nodes

### LINKED ISSUE #2791: v3 policy: create() with a Json value containing a top-level "kind" key fails with "Invariant failed: expecting a ValueNode"
### Description

A `create()` whose **Json field value contains a top-level `kind` key** fails through the policy plugin with:

```
Invariant failed: expecting a ValueNode
```

The same payload succeeds via `update()`, and via `create()` when the `kind` key is renamed or omitted — so the JSON value itself is fine; only the create path through the policy plugin rejects it.

This looks like the policy plugin's pre-create check (`unwrapCreateValueRow`) walking each column value looking for embedded Kysely query/operation nodes, which it recognises by their `kind` discriminator property. A plain JSON object that happens to have its own `kind` key is mistaken for a node and fails the `item.kind === "ValueNode"` invariant.

`kind` is a fairly natural domain field name (ours tags imagery/artwork payloads), so this is easy to hit and confusing to diagnose — the error surfaces as a policy-layer invariant with no hint that a payload key is the trigger.

### Reproduction

Schema:

```zmodel
model Item {
    id      Int  @id @default(autoincrement())
    payload Json

    @@allow('all', true)
}
```

Code:

```ts
import { ZenStackClient } from "@zenstackhq/orm";
import { PolicyPlugin } from "@zenstackhq/plugin-policy";
import { schema } from "./zenstack/schema";

const db = new ZenStackClient(schema, { dialect }).$use(new PolicyPlugin());

// ❌ throws: Invariant failed: expecting a ValueNode
await db.item.create({ data: { payload: { kind: "artwork" } } });

// ✅ works — same shape, key renamed
await db.item.create({ data: { payload: { type: "artwork" } } });

// ✅ works — same value via update
const item = await db.item.create({ data: { payload: {} } });
await db.item.update({ where: { id: item.id }, data: { payload: { kind: "artwork" } } });
```

### Expected behaviour

A JSON column value is treated as opaque data regardless of its keys; `create()` and `update()` accept the same payloads.

### Actual behaviour

`create()` through the policy plugin throws `Invariant failed: expecting a ValueNode` whenever the JSON value has a top-level `kind` key. `update()` accepts the identical value.

### Workaround

Create without the `kind` key, then patch it in with an `update()` in the same transaction — the stored payload ends up identical. We pin the quirk (and the workaround) with an integration test so we can drop the dance once fixed.

### Environment

- `@zenstackhq/orm` 3.9.0, `@zenstackhq/plugin-policy` 3.9.0, `@zenstackhq/cli` 3.9.0
- Postgres dialect (`@zenstackhq/orm/dialects/postgres` + `pg`)
- Also reproduced on 3.7.2

BetterAndBetterII:
Policy pre-create unwrapping treated any object with a kind field as a Kysely operation node, so create({ payload: { kind: "artwork" } }) failed with expecting a ValueNode. Only ValueNode and DefaultInsertValueNode are treated as nodes.

Fixes #2791

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed handling of JSON data containing a top-level `kind` field when policies are enabled.
  - JSON values now remain intact during creation, bulk creation, retrieval, and updates.

- **Tests**
  - Added regression coverage for JSON payloads with top-level `kind` fields in PostgreSQL schemas.

### DIFF
--- a/packages/plugins/policy/src/policy-handler.ts
+++ b/packages/plugins/policy/src/policy-handler.ts
@@ -1079,7 +1079,12 @@ export class PolicyHandler<Schema extends SchemaDef> extends OperationNodeTransf
         const result: { node: OperationNode; raw: unknown }[] = [];
         for (let i = 0; i < data.length; i++) {
             const item = data[i]!;
-            if (typeof item === 'object' && item && 'kind' in item) {
+            if (
+                typeof item === 'object' &&
+                item &&
+                'kind' in item &&
+                (item.kind === 'ValueNode' || item.kind === 'DefaultInsertValueNode')
+            ) {
                 if (item.kind === 'DefaultInsertValueNode') {
                     result.push({ node: ValueNode.create(null), raw: null });
                     continue;
--- a/tests/regression/test/issue-2791.test.ts
+++ b/tests/regression/test/issue-2791.test.ts
@@ -1,0 +1,54 @@
+import { createPolicyTestClient } from '@zenstackhq/testtools';
+import { describe, expect, it } from 'vitest';
+
+// https://github.com/zenstackhq/zenstack/issues/2791
+// Policy pre-create checks unwrap insert values and used to treat any object with a
+// `kind` key as a Kysely operation node. A Json payload like `{ kind: "artwork" }`
+// then failed with `Invariant failed: expecting a ValueNode`.
+const schema = `
+model User {
+    id Int @id @default(autoincrement())
+}
+
+model Item {
+    id      Int  @id @default(autoincrement())
+    payload Json
+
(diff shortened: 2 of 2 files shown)


sanny-io:
coderabbit's comment looks correct to me in that this would fail similarly to the original issue, just less frequently.

See potential solution below.

https://github.com/zenstackhq/zenstack/issues/2791#issuecomment-5215606603

ymc9 (commented):
@BetterAndBetterII thanks for working on this! I've made an additional commit based on coderabbit's comments. Will merge once CI passes.

ymc9 (approved):

PR #2819, opened 2026-08-25
fix(orm): handle mapping enum array (#2818)

### LINKED ISSUE #2818: v3 doesnt support enum array
Enum array is not handled correctly
Its cannot be select/insert/update

```zmodel
enum TestTag {
  INFO @map("info")
  WARN @map("warn")

  @@map("post_tag")
}

model Test {
  id   String  @id @default(cuid())
  tags TestTag[]
}
```

```txt
    dbErrorMessage: 'malformed array literal: "info"',
    sql: 'select "Test"."id" as "id", case when "Test"."tags" = $1 then $2 when "Test"."tags" = $3 then $4 else cast("Test"."tags" as text) end as "tags" from "public"."Test" limit $5 ',
    sqlParams: [ 'info', 'INFO', 'warn', 'WARN', 1 ],
    reason: 'db-query-error',
```

- ZenStack version: 3.9.2
- Database type: PostgreSQL
- Node.js/Bun version: 22
- Package manager: npm

**Additional context**
Repro repo: https://github.com/DoctorFTB/zenstack-v3-enum-bug

DoctorFTB:
Handle enum array insert / select / update

Closes #2818

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of PostgreSQL enum arrays during data selection.
  * Corrected mapping for optional, empty, and populated enum values.
  * Preserved expected ordering when retrieving records containing enum arrays.

* **Tests**
  * Added regression coverage for enum fields and enum arrays.

### DIFF
--- a/packages/orm/src/client/executor/name-mapper.ts
+++ b/packages/orm/src/client/executor/name-mapper.ts
@@ -16,6 +16,7 @@ import {
     OperationNodeTransformer,
     PrimitiveValueListNode,
     type QueryId,
+    RawNode,
     ReferenceNode,
     ReturningNode,
     SelectAllNode,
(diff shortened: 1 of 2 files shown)


ymc9 (commented):
Hey @DoctorFTB , thanks for working this and my apologies for following up late. Overall. the code looks good to me. I think it's missing the case where part of the fields are mapped in an enum. I'll make a commit for it and get it merged.

ymc9 (approved):
LGTM. Pushed a follow-up commit to map enum array values per element so enums with partially mapped members work, plus a regression case. Array filters (has/hasSome/hasEvery/equals) on mapped enum arrays will be handled in a separate follow-up.

PR #2842, opened 2026-09-16
fix(policy): skip per-row create checks when the filter doesn't reference the row

### LINKED ISSUE #2841: createMany under the policy plugin issues one serial round trip per row, even when the policy filter is constant
### Summary

`createMany` on a model with a create policy is checked row by row: `enforcePreCreatePolicy` loops the value rows and `await`s a separate `select exists (...)` query for each one. A 32-row `createMany` costs 32 sequential database round trips before the insert.

When the policy expression doesn't reference the row — anything built only from `auth()` — every one of those queries is identical and the compiled filter collapses to `where true`, so the loop does no work at all. We measured 32 serial round trips evaluating `where true` on a single `createMany`.

### Versions

`@zenstackhq/orm` and `@zenstackhq/plugin-policy` 3.6.4, reproduced by inspection on 3.9.4 (unchanged). Postgres.

### Reproduction

```zmodel
model AuditLog {
  id        String   @id @default(uuid())
  userId    String
  action    String
  createdAt DateTime @default(now())

  @@allow('create', auth().isSystem)
  @@allow('read', auth().isAdmin)
}
```

```ts
const db = client.$setAuth({ id: 'system', isSystem: true, isAdmin: false });

await db.auditLog.createMany({
  data: Array.from({ length: 32 }, (_, i) => ({ userId: `u${i}`, action: 'x' })),
});
```

### Observed

32 sequential queries of this shape, then the insert:

```sql
select exists (
  select 1 as "_"
  from (select cast("$values"."column1" as text) as "id", /* ... */) as "AuditLog"
  where true
) as "$condition"
```

The `where` clause is literally `where true` — `auth().isSystem` is resolved from the auth context while the filter is built, so the predicate never references the values table and cannot differ between rows.

### Expected

One check for the batch. For a filter that doesn't reference the row, ideally none.

### Cause

`enforcePreCreatePolicy` ([policy-handler.ts:914-941](https://github.com/zenstackhq/zenstack/blob/v3.9.4/packages/plugins/policy/src/policy-handler.ts#L914-L941)):

```ts
for (const values of valueRows) {
    /* ... */
    await this.enforcePreCreatePolicyForOne(mutationModel, fields, values.map((v) => v.node), proceed);
}
```

`enforcePreCreatePolicyForOne` ([:1006](https://github.com/zenstackhq/zenstack/blob/v3.9.4/packages/plugins/policy/src/policy-handler.ts#L1006)) builds a single-row values table (`buildValuesTableSelect(allFields, [allValues])`), rebuilds the filter with `buildPolicyFilter(model, undefined, 'create')`, and issues one query. The filter is never examined to see whether it depends on the row.

There is a constant-policy short-circuit, but it doesn't apply here. `preCreateCheck` ([:180](https://github.com/zenstackhq/zenstack/blob/v3.9.4/packages/plugins/policy/src/policy-handler.ts#L180)) calls `tryGetConstantPolicy`, which tests the **ZModel expression**:

```ts
private isTrueExpr(expr: Expression) {
    return ExpressionUtils.isLiteral(expr) && expr.value === true;
}
```

That matches only a literal `@@allow('create', true)` written in the schema. An `auth()`-derived predicate is not a literal, so it never short-circuits — even though the **compiled** filter is a true node.

Elsewhere in this file the compiled node *is* what gets tested: `preUpdateCheck` uses `isTrueNode(fieldLevelFilter)` at [:204](https://github.com/zenstackhq/zenstack/blob/v3.9.4/packages/plugins/policy/src/policy-handler.ts#L204), and `isTrueNode` is used in several other places. The create path is the one that only inspects the source expression.

### Suggested fix

Hoist the filter out of the loop in `enforcePreCreatePolicy` and branch on it:

1. **`isTrueNode(filter)` → skip entirely.** No query. This covers every `auth()`-only create policy.
2. **Filter doesn't reference the values table → check once**, not per row. The answer is identical for every row by construction.
3. **Otherwise → check the whole batch in one query.** Build the values table from *all* rows rather than `[allValues]`, and assert that the surviving count equals the input count.

Option 3 is the technique already used for post-update verification at [:416-421](https://github.com/zenstackhq/zenstack/blob/v3.9.4/packages/plugins/policy/src/policy-handler.ts#L416-L421):

```ts
eb(eb.fn('COUNT', [eb.lit(1)]), '=', Number(updateResult.numAffectedRows ?? 0)).as('$condition')
```

Applying the same shape to creates would make `createMany` a constant two queries regardless of batch size, and preserve the current all-or-nothing rejection semantics.

### Impact

The cost scales with batch size and the round trips are serial, so it shows up as request latency rather than only connection churn. It affects any policy-protected model written in bulk. We hit it on an audit-log write whose recipient count grows with team size, so the cost grows silently over time.

bgowers:
Fixes #2841.

### Problem

`createMany` on a policy-protected model checks each row with its own `select exists (...)` round trip, awaited in sequence. When the create filter is built only from `auth()` it doesn't reference the row, so it compiles to a constant and every check re-asks the same question. A 32-row batch cost 32 serial queries, each evaluating `where true`.

### Change

`enforcePreCreatePolicy` now builds the model-level create filter once and returns early when `isTrueNode(filter)` holds.

`preCreateCheck` already had a constant-policy short-circuit, but it goes through `tryGetConstantPolicy` → `isTrueExpr`, which only matches a literal `true` written in the ZModel. An `auth()`-derived predicate is never a literal, so it never qualified. `preUpdateCheck` already tests the *built* node with `isTrueNode`; this applies the same test on the create path.

Row-dependent filters are unaffected — they don't produce a true node, so the per-row loop runs exactly as before. The many-to-many join-table path is left alone.

I kept this to the constant case. The issue also suggests checking a whole batch in one query for row-dependent filters, using the `COUNT(1) = n` shape already used for post-update verification — happy to follow up with that separately if you'd like it.

### Tests

`tests/regression/test/issue-2841.test.ts` covers three cases:

- an `auth()`-only create filter issues **no** policy check for a 5-row `createMany`, and all rows land
- the same filter still **rejects** the batch when `auth()` makes it false
- a row-referencing filter still runs **one check per row**, and a single offending row still rejects the whole batch

The second and third are the ones that would catch this change going too far.

### Verification

- New tests: 3 passed
- `tests/e2e/orm/policy`: 46 files, 281 passed, 21 skipped, no type errors
- `tests/regression`: 161 passed, 18 skipped, 1 failed — `issue-2603` ("implicit m2m with models in different custom schemas"). I confirmed that one fails identically on a clean `dev` with this change reverted and the package rebuilt, so it's pre-existing and unrelated. It uses `createTestClient` with no policy plugin, so `PolicyHandler` isn't constructed at all in that test.
- `prettier --check` clean on both files

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved policy enforcement for batched record creation.
  - Batches governed by universally allowing policies can now be processed without unnecessary per-record checks.
  - Universally denying policies correctly reject the entire batch without creating records.
  - Policies that depend on individual record values continue to validate every record, rejecting the full batch if any record fails.


ymc9 (dismissed):
Hi @bgowers , thanks for making this improvement. The changes look good to me. I'll fix the test build error and merge.

ymc9 (approved):

